Issue #152, opened 2017-09-20
Release at Pypi


lresende:
http://peterdowns.com/posts/first-time-with-pypi.html

lresende:
We have already completed the release
https://pypi.python.org/pypi/jupyter_enterprise_gateway/0.6.0

PR #164, opened 2017-09-28
Tentative changes on Makefile and setup.py for PyPI release

liukun1016:
Related to issue #152 . Created this PR tentatively as a reference.

Besides following the tutorial provided there, we need to remove the `register` in `make release`, and add `-r pypi` as a flag to the command `python setup.py` for both bdist and sdist.

Also parameterize some variables inside `setup.py`.


Makefile
```diff
@@ -68,12 +68,12 @@ install: ## Make a conda env with dist/*.whl and dist/*.tar.gz installed
 	conda env remove -y -n $(ENV)-install
 
 bdist: ## Make a dist/*.whl binary distribution
-	$(SA) $(ENV) && python setup.py bdist_wheel $(POST_SDIST) \
-		&& rm -rf *.egg-info
+	$(SA) $(ENV) && python setup.py bdist_wheel $(POST_SDIST) -r pypi \
```
kevin-bates: The `-r pypi` argument should be part of the `$(POST_SDIST)` variable since it only applies to upload and should only be used on `make release` invocations.  Same goes for the `sdist:` target on L75.

If I run `python setup.py bdist_wheel -r pypi` I get the following output:
```
python setup.py bdist_wheel -r pypi
usage: setup.py [global_opts] cmd1 [cmd1_opts] [cmd2 [cmd2_opts] ...]
   or: setup.py --help [cmd1 cmd2 ...]
   or: setup.py --help-commands
   or: setup.py cmd --help

error: option -r not recognized
```

liukun1016: yes it makes sense to specify the arguments for pypi, already pushed the changes

liukun1016:
hmm, it's really weird that after we migrated our project to jupyter-incubator, for my previously repo forked from our STC private elyra/enterprise_gateway repo, I pushed my changes there related to this PR, but the new commit would not be showed/appended here. 

Then I forked this repo from jupyter-incubator again to my repo and pushed my changes there for the issue #152, but it was not showed here as well. So I created another PR for the same issue (https://github.com/jupyter-incubator/enterprise_gateway/pull/199). 

If necessary, we could (and probably should) close this one and use the other to track the issue. 

Let me know if I have any misunderstanding here. (due to my lack of expertise on Github repo migration)

lresende:
Obsolete

PR #187, opened 2017-10-17
Skip test_default_kernel_name until Notebook issue is resolved.

kevin-bates:
A recent change to Notebook (PR 2853) introduced an issue such that
the NoSuchKernel string is no longer present in the response body
when attempting to use a non-existent kernel.  This condition is
the basis for checking of the default kernel is used.  As a result,
that test is now marked as skipped until that Notebook issue (2942)
is resolved or addressed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ python:
     - 3.6
 
 env:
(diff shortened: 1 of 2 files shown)


kevin-bates:
PR merged.

Issue #193, opened 2017-10-18
Update PYTHONPATH in kernel.json to be able to import locally installed modules 


sanjay-saxena:
To be able to successfully `import` modules that are installed using `!pip install --user <module_name>` from within a notebook cell, `PYTHONPATH` defined in `kernel.json` must be augmented to include the following path:

```
${HOME}/.local/lib/python-2.7/site-packages
```

This change should be made to the `kernel.json` files defined for both `YARN Client Mode`(aka `Distributed Proxy Mode`) and `YARN Cluster Mode`.

sanjay-saxena:
@kevin-bates -- Who owns `/opt/anaconda2/share/jupyter/kernels/kernel.json`? Was it populated/created by JKG?

kevin-bates:
Its "brought into existence" by virtue of installing the ipython kernel - which comes, in this environment, as the result of installing anaconda.  I'm not entirely sure where it lives in github.  Modifications like you're suggesting are the responsibility of the entity producing the end result product and not necessarily the provider.  

JKG does not deliver any kernelspecs and the kernelspecs we deliver are for example purposes.  All envs will likely need modifications here and there to what we provide.  That said, I agree that our _example_ python kernelspecs should include the `${HOME}` segment entry since we're actually introducing multi-user _sensitivity_.

sanjay-saxena:
This was fixed by PR #194.

PR #194, opened 2017-10-19
Updated PYTHONPATH in kernel.json for locally installed modules

sanjay-saxena:
Issue #193: Update PYTHONPATH in kernel.json for locally installed modules

Python modules that are installed from within a notebook cell using `!pip install --user <module-name>` get installed under `$HOME/.local/lib/python-<ver>/site-packages` folder. For the module to be successfully imported using `import <module-name>` in a notebook cell, `PYTHONPATH` defined in `kernel.json` for Python-specific kernelspecs should be augmented to include the aforementioned path.

### DIFF
--- a/etc/kernelspecs/spark_2.1_python_yarn_client/kernel.json
+++ b/etc/kernelspecs/spark_2.1_python_yarn_client/kernel.json
@@ -7,7 +7,7 @@
   "env": {
     "SPARK_HOME": "/usr/hdp/current/spark2-client",
     "PYSPARK_PYTHON": "/opt/anaconda2/bin/python",
-    "PYTHONPATH": "/usr/hdp/current/spark2-client/python:/usr/hdp/current/spark2-client/python/lib/py4j-0.10.4-src.zip",
+    "PYTHONPATH": "${HOME}/.local/lib/python2.7/site-packages:/usr/hdp/current/spark2-client/python:/usr/hdp/current/spark2-client/python/lib/py4j-0.10.4-src.zip",
     "SPARK_YARN_USER_ENV": "PYTHONUSERBASE=/home/yarn/.local,PYTHONPATH=/usr/hdp/current/spark2-client/python:/usr/hdp/current/spark2-client/python/lib/py4j-0.10.4-src.zip,PATH=/opt/anaconda2/bin:$PATH",
     "SPARK_OPTS": "--master yarn --deploy-mode client --name ${KERNEL_ID:-ERROR__NO__KERNEL_ID}",
     "LAUNCH_OPTS": ""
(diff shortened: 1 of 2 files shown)


kevin-bates:
LGTM

lresende:
What are you trying to solve here ? If the environment has "native" python, why then only adding the  user specific packages and not the "global packages" as well ? 

Also, how this works when you have both Anaconda + local python with different set of packages ? 

I am inclined to say we "recommend" anaconda, and document the changes for any other way ? 

Thoughts ?

sanjay-saxena:
@lresende -- The problem we are trying to solve is this -- If a python module is installed from a notebook which is tied to a remote Python kernel using  `!pip install --user <module-name>`, then it gets installed under `$HOME/.local/lib/python-2.7/site-packages folder`. Note that the module was installed using the recommended `--user` switch. However, when the module is imported from another cell using `import <module-name>`, then currently we see an `ImportError`.

To address this problem, we thought augmenting `PYTHONPATH` defined in the `kernel.json` files that are used by the remote Python kernels was needed. Please let us know if there is a better way to address this issue.

There is another problem where if the notebook is tied to a local/default Python kernel(represented by `Python 2` in the drop-down in the browser UI), then we need to modify `kernel.json` under `/opt/anaconda2//share/jupyter/kernels/python2/`. @kevin-bates suggested that the site-admins who would be setting up the environment for the notebook authors will be responsible for updating  `/opt/anaconda2//share/jupyter/kernels/python2/kernel.json`.

lresende:
These kernelspecs are examples on how one could customize it to run distributed kernels. As I mentioned I am inclined to suggest users to use Anaconda, as the regular data science packages are already available. The solution then is, to deploy anaconda and make it the default python in all worker nodes that can be running kernels. This is what we do in the ansible helper scripts.

kevin-bates:
I guess I'm confused.  Does the Anaconda configuration support per-user package locations or is it inherently a shared environment?  If the latter how do we support sand-boxing?  If the former, how do you specify a particular package should be placed into the per-user location?

I completely agree that the kernelspecs we provide should be treated as nothing more than examples, but I also believe those examples need to address the multi-tenancy functionality that we advertise and which I don't think we've got completely identified.

lresende:
I did some further investigation and it seems that this is the place where packages installed with --user are located. The issue is that it depends on the version of python you are using anaconda with (e.g. in my machine, it goes to /Users/lresende/.local/lib/python3.6/site-packages/). Then we need to consider which version of python got installed.

sanjay-saxena:
@lresende  -- @kevin-bates and I did talk about the issue with respect to different versions of Python. We felt that the `kernel.json` that we provide is more of template/sample which need to be tweaked based on whichever version of Python or Spark is being used. So, if we clearly document the fact that the administrators should customize `kernel.json` based on the Python version that they are using, then the onus will be on them to update the version number accordingly.

PR #196, opened 2017-10-20
Documented that Enterprise Gateway be started by a non-root service user

### LINKED ISSUE #191: Enterprise Gateway should be run under a non-root service user
Enterprise Gateway's documentation should be augmented to include the fact that it **MUST** be run under a _non-root_ service user.

kevin-bates:
Also documented what the remote user is used for and that the we recommend
the two accounts be the same.  Added warning message to start script to
emphasize non-root startups.

While in the file I replaced the NB2KG documentation to not include the
reference to the incorrect docker image and how to get the correct version
deployed.  Also added a comment that once NB2KG is updated, the document
will be changed accordingly.

Fixes #191


lresende:
LGTM

PR #200, opened 2017-10-23
Remove use remote-user command line option and env variable.

### LINKED ISSUE #198: Look into removing EG_REMOTE_USER and command-line option
TIL that running EG as one user and `EG_REMOTE_USER` pointing at a different user failed to startup (client-mode) kernels on remote hosts due to password-less ssh authentication failures - even when BOTH users are properly configured for ssh.  Since the default value for `EG_REMOTE_USER` is `$USER`, this suggests that we should drop `EG_REMOTE_USER` and its command line option (`--EnterpriseGatewayApp.remote_user`) and just let the "service user" be the "remote user" (as suggested in the [recent doc update](http://jupyter-enterprise-gateway.readthedocs.io/en/latest/getting-started.html#service-and-remote-users)).

We could then just focus on using a non-root service user (also suggested in the recent update) and simplify things.  The ssh library already uses the current user as the default `username` anyway as well.

Documentation surrounding the remote user (and config options) will need to be removed or modified to reference the service or current user.

kevin-bates:
We no longer document the use of `EG_REMOTE_USER` or
`--EnterpriseGatewayApp.remote_user` but instead that the _service user_
needs to be configured for password-less ssh.

In the code, the use of `EG_REMOTE_USER` and optional `EG_REMOTE_PWD`
is still present, but remote user defaults to `$USER` and password defaults
to `None`.  These merely exist in case we need to leverage this capability.

Added another example (`SSHException`) to the troubleshooting instance for
password-less ssh issues.

Fixes #198

### DIFF
--- a/docs/source/config-options.md
+++ b/docs/source/config-options.md
@@ -187,10 +187,6 @@ EnterpriseGatewayApp options
     Bracketed comma-separated list of hosts on which DistributedProcessProxy
     kernels will be launched e.g., ['host1','host2']. (EG_REMOTE_HOSTS env var -
     non-bracketed, just comma-separated)
---EnterpriseGatewayApp.remote_user=<Unicode>
-    Default: ''
-    The username used for remote operations (ssh).  Password-less ssh is
-    required.  (EG_REMOTE_USER env var)
 --EnterpriseGatewayApp.seed_uri=<Unicode>
     Default: None
     Runs the notebook (.ipynb) at the given URI on every kernel launched. No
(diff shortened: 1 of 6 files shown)


lresende:
I am fine making this an undocumented feature, but I would like to understand if there is any strong motivation to remove a functionality that is working and might be useful in a place where the user has no control or have to wait for having ssh configured?

kevin-bates:
Well, the primary issue is that when EG_REMOTE_USER differs from the "service user", ssh logins fail. So I view this as a significant simplification.

sanjay-saxena:
It would be good to add functionality based on use-cases that we understand and think are useful. Otherwise, these things can come back and haunt us later in terms of backward compatibility as we try to add legit features and our mental model now needs to take into account some legacy piece of code that is never used but nobody is sure about the implications of removing/updating it. Note that whenever we have feature(documented or not), there is always the added cost of unit tests, integration tests, and/or documentation. Just my two cents.

Issue #168, opened 2017-10-02
Address scalability issues around ssh tunneling


kevin-bates:
Since each tunneled port requires an ssh process, this might introduce scalability issues with max-process limits, etc. 

Here's some output of a system running two kernels over tunneled ports.  Note that the 6th (of 12) ssh process for each pertains to the newly introduced "gateway communication port" used for signalling (interrupt) kernels (and possibly other things).

```
elyra     5577     1  0 10:00 pts/0    00:00:01 /opt/anaconda2/bin/python /opt/anaconda2/bin/jupyter-enterprisegateway --ip=0.0.0.0 --port=8888 --port_retries=0 --log-level=DEBUG
elyra     5595     1  0 10:00 ?        00:00:00 /usr/bin/ssh -f -S none -L 127.0.0.1:60629:172.16.187.221:54073 172.16.187.221 sleep 9223372036854775807
elyra     5598     1  0 10:00 ?        00:00:00 /usr/bin/ssh -f -S none -L 127.0.0.1:44838:172.16.187.221:47418 172.16.187.221 sleep 9223372036854775807
elyra     5601     1  0 10:00 ?        00:00:00 /usr/bin/ssh -f -S none -L 127.0.0.1:38697:172.16.187.221:33526 172.16.187.221 sleep 9223372036854775807
elyra     5604     1  0 10:00 ?        00:00:00 /usr/bin/ssh -f -S none -L 127.0.0.1:36525:172.16.187.221:43768 172.16.187.221 sleep 9223372036854775807
elyra     5607     1  0 10:00 ?        00:00:00 /usr/bin/ssh -f -S none -L 127.0.0.1:58505:172.16.187.221:50969 172.16.187.221 sleep 9223372036854775807
elyra     5612     1  0 10:00 ?        00:00:00 /usr/bin/ssh -f -S none -L 127.0.0.1:57346:172.16.187.221:49737 172.16.187.221 sleep 9223372036854775807
elyra     6608     1  0 10:04 ?        00:00:00 /usr/bin/ssh -f -S none -L 127.0.0.1:39107:172.16.187.129:37925 172.16.187.129 sleep 9223372036854775807
elyra     6611     1  0 10:04 ?        00:00:00 /usr/bin/ssh -f -S none -L 127.0.0.1:58035:172.16.187.129:49791 172.16.187.129 sleep 9223372036854775807
[...]

PR #201, opened 2017-10-26
Address scalability and other issues wth SSH tunnels

sanjay-saxena:
Issue #168: Address scalability and other issues with SSH tunnels

Use `pexpect.spawn()` to spawn a child process to execute the SSH command. As a result, the SSH processes are no longer orphaned and become children of the parent Enterprise Gateway process. And, when the Kernel is being terminated, the Enterprise Gateway can terminate the child processes cleanly.

This commit also addresses issues where the number of child SSH processes fluctuated between 3 and 6. Now, Enterprise Gateway consistently creates 6 SSH tunnels to the Kernel process.

Also, added support for Win32 using Paramiko. This means, there is no need to expose any public APIs for time being till other use-cases emerge.

Updated troubleshooting guide to address SSHException when the fingerprint for ECDSA key is not added to the list of known hosts.

### DIFF
--- a/enterprise_gateway/services/processproxies/processproxy.py
+++ b/enterprise_gateway/services/processproxies/processproxy.py
@@ -11,6 +11,7 @@ import paramiko
 import logging
 import time
 import tornado
+import pexpect
 from tornado import web
 import subprocess
 from ipython_genutils.py3compat import with_metaclass
@@ -44,7 +45,6 @@ remote_user = os.getenv('EG_REMOTE_USER', os.getenv('USER'))
 remote_pwd = os.getenv('EG_REMOTE_PWD')  # this should use remote_pwd-less ssh
 
 local_ip = localinterfaces.public_ips()[0]
-tunnel_ip = localinterfaces.local_ips()[0]
 
 
 class BaseProcessProxyABC(with_metaclass(abc.ABCMeta, object)):
@@ -311,6 +311,7 @@ class RemoteProcessProxy(with_metaclass(abc.ABCMeta, BaseProcessProxyABC)):
         self.comm_ip = None
         self.comm_port = 0
         self.dest_comm_port = 0
+        self.tunnel_processes = []
         self._prepare_response_socket()
 
     def launch_process(self, kernel_cmd, **kw):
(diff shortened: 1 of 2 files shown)


enterprise_gateway/services/processproxies/processproxy.py
```diff
@@ -369,10 +371,21 @@ def _create_ssh_tunnel(self, port_name, local_port, remote_port, remote_ip, ssh_
...
         except Exception as e:
             raise RuntimeError("Could not open SSH tunnel for port {}. Exception: '{}'".format(port_name, e))
 
+    def _spawn_ssh_tunnel(self, port_name, local_port, remote_port, remote_ip, ssh_server, ssh_key=None):
+        """ This method spawns a child process to create a SSH tunnel and returns the spawned process.
+        """
+        if sys.platform == 'win32':
+            return tunnel.paramiko_tunnel(local_port, remote_port, ssh_server, remote_ip, ssh_key, timeout=maxint)
+        else:
+            cmd = "%s -S none -L 127.0.0.1:%i:%s:%i %s sleep %i" % (
```
kevin-bates: It might be good to add a comment as to why the openssh version is implemented here without the `-f` option.  That prevents the `ssh` process from getting orphaned - which invalidates any handles to that process.  In addition, the zmq implementation returns the `pid` for the process rather than the process handle itself (unlike paramiko's implementation), so this has the added benefit of consistent return values from both implementations - where we can then terminate the tunnel processes during shutdown.

kevin-bates: Since these implementations clearly use `127.0.0.1`, I think we should remove the [single use of `tunnel_ip`](
https://github.com/jupyter-incubator/enterprise_gateway/pull/201/files#diff-71b92028d190d942e7926650b50d055cL405) and replace it with `127.0.0.1` (and remove its declaration).

sanjay-saxena: Done

sanjay-saxena: Done

PR #203, opened 2017-10-29
Add support for alternate ssh ports

### LINKED ISSUE #202: Add support for alternate SSH ports
When setting up the docker image, the base image containing Spark and YARN happened to use an SSH port of 2122.  Since the SSH APIs allow alternate ports (other than 22) to be specified by appending `:<port>` to the server name, we need to be capable of doing the same.

kevin-bates:
Some installations may choose to configure the ssh server to be associated
with a port other than the default of `22`.  This change enables an alternate
port to be specified via the environment variable `EG_SSH_PORT` - which defaults
to `22`.

Updated docs with new environment variable.  Removed back-quotes for some 
entries since they don't take affect in code blocks.

Fixes #202

### DIFF
--- a/docs/source/config-options.md
+++ b/docs/source/config-options.md
@@ -247,7 +247,7 @@ JupyterWebsocketPersonality options
       stages but we intend to flip the default value to True at some point.  
         
   EG_PROXY_LAUNCH_LOG=/tmp/jeg_proxy_launch.log 
-      The log file used during remote kernel launches of `DistributedProcessProxy`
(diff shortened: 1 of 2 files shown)


kevin-bates:
@sanjay-saxena - The `server:port` format is consistent with both paramiko and openssh tunnel functions but I agree that it should probably be changed.  I'll make the changes and retest.

kevin-bates:
@lresende, is there a reason this was closed? I was in the process of applying and testing changes suggested by @sanjay-saxena.

Issue #192, opened 2017-10-17
Document that installing python modules using pip from a notebook must use --user


sanjay-saxena:
This is necessary to honor user isolation.

sanjay-saxena:
Waiting for @kevin-bates's PR to be merged as couple of files that are updated for this issue are part of the PR.

PR #204, opened 2017-10-30
Instructions to use pip with --user in multi-tenant world

sanjay-saxena:
Instructions to use pip with --user in multi-tenant world

Issue #192: In order to honor user isolation in a multi-tenant world, it is very 
important that users do not trample on each other's environments. So, if a
Python module is being installed from within a Notebook Cell, then the
`--user` command-line option should be used as shown below:

```
!pip install --user <module-name>
```

This will result in the module to be installed under 
`$HOME/.local/lib/python<version>/site-packages` folder. Also, `PYTHONPATH`
defined in `kernel.json` should include the aforementioned path so that the
module can be imported in a subsequent Notebook Cell as shown below:

```
import <module-name>
```

### DIFF
--- a/docs/source/getting-started.md
+++ b/docs/source/getting-started.md
@@ -353,6 +353,23 @@ jupyter notebook \
   --NotebookApp.kernel_spec_manager_class=nb2kg.managers.RemoteKernelSpecManager
 ```
 
+### Installing Python modules from within Notebook Cell
+To be able to honor user isolation in a multi-tenant world, installing Python modules using `pip` from
+within a Notebook Cell should be done using the `--user` command-line option as shown below:
+
+```
+!pip install --user <module-name>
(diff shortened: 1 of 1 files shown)


lresende:
@kevin-bates yes, --user is python supported and nothing something specific to Anaconda. Note that Python and Anaconda might use different locations to store user specific packages.

ckadner:
RE: @rgbkrk comment about setting an environment variable to drive the default location for `pip` installs, we could try to follow this https://pip.pypa.io/en/stable/user_guide/#environment-variables, i.e.

```
PIP_TARGET=${HOME}/.locals
```

rgbkrk:
@ckadner cool! Yeah you could stick that in the `env` key in the kernelspec then. 😄

PR #206, opened 2017-11-01
Eliminated sleep to keep SSH session alive

sanjay-saxena:
Eliminated sleep to keep SSH session alive

Earlier, `sleep maxint` was being used to keep the SSH session 
alive. However, even after the kernel process was closed/halted 
and the SSH processes were terminated, the  child processes 
would not get terminated. As a result, the number of child 
processes running under the service-user would keep 
increasing as more and more kernels were launched on the 
remote host.

This change eliminates the sleep child process as we no 
longer need it to keep the SSH session alive.

Addresses issue #168.


sanjay-saxena:
@kevin-bates -- Thanks for the review! I have updated the call to `paramiko_tunnel()` as per your suggestion. I am keeping references to `maxint` as we may need it going forward for other things such as `keepalive`.

Issue #207, opened 2017-11-01
Investigate SSH keep-alive using ServerAliveInterval and the overlap with CullingTimeout


sanjay-saxena:
EG supports the notion of `keepalive` by means of the configuration parameter `--MappingKernelManager.cull_idle_timeout`. EG recommends setting the configuration parameter to 12hours in a production environment.

SSH supports the notion of `keepalive` by means of `ClientAliveInterval`(on the server -- `/etc/ssh/sshd_config`) and `ServerAliveInterval`(on the client --`/etc/ssh/ssh_config`). By default, these parameters are not set. Since SSH runs on top of TCP, I am assuming that it falls back to TCP's `keepalive` characteristics. On RedHat 7.3 based machines, the default TCP `keepalive` is configured as shown below:

```
[root@hostname .ssh]# sysctl --all | grep keepalive
net.ipv4.tcp_keepalive_intvl = 75
net.ipv4.tcp_keepalive_probes = 9
net.ipv4.tcp_keepalive_time = 7200
```

Note that `net.ipv4.tcp_keepalive_time` is in seconds. This means, if there is no activity between the client and the server for 2hours(or 7200 seconds), TCP will initiate 9 probes that are 75seconds apart. And, if the other side does not respond to any of those probes, then TCP will drop the connection.

This ticket is to investigate SSH's `keepalive` mechanism and see how it works with EG's `keepalive` mechanism. Based on the findings, this may involve code changes as well as documentation changes. As far as I see, there are two main objectives here:

1. SSH connections should not get terminated due to inactivity while the Kernel is still running. This means that the Kernel's `cull_idle_timeout` must be greater than or equal to the `keep-alive` interval of the individual SSH connections.

2. When there is any user activity in the notebook, all six SSH connections should be able to recognize it to be able to reset the timeout. This is needed to ensure that all six SSH connections stay intact and we don't end up in a situation where some of the SSH connections get torn down as there was no traffic on them.

kevin-bates:
@sanjay-saxena - thanks for writing this up!  Yeah, I think we should set the ssh keep alive to whatever the idle timeout value is plus some factor (thinking `60` seconds).  If culling is not enabled (i.e., idle timeout is `0`) then we should probably use a `maxint` value.

sanjay-saxena:
[...]

PR #211, opened 2017-11-02
Specify ServerAliveInterval on SSH sessions

sanjay-saxena:
In order to ensure that SSH connections stay alive longer
than the Kernels, Enterprise Gateway now explicitly
specifies ServerAliveInterval option on the SSH sessions.

For CONTROL and COMMUNICATION channels, SeverAliveInterval
is currently hardcoded to the number of seconds in 100years.
For other(SHELL, IOPUB, STDIN, and HEARTBEAT) channels, the
value of ServerAliveInterval is computed by adding 60 seconds
to the value of the MappingKernelManager.cull_idle_timeout
command-line parameter. If MappingKernelManager.cull_idle_timeout
is unspecified or beyond acceptable range, then the channels
get treated like CONTROL and COMMUNICATION channels with respect
to ServerAliveInterval value.

PR is for issue #207.

### DIFF
--- a/enterprise_gateway/services/processproxies/processproxy.py
+++ b/enterprise_gateway/services/processproxies/processproxy.py
@@ -20,10 +20,7 @@ from jupyter_client import launch_kernel, localinterfaces
 from calendar import timegm
 from notebook import _tz
 from zmq.ssh import tunnel
-try:
-    from sys import maxint as maxint
-except ImportError:
-    from sys import maxsize as maxint
+from enum import Enum
 
 # Default logging level of paramiko produces too much noise - raise to warning only.
 logging.getLogger('paramiko').setLevel(os.getenv('EG_SSH_LOG_LEVEL', logging.WARNING))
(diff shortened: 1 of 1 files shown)


enterprise_gateway/services/processproxies/processproxy.py
```diff
@@ -47,6 +47,13 @@
 
 local_ip = localinterfaces.public_ips()[0]
 
+class KernelChannel(Enum):
+    SHELL = "SHELL"
+    IOPUB = "IOPUB"
+    STDIN = "STDIN"
+    HEARTBEAT = "HB"
+    CONTROL = "CONTROL"
+    COMMUNICATION = "EG_COMM"
```
kevin-bates: I think we should add a comment for the EG_COMM channel, something like...
`# Optional channel created by remote launcher used to issue remote interrupts - not a ZMQ channel`
so that its not confused as a ZMQ construct.

enterprise_gateway/services/processproxies/processproxy.py
```diff
@@ -47,6 +47,13 @@
 
 local_ip = localinterfaces.public_ips()[0]
 
+class KernelChannel(Enum):
```
kevin-bates: Nice - thanks for setting up this enum.

sanjay-saxena:
@kevin-bates -- Thanks for your feedback! Comment has been added.

PR #220, opened 2017-11-09
Refactor user documentation

lresende:
Refactor existing documentation to better express the flow
for different resource manager use cases.


docs/source/getting-started-client-mode.md
```diff
@@ -0,0 +1,98 @@
+## Enabling Client Mode/Standalone Support
+
+Even if you are not leveraging the full distributed capabilities of Jupyter Enterprise Gateway, the client mode can
+still help mitigate the resource estarvation by enabling pseudo-distributed mode where kernels are started in client
+mode in different nodes of the cluster utilizing a round-robin algorithm. In this case, you can still get bottlenecks
+in a given node that might get requests to start "large" kernels, but otherwise, you will be better of compared to when
+all kernels are stsarted from a single node or as local processes which is the default for vanilla Jupyter Notebook.
+
```
kevin-bates: Typos: 
`estarvation` -> `starvation`
`better of compared` -> `better off compared`
`stsarted` -> `started`

Suggested phrasing:
`Even if you are not leveraging the full distributed capabilities of Jupyter Enterprise Gateway, client mode can still help mitigate resource starvation by enabling a pseudo-distributed mode, where kernels are started ion different nodes of the cluster utilizing a round-robin algorithm. In this case, you can still experience bottlenecks on a given node that receives requests to start "large" kernels, but otherwise, you will be better off compared to when all kernels are started on a single node or as local processes, which is the default for vanilla Jupyter Notebook.`

docs/source/getting-started-client-mode.md
```diff
@@ -0,0 +1,98 @@
...
+require the following environment variables to be set:
+
+* SPARK_HOME: Must point to the Apache Spark installation path
+```
+SPARK_HOME:/usr/hdp/current/spark2-client                            #For HDP distribution
+```
+
+**Configuring Kernels for YARN Client mode**
+
+For each supported Jupyter Kernel, we have provided sample kernel configuration and launcher available as part of the release
```
kevin-bates: Suggested change:
...`we have provided sample kernel configurations and launchers as part`...

docs/source/getting-started-cluster-mode.md
```diff
@@ -0,0 +1,68 @@
...
+```
+
+* EG_YARN_ENDPOINT: Must point to the YARN Resource Manager endpoint
+```
+EG_YARN_ENDPOINT=http://${YARN_RESOURCE_MANAGER_FQDN}:8088/ws/v1/cluster #Common to YARN deployment
+``` 
+
+**Configuring Kernels for YARN Cluster mode**
+
+For each supported Jupyter Kernel, we have provided sample kernel configuration and launcher available as part of the release
```
kevin-bates: see previous suggestion.

docs/source/getting-started-client-mode.md
```diff
@@ -0,0 +1,98 @@
...
+    "/usr/local/share/jupyter/kernels/spark_2.1_python_yarn_client/bin/run.sh",
+    "{connection_file}",
+    "--RemoteProcessProxy.response-address",
+    "{response_address}"
+  ]
+}
+```
+
+After making any necessary adjustments such as updating SPARK_HOME or other environment specific configuration, you now should have 
+a new Kernel available which will use Jupyter Enterprise Gateway to execute your notebook cell contents.
```
kevin-bates: This needs to be mentioned somewhere in the above text.

Note: to gain complete round-robin functionality, the following steps must be taken:
1. Each of the kernelspec file-sets (`kernel.json` file and `bin` and `scripts` directories) must be copied to each of the nodes in the same folder as on the Enterprise Gateway server (e.g., `/usr/local/share/jupyter/kernels/<kernel-name>`).
2. The env variable `EG_REMOTE_HOSTS` must be set to a comma-separated set of host names indicating the hosts available for running kernels.  (This can be specified via the command line as well: `--EnterpriseGatewayApp.remote_hosts`)
3. The corresponding kernel runtime dependencies should be present on each of these nodes.

docs/getting-started-copy.md
```diff
@@ -0,0 +1,144 @@
...
+
+# OPTIONAL: check the installed R packages
+ls $ANACONDA_HOME/lib/R/library
+
+Next copy the R kernelspecs to all YARN workers
+[ ENTERPRISE_GATEWAY ] is the root directory of the JEG github repository
+cp -r [ ENTERPRISE_GATEWAY ]/etc/kernelspecs/spark_2.1_R* /usr/local/share/jupyter/kernels/
+cp -r [ ENTERPRISE_GATEWAY ]/etc/kernel-launchers/R/scripts /usr/local/share/jupyter/kernels/spark_2.1_R_yarn_client/
+cp -r [ ENTERPRISE_GATEWAY ]/etc/kernel-launchers/R/scripts /usr/local/share/jupyter/kernels/spark_2.1_R_yarn_cluster/
+
```
kevin-bates: The above should be replaced with reference to the `jupyter_enterprise_gateway_kernelspecs-<release>.tar.gz` as is the case for the `scala kernel` section.  No need to pull from the "JEG" repo.

Also, shouldn't the `ipython kernel` section be updated in a similar manner?

Btw, the scala section only mentions `cluster` and not `client` mode.

docs/source/getting-started-client-mode.md
```diff
@@ -94,5 +102,29 @@ Please see below how a kernel.json would look like for integrating with Spark St
...
+    "SPARK_YARN_USER_ENV": "PYTHONUSERBASE=/home/yarn/.local,PYTHONPATH=${HOME}/.local/lib/python3.6/site-packages:/usr/hdp/current/spark2-client/python:/usr/hdp/current/spark2-client/python/lib/py4j-0.10.4-src.zip,PATH=/opt/anaconda2/bin:$PATH",
+    "SPARK_OPTS": "--master spark://127.0.0.1:7077  --name ${KERNEL_ID:-ERROR__NO__KERNEL_ID} --conf spark.yarn.submit.waitAppCompletion=false",
+    "LAUNCH_OPTS": ""
+  },
+  "argv": [
+    "/usr/local/share/jupyter/kernels/spark_2.1_python_yarn_client/bin/run.sh",
+    "{connection_file}",
+    "--RemoteProcessProxy.response-address",
+    "{response_address}"
+  ]
```
kevin-bates: We should remove the `response-address` entry (and its preceding comma)...
```
,
   "--RemoteProcessProxy.response-address",
   "{response_address}"
```

They probably shouldn't configure the launchers in the run.sh - so this gets a little messy.  Might be best just to say "standard kernelspec configurations" - rather than implying the use of our sample kernelspecs.

lresende: Good point, I removed the section and just add a new paragraph into the introduction of the section, please check if it looks better now.

PR #224, opened 2017-11-13
Add toree-launcher to enterprise-gateway repo

### LINKED ISSUE #223: Get toree-launcher (scala) code into repo
It would be good to have all the launcher code files in the repo.  The python and R kernel launchers are scripts, so they aren't issues.  The toree launcher code is scala code so we've been more hesitant.  However, since the launchers are already separated and we build using make, we should be able to have the make file invoke `sbt` to perform the appropriate actions for the toree launcher jar.  This way all examples can be fully defined.

Although not required, it would be good if this could get into the v0.7 release.

kevin-bates:
Moved the toree-launcher from a private repo into EG. As a result, performing
`make kernelspecs` now requires the installation of `sbt` - which will build
the toree launcher jar and place it into the `etc/kernel-launchers/scala/lib`
directory (where it was located prior to this change).  From there, the build
then includes it when archiving the kernelspecs into a tar file.

The file name for the torre-launcher has been altered from `toree-launcher_2.11-1.0.jar`
to `toree-launcher_2.11-0.7.0.dev0.jar` (or whatever EG version is configured).

Since there is now a `Makefile` in the `etc` directory to logically split the
building of the kernelspecs to their repo location, the similar split has taken
place for the docker images.  All targets are still accepted by the primary `Makefile`,
just that the `kernelspecs` and `docker *` targets get propagated to `etc/Makefile`.

Fixes #223


etc/kernel-launchers/scala/toree-launcher/build.sbt
```diff
@@ -0,0 +1,20 @@
+/*
+ * Copyright (c) Jupyter Development Team.
+ * Distributed under the terms of the Modified BSD License.
+ */
+name := "toree-launcher"
+
+version := sys.props.getOrElse("version", default = "1.0")
+
```
lresende: This will bring the dev0 as part of the artifact version which is really a Python convention, should we use replace dev0 for SNAPSHOT which then is more towards the Java/Scala maven convention ? Changing the line above to the following should solve the problem

version := sys.props.getOrElse("version", default = "1.0").replaceAll("dev[0-9]", "SNAPSHOT")

kevin-bates:
Good point @lresende.  I've updated the sbt build and the Makefile accordingly.

lresende:
LGTM, thanks.

Issue #225, opened 2017-11-13
Keep Toree Launcher in sync with Apache Toree


sanjay-saxena:
It would be a good idea to keep both the launcher and the kernel in sync or consistent with respect to various tools such as SBT.

Currently, Apache Toree is using `SBT 1.0.3` and Toree Launcher that is built as part of Enterprise Gateway is based on `SBT 0.13.13`. Keeping them in sync will make our lives convenient. Otherwise, we will need to have multiple versions of SBTs installed and switch them based on the environment.

It would also be a good idea to make sure that they are both based on the same Scala version as well. Apache Toree is currently based on `Scala 2.11`.

PR #226, opened 2017-11-13
Change sbt.version of Toree Launcher to 1.0.3

liukun1016:
Small fix. Related to issue #225

### DIFF
--- a/docs/source/devinstall.md
+++ b/docs/source/devinstall.md
@@ -83,7 +83,7 @@ make bdist
 Enterprise Gateway includes two sets of kernelspecs for each of the three primary kernels: `IPython`,`IR`, 
 and `Toree` to demonstrate remote kernels and their corresponding launchers.  One set uses the 
 `DistributedProcessProxy` while the other uses  the `YarnClusterProcessProxy`. The following makefile 
-target produces a tar file (`enterprise_gateway_kernelspecs.tar.gz`) in the `dist` directory.
+target produces a tar file (`enterprise_gateway_kernelspecs.tar.gz`) in the `dist` directory. (We are using `sbt` for the build so you might need to install/upgrade it.)
 
 ```
 make kernelspecs
--- a/docs/source/devinstall.md
+++ b/docs/source/devinstall.md
@@ -83,7 +83,9 @@ make bdist
 Enterprise Gateway includes two sets of kernelspecs for each of the three primary kernels: `IPython`,`IR`, 
 and `Toree` to demonstrate remote kernels and their corresponding launchers.  One set uses the 
 `DistributedProcessProxy` while the other uses  the `YarnClusterProcessProxy`. The following makefile 
-target produces a tar file (`enterprise_gateway_kernelspecs.tar.gz`) in the `dist` directory. (We are using `sbt` for the build so you might need to install/upgrade it.)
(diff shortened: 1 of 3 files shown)


kevin-bates:
@sanjay-saxena @liukun1016 - yeah, that's probably a good idea.  Not so much about the specific version (since it auto-upgrades) but we should probably mention that `sbt` is used to build the kernelspecs tar file - so I'd add it where we mention `make kernelspecs`.

sanjay-saxena:
@kevin-bates -- On RedHat 7.3, `sbt` did not auto-install for me. Instead, I got this error message:

```
....
mkdir -p kernel-launchers/scala/lib
/bin/bash: sbt: command not found
cp: cannot stat ‘target/scala-2.11/*.jar’: No such file or directory
make[1]: *** [kernel-launchers/scala/lib] Error 1
make[1]: Leaving directory `/root/Development/enterprise_gateway/etc'
make: *** [kernelspecs] Error 2
```

liukun1016:
@kevin-bates @sanjay-saxena 
Thanks Kevin and Sanjay!

@kevin-bates 
By "I'd add it where we mention `make kernelspecs`", do you mean to add it on your own or want me to do the change?

kevin-bates:
@liukun1016 I was suggesting where the comment should be added, but didn't mean to imply that I was going to make the change.

liukun1016:
@kevin-bates @sanjay-saxena 

Comment added. Let me know if it is ok to merge this PR.

docs/source/devinstall.md
```diff
@@ -83,7 +83,7 @@ make bdist
 Enterprise Gateway includes two sets of kernelspecs for each of the three primary kernels: `IPython`,`IR`, 
 and `Toree` to demonstrate remote kernels and their corresponding launchers.  One set uses the 
 `DistributedProcessProxy` while the other uses  the `YarnClusterProcessProxy`. The following makefile 
-target produces a tar file (`enterprise_gateway_kernelspecs.tar.gz`) in the `dist` directory.
+target produces a tar file (`enterprise_gateway_kernelspecs.tar.gz`) in the `dist` directory. (We are using `sbt` for the build so you might need to install/upgrade it.)
```
kevin-bates: I'd rather not use the parenthesis and be a little more specific.  How about something along these lines (sans italics) *after* the `make kernelspecs` entry ...

_"Note: Because the scala launcher requires a jar file, `make kernelspecs` requires the use of `sbt` to build the scala launcher jar.  Please consult the [sbt site](http://www.scala-sbt.org/) for directions to install/upgrade `sbt` on your platform.  We currently prefer the use of `1.0.3`."_

liukun1016: Much better

kevin-bates:
Sorry - but do you mind moving this **after** the command?  The reason is because in the previous statement we say "the following command", but in the text about `sbt` we refer to `make kernelspecs` before its "defined".   Sorry, I meant to bold **after** but had only used one asterisks.

If you'd rather not, this isn't a big deal.

liukun1016:
@kevin-bates no problem. A small change.

liukun1016:
Thanks @kevin-bates @sanjay-saxena 

Merge it now.

Issue #161, opened 2017-09-25
Add proxy configuration values to kernelspec


kevin-bates:
By extending the `process_proxy` stanza within the kernelspec, I think we could achieve some significant functional enhancements that make for a more flexible, finer-grained configuration.  The proposal would be to define a `proxy_config` stanza within `process_proxy` (and sibling to `class_name`) which contains a set of fields pertinent to the target process proxy class (and its super-classes).  Which fields are pertinent to which classes would need to be documented - along with their defaulted values (if not already).

A good example of this would be to have `yarn_endpoint` be a field that pertains to `YarnClusterProcessProxy`.  When present, kernels derived from that kernelspec would use that yarn endpoint for their activity.  This would allow administrators to utilize multiple Yarn clusters from a single Enterprise Gateway.

We could also do the same with remote hosts used to balance loads of `DistributedProcessProxy` kernels where sets of kernels are essentially assigned different sets of machines.

This would also allow greater flexibility to process proxy authors that require different (undetermined) kinds of information.

lresende:
@kevin-bates aren't we done with this ?

kevin-bates:
@lresende - no, this hasn't been implemented yet.

PR #227, opened 2017-11-14
Remove spark version indicator from kernelspecs

### LINKED ISSUE #174: Make Enterprise Gateway kernelspecs independent of Spark version
Currently, the folder name and display name for kernelspec have hardcoded Spark version

kevin-bates:
Removed references to the spark version (`2.1`) from the kernelspecs (display
names and echo statements), renamed directories accordingly, and updated like
references in the docs.

Also, updated the doc reference to the `0.6.0` version of the kernelspecs tar
file to `0.7.0` since we will be cutting a release later this week.  As a result,
there may be a brief window where that corresponding link is invalid.

Fixes #174

### DIFF
--- a/docs/source/getting-started-client-mode.md
+++ b/docs/source/getting-started-client-mode.md
@@ -32,7 +32,7 @@ For each supported Jupyter Kernel, we have provided sample kernel configurations
 [e.g. jupyter_enterprise_gateway_kernelspecs-0.6.0.tar.gz](https://github.com/jupyter-incubator/enterprise_gateway/releases/download/v0.6.0/jupyter_enterprise_gateway_kernelspecs-0.6.0.tar.gz).
 
 Considering we would like to enable the iPython Kernel that comes pre-installed with Anaconda to run on
-Yarn Client mode, we would have to copy the sample configuration folder **spark_2.1_python_yarn_client**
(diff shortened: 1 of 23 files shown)


rgbkrk:
Was this primarily for demo purposes to simplify what spark shows as?

On our deployment (of classic notebook server) we keep the version of spark as part of it so that people can pin to exactly the version of spark they're working with. Would you generally be advising against that?

kevin-bates:
@rgbkrk - Yes - this was primarily to simply things from an example perspective.  

We too have found a tendency to include the spark version indicator in display names - which is fine - despite the fact that clusters don't tend to vary wrt spark versions.  In addition, things can get a little messy when trying to keep track of version idiosyncrasies and that's something we'd rather not get caught up maintaining.  That said, one of things I'd like to add is the ability to target different YARN endpoints at the kernel level (Issue #161) - so, in cases where those endpoints correspond to different versions, it might behoove the user to add indicators, etc.

I'm curious about your question: _"...to simplify what spark shows as?"_  Did you mean to say 'what the notebook shows as' (i.e., the kernelspec display name)?  Or is there some other display value that is typical in these envs?

rgbkrk:
Yeah, I mean "what the notebook shows as" within demo purposes. It sounds like this is exactly that. Thanks for the answer!

lresende:
LGTM

Issue #153, opened 2017-09-20
Release at conda-forge


lresende:
https://github.com/conda-forge/staged-recipes

liukun1016:
Here is the PR created for this issue on the conda-forge side:
https://github.com/conda-forge/staged-recipes/pull/4376

liukun1016:
The conda recipe for kernel gateway as a reference:
https://github.com/conda-forge/jupyter_kernel_gateway-feedstock/blob/master/recipe/meta.yaml

liukun1016:
The PR passed all the test cases.

But we need to first upload our 0.7.0 tar file to Pypi before we release the 0.7.0 version on Conda, since the PR would need to download the tar file with the corresponding version before build it.

liukun1016:
Need to add `pexpect` as a required lib inside the `requirements-conda.txt` since it is not in the conda side.

Here is the link to the build where the `pexpect` lib was not listed as one of the `requirements` in `meta.yaml` and thus an error of `ImportError`:
https://ci.appveyor.com/project/conda-forge/staged-recipes/build/1.0.18805/job/8jcd6gm8hhkcw6e3#L551


Going to open a PR for this, and as long as it is merged, we can close this issue.

liukun1016:
The PR was merged and we could now use `conda install jupyter_enterprise_gateway` to install our 0.7 version. So now close this issue.

PR #232, opened 2017-11-22
Add pexpect as requirement for conda

liukun1016:
For issue #153


kevin-bates:
@liukun1016 - I think this should be added to `requirements.txt` as well since its a general dependency and its addition to that file will remove the warning that is produced in the IDE:
```
Package 'pexpect' is not listed in project requirements
```

liukun1016:
Got an error when testing test_jupyter_websocket.py:
https://travis-ci.org/jupyter-incubator/enterprise_gateway/jobs/308050794#L895

Issue #30, opened 2017-06-07
Replicate integration tests infrastructure for all Kernels


lresende:
Enable basic integration tests for all supported kernels using docker

liukun1016:
The PR #245 was merged. Now close this issue.

PR #245, opened 2017-12-07
Integration test enhancement

liukun1016:
Related to issue #30 

Currently the content of each language's notebook is not consistent with other language, and in order to make us easier to compare the testing for each kernel, I normalized those notebooks so their contents are almost the same (Except for Scala kernels since we need to test `Toree magic`).

Also, now itest supports printing each code-cell/step of each notebook so when get stuck, we could detect which code cell might be the reason.

### DIFF
--- a/enterprise_gateway/itests/src/elyra_client.py
+++ b/enterprise_gateway/itests/src/elyra_client.py
@@ -82,7 +82,8 @@ class ElyraClient(object):
             code_cell_count = 1
             for code_cell in nb_code_entity.get_all_code_cell():
                 if code_cell.is_executed():
-                    print("---{})\n {}".format(code_cell_count, code_cell.get_source_for_execution()))
+                    print("---{}) {}\n{}".format(code_cell_count, code_cell,
+                                                 code_cell.get_source_for_execution()))
                     code_cell_count += 1
                     code_source = code_cell.get_source_for_execution()
                     ws.send(ElyraClient.new_code_message(code_source))
--- a/enterprise_gateway/itests/src/elyra_client.py
+++ b/enterprise_gateway/itests/src/elyra_client.py
@@ -1,6 +1,6 @@
 from tornado.escape import json_encode, json_decode, url_escape
 from uuid import uuid4
-from nb_entity import NBCodeCell, NBCodeEntity
+from nb_entity import NBCodeCell
 import websocket
 from collections import deque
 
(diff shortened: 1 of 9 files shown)


enterprise_gateway/itests/src/elyra_client.py
```diff
@@ -76,11 +75,17 @@ def receive_target_messages(ws, target_msg_type_queue):
...
+        ws_url = self.get_ws_kernel_endpoint(nb_code_entity.kernel_id)
+        ws = websocket.create_connection(url=ws_url)
+        print("Connection created for web socket {}".format(ws_url))
         message_code_cell_list = list([])
         try:
+            code_cell_count = 1
             for code_cell in nb_code_entity.get_all_code_cell():
                 if code_cell.is_executed():
+                    print("---{}) {}".format(code_cell_count, code_cell))
+                    print("{}".format(code_cell.get_source_for_execution()))
```
kevin-bates: Can you please provide an example of what a test's output would look like across a couple of cell executions?  I'd prefer to keep the output of each cell to one line and this looks like it would produce two lines. 

What is the difference between `code_cell` and `code_cell.get_source_for_execution()`?  I guess I was thinking only the latter would be sufficient (alone with `code_cell_count`).

liukun1016: Here is an example:

```
1. Kernel ID: 461eb9a4-c793-49f7-948d-cd6df8cb1c46, Notebook: Python_Client1.ipynb, Kernel spec: spark_python_yarn_client, Number of code cells: 6
Connection created for web socket ws://9.30.230.137:8888/api/kernels/461eb9a4-c793-49f7-948d-cd6df8cb1c46/channels
---1) Execution_count 1, 1 lines of codes, 1 outputs
print("Hello World")
---2) Execution_count 2, 2 lines of codes, 1 outputs
# DIFFERENT
print sc.applicationId
---3) Execution_count 3, 2 lines of codes, 1 outputs
# DEPENDS
sc.version
---4) Execution_count 4, 1 lines of codes, 1 outputs
sc.getConf().get("spark.master")
---5) Execution_count 5, 1 lines of codes, 1 outputs
sc.getConf().get("spark.submit.deployMode")
---6) Execution_count 6, 2 lines of codes, 1 outputs
# DEPENDS
print sc.getConf().get("spark.driver.host")

Finish execution of codes, now compare/assert
Testing completed: Kernel ID: 461eb9a4-c793-49f7-948d-cd6df8cb1c46, Notebook: Python_Client1.ipynb, Kernel spec: spark_python_yarn_client, Number of code cells: 6
```

And yes it would be more clear if only the real code cells would be printed, like:

```
---1) 
print("Hello World")
---2)
# DIFFERENT
print sc.applicationId
```

Issue #238, opened 2017-12-04
Create user impersonation integration tests


lresende:
Enable user impersonation via EG_IMPERONSATION_ENABLED and make sure !whoami result is same as KERNEL_USERNAME

liukun1016:
Currently I have two ideas about the implementation of this issue:

1. add a `!whoami` code cell on one of the python notebooks (client and/or cluster mode), and use a flag argument that when it is specified as e.g. `bob` instead of `None`, compare the result from the `whoami` cell with `bob`

So this way requires changes on both notebook side (just one more code cell) and some logic on the itest src codes (e.g. another argument like `impersonation_username`)

2. add a function to compare the flag argument username when it is specified, by sending a code message `whoami` to a python kernel by socket. 

So this way requires more changes on the src codes side

PR #246, opened 2017-12-08
First version implementation of impersonation integration test

liukun1016:
Related to issue #238 

Now we could do impersonation test by providing the `impersonation_username` flag with the target username, e.g. `bob`. Here is the logic:

1. For python client notebook, the newly added code cell `!whoami` would be executed
2. The first line of that code cell should contain `IMPERSONATION`, so the assert code would be returned as 3
3. If the assert code is 3, and the variable `impersonation_username` is not `None` in Python, then the output of the `!whoami` code cell would be compared with the `impersonation_username`
4. If they are the same, then it is a pass test case

Some other things to notice here:
1) It actually doesn't matter what output would be returned from the code cell of `!whoami` on the python client notebook file. The only thing matters here is if the `impersonation_username` is the same as the output or not. Hence, unlike the case of assert code 0 and 1, there is no comparison between the `test_output_str` and `real_output_str`.
2) Since this only needs to be done once for all notebook kernels, it is redundant to do the same test in either R or scala kernel. So here only the Python client notebook has this `!whoami` code cell. If we want to add this code cell to R or scala kernel anyway, might need to change the syntax e.g. in Toree scala using Toree magic (if provided).

### DIFF
--- a/enterprise_gateway/itests/src/itest_notebook.py
+++ b/enterprise_gateway/itests/src/itest_notebook.py
@@ -6,11 +6,12 @@ import sys
 
 class NotebookTestCase(TestCase):
 
-    def __init__(self, test_method, nb_entities_list, continue_when_error, host, username):
+    def __init__(self, test_method, nb_entities_list, continue_when_error, host, username, impersonation_username):
         TestCase.__init__(self, methodName=test_method)
         self.nb_entities_list = nb_entities_list
         self.continue_when_error = continue_when_error
         self.elyra_client = ElyraClient(host=host, username=username)
+        self.impersonation_username = impersonation_username
 
     @staticmethod
     def get_assert_code(first_line_code):
@@ -19,12 +20,15 @@ class NotebookTestCase(TestCase):
         0 : must be exactly the same, i.e. using assertEqual (default)
         1 : OK if test output not the same as input, i.e. ignore assert as long as no error
         2 : must not be the same as each time the execution will definitely be different, then use assertNotEqual
+        3 : Impersonation test, i.e. compare the output with the self.impersonation_username if it is not None
         """
         if first_line_code:
             if first_line_code.find("DIFFERENT") > 0:
                 return 1
             elif first_line_code.find("DEPENDS") > 0:
                 return 2
+            elif first_line_code.find("IMPERSONATION")> 0:
+                return 3
         return 0
 
     def execute_codes(self, nb_code_entity):
@@ -55,21 +59,28 @@ class NotebookTestCase(TestCase):
                 test_output = test_code_cell_list[index]
                 self.assertIsNotNone(test_output)
                 assert_code = NotebookTestCase.get_assert_code(real_code_cell.get_first_line_code())
-                test_output = test_output.seralize_output()
-                real_output = real_code_cell.seralize_output()
+                test_output_str = test_output.seralize_output()
+                real_output_str = real_code_cell.seralize_output()
                 try:
                     if assert_code != 2:
                         if assert_code == 0:
-                            self.assertEqual(test_output, real_output)
+                            self.assertEqual(test_output_str, real_output_str)
                         elif assert_code == 1:
-                            self.assertNotEqual(test_output, real_output)
+                            self.assertNotEqual(test_output_str, real_output_str)
+                        elif assert_code == 3 and self.impersonation_username is not None:
+                            # Do impersonation test if and only if the first line is IMPERSONATION (assert code = 3)
+                            # and self.impersonation_username is not None
+                            test_username = test_output.code_output_list[0].raw_output.get('text')
+                            self.assertIsNotNone(test_username)
+                            test_username = str(test_username).replace("\r\n", "")
+                            self.assertEqual(test_username, self.impersonation_username)
                 except Exception as e:
                     print("===================================")
                     print(e.message)
                     print(traceback.format_exc())
                     print("{}\nExecute count {}".format(nb_code_entity, real_code_cell.execute_count))
-                    print("[Test output]\n", test_output)
-                    print("[Real output]\n", real_output)
+                    print("[Test output]\n", test_output_str)
+                    print("[Real output]\n", real_output_str)
                     if not self.continue_when_error:
                         sys.exit(-1)
             index += 1
--- a/enterprise_gateway/itests/src/itest_notebook.py
+++ b/enterprise_gateway/itests/src/itest_notebook.py
@@ -6,12 +6,13 @@ import sys
 
 class NotebookTestCase(TestCase):
 
-    def __init__(self, test_method, nb_entities_list, continue_when_error, host, username, impersonation_username):
+    def __init__(self, test_method, nb_entities_list, **kwargs):
         TestCase.__init__(self, methodName=test_method)
         self.nb_entities_list = nb_entities_list
-        self.continue_when_error = continue_when_error
-        self.elyra_client = ElyraClient(host=host, username=username)
-        self.impersonation_username = impersonation_username
+        self.continue_when_error = kwargs.get("continue_when_error")
+        self.username = kwargs.get("username")
+        self.elyra_client = ElyraClient(host=kwargs.get("host"), username=self.username)
+        self.enforce_impersonation = kwargs.get("enforce_impersonation")
 
     @staticmethod
     def get_assert_code(first_line_code):
@@ -20,7 +21,7 @@ class NotebookTestCase(TestCase):
         0 : must be exactly the same, i.e. using assertEqual (default)
         1 : OK if test output not the same as input, i.e. ignore assert as long as no error
         2 : must not be the same as each time the execution will definitely be different, then use assertNotEqual
-        3 : Impersonation test, i.e. compare the output with the self.impersonation_username if it is not None
+        3 : Impersonation test, i.e. compare the output username if the self.enforce_impersonation is set as True
         """
         if first_line_code:
             if first_line_code.find("DIFFERENT") > 0:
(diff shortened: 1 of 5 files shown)


enterprise_gateway/itests/src/itest_notebook.py
```diff
@@ -55,21 +59,28 @@ def execute_and_assert(self, nb_code_entity):
...
+                            self.assertEqual(test_output_str, real_output_str)
                         elif assert_code == 1:
-                            self.assertNotEqual(test_output, real_output)
+                            self.assertNotEqual(test_output_str, real_output_str)
+                        elif assert_code == 3 and self.impersonation_username is not None:
+                            # Do impersonation test if and only if the first line is IMPERSONATION (assert code = 3)
+                            # and self.impersonation_username is not None
+                            test_username = test_output.code_output_list[0].raw_output.get('text')
+                            self.assertIsNotNone(test_username)
+                            test_username = str(test_username).replace("\r\n", "")
```
kevin-bates: I'm surprised this call to `replace()` isn't necessary for the `assert_code == 0` case.  Why can't you just do the comparison against `test_output_str`?  Did you find that `whoami` always concatenates `\r\n`?

We might want to split the replace call into two, one for `\r` and another for `\n`, because I could see this behavior being different depending on the flavor of the OS.

enterprise_gateway/itests/src/main.py
```diff
@@ -14,6 +14,7 @@ def parse_arg():
     parser.add_argument('--continue_when_error', default=True)
     parser.add_argument('--host', default='localhost:8888')
     parser.add_argument('--username', default='root')
+    parser.add_argument('--impersonation_username', default=None)
```
kevin-bates: I prefer we not introduce another `username` parameter.  Can we just use `username` and change `--impersonation_username` be a boolean flag that indicates that impersonation be enforced?  (e.g., `--enforce_impersonation`)
`username` will ALWAYS equal `impersonation_username`, so the latter is redundant.

enterprise_gateway/itests/src/main.py
```diff
@@ -39,7 +40,8 @@ def init_nb_test_case(args):
                 if not target_kernels or nb_entity.kernel_spec_name in target_kernels:
                     nb_entities_list.append(nb_entity)
 
-    return NotebookTestCase(method, nb_entities_list, args.continue_when_error, args.host, args.username)
+    return NotebookTestCase(method, nb_entities_list, args.continue_when_error,
+                            args.host, args.username, args.impersonation_username)
```
kevin-bates: Let's just pass args, and let NotebookTestCase decide what it wants to pull out.  This way we don't modify the signature for each new argument.  Seems like we could use the `**kwargs` convention here.

enterprise_gateway/itests/src/main.py
kevin-bates: Not sure how `usename` is used, but if it's purely there to set into `KERNEL_USERNAME` or not generally used, then we should probably default it to `None` and make the setting of `KERNEL_USERNAME` conditional to mimic users not setting `KERNEL_USERNAME`.

enterprise_gateway/itests/src/main.py
liukun1016: Good point. Will make the change.

enterprise_gateway/itests/src/main.py
liukun1016: The `username` parameter is only used as `KERNEL_USERNAME` when a new kernel is to be created:
https://github.com/jupyter-incubator/enterprise_gateway/blob/master/enterprise_gateway/itests/src/elyra_client.py#L31

And yes I'd better change the argument's name as `enforce_impersonation`, a boolean flag. When it is set as `True`, the `username` and the test output username should be compared to verify if it is a valid result of impersonation.

Will make the change.

liukun1016: I would recommend setting the default value as a non-None value e.g. `username` as `root` or `elyra`.

Based on my observation, it is ok to open new kernels in client mode with `None` as the `KERNEL_USERNAME`/`username`. But for all kernels in cluster mode, there would be an error like this:

```
[D 2017-12-08 11:28:15.887 EnterpriseGatewayApp] BaseProcessProxy.launch_process() env: {u'EG_ALTERNATE_SIGINT': u'SIGUSR2', u'__TOREE_SPARK_OPTS__': u'--master yarn --deploy-mode cluster --name ${KERNEL_ID:-ERROR__NO__KERNEL_ID} --conf spark.yarn.submit.waitAppCompletion=false --conf spark.yarn.am.waitTime=1d', u'SPARK_YARN_USER_ENV': u'TOREE_ALTERNATE_SIGINT=USR2', 'KERNEL_ID': u'4def8b26-1045-4987-9f4f-edabda6fc199', u'SPARK_HOME': u'/usr/hdp/current/spark2-client', 'KERNEL_GATEWAY': '1', u'TOREE_ALTERNATE_SIGINT': u'USR2', u'__TOREE_OPTS__': u'', u'KERNEL_USERNAME': None, u'DEFAULT_INTERPRETER': u'Scala', 'PATH': '/opt/anaconda2/bin:/usr/lib64/qt-3.3/bin:/opt/anaconda2/bin:/usr/local/bin:/bin:/usr/bin:/usr/local/sbin:/usr/sbin:/home/elyra/.local/bin:/home/elyra/bin', u'LAUNCH_OPTS': u''}
[E 2017-12-08 11:28:15.896 EnterpriseGatewayApp] Failed to run command:
    [u'/usr/local/share/jupyter/kernels/spark_scala_yarn_cluster/bin/run.sh', u'--profile', u'/home/elyra/.local/share/jupyter/runtime/kernel-4def8b26-1045-4987-9f4f-edabda6fc199.json', u'--RemoteProcessProxy.response-address', u'172.16.155.185:44558']
        PATH='/opt/anaconda2/bin:/usr/lib64/qt-3.3/bin:/opt/anaconda2/bin:/usr/local/bin:/bin:/usr/bin:/usr/local/sbin:/usr/sbin:/home/elyra/.local/bin:/home/elyra/bin'
        with kwargs:
    {'cwd': None, 'stdin': -1, 'preexec_fn': <function <lambda> at 0x7f6dd01c2668>, 'stderr': None, 'stdout': None}
    
[E 2017-12-08 11:28:15.897 EnterpriseGatewayApp] Unhandled error in API request
      File "/opt/anaconda2/lib/python2.7/site-packages/notebook/base/handlers.py", line 540, in wrapper
        result = yield gen.maybe_future(method(self, *args, **kwargs))
      File "/opt/anaconda2/lib/python2.7/site-packages/tornado/gen.py", line 1055, in run
        value = future.result()
      File "/opt/anaconda2/lib/python2.7/site-packages/tornado/concurrent.py", line 238, in result
        raise_exc_info(self._exc_info)
      File "/opt/anaconda2/lib/python2.7/site-packages/tornado/gen.py", line 1063, in run
        yielded = self.gen.throw(*exc_info)
      File "/opt/anaconda2/lib/python2.7/site-packages/notebook/services/kernels/handlers.py", line 49, in post
        kernel_id = yield gen.maybe_future(km.start_kernel(kernel_name=model['name']))
      File "/opt/anaconda2/lib/python2.7/site-packages/tornado/gen.py", line 1055, in run
        value = future.result()
      File "/opt/anaconda2/lib/python2.7/site-packages/tornado/concurrent.py", line 238, in result
        raise_exc_info(self._exc_info)
      File "/opt/anaconda2/lib/python2.7/site-packages/tornado/gen.py", line 1063, in run
        yielded = self.gen.throw(*exc_info)
      File "/opt/anaconda2/lib/python2.7/site-packages/enterprise_gateway/services/kernels/remotemanager.py", line 28, in start_kernel
        kernel_id = yield gen.maybe_future(super(RemoteMappingKernelManager, self).start_kernel(*args, **kwargs))
      File "/opt/anaconda2/lib/python2.7/site-packages/tornado/gen.py", line 1055, in run
        value = future.result()
      File "/opt/anaconda2/lib/python2.7/site-packages/tornado/concurrent.py", line 238, in result
        raise_exc_info(self._exc_info)
      File "/opt/anaconda2/lib/python2.7/site-packages/tornado/gen.py", line 1063, in run
        yielded = self.gen.throw(*exc_info)
      File "/opt/anaconda2/lib/python2.7/site-packages/kernel_gateway/services/kernels/manager.py", line 81, in start_kernel
        kernel_id = yield gen.maybe_future(super(SeedingMappingKernelManager, self).start_kernel(*args, **kwargs))
      File "/opt/anaconda2/lib/python2.7/site-packages/tornado/gen.py", line 1055, in run
        value = future.result()
      File "/opt/anaconda2/lib/python2.7/site-packages/tornado/concurrent.py", line 238, in result
        raise_exc_info(self._exc_info)
      File "/opt/anaconda2/lib/python2.7/site-packages/tornado/gen.py", line 307, in wrapper
        yielded = next(result)
      File "/opt/anaconda2/lib/python2.7/site-packages/notebook/services/kernels/kernelmanager.py", line 123, in start_kernel
        super(MappingKernelManager, self).start_kernel(**kwargs)
      File "/opt/anaconda2/lib/python2.7/site-packages/jupyter_client/multikernelmanager.py", line 110, in start_kernel
        km.start_kernel(**kwargs)
      File "/opt/anaconda2/lib/python2.7/site-packages/enterprise_gateway/services/kernels/remotemanager.py", line 100, in start_kernel
        return super(RemoteKernelManager, self).start_kernel(**kw)
      File "/opt/anaconda2/lib/python2.7/site-packages/jupyter_client/manager.py", line 257, in start_kernel
        **kw)
      File "/opt/anaconda2/lib/python2.7/site-packages/enterprise_gateway/services/kernels/remotemanager.py", line 126, in _launch_kernel
        return self.process_proxy.launch_process(kernel_cmd, **kw)
      File "/opt/anaconda2/lib/python2.7/site-packages/enterprise_gateway/services/processproxies/yarn.py", line 57, in launch_process
        self.local_proc = launch_kernel(kernel_cmd, **kw)
      File "/opt/anaconda2/lib/python2.7/site-packages/jupyter_client/launcher.py", line 128, in launch_kernel
        proc = Popen(cmd, **kwargs)
      File "/opt/anaconda2/lib/python2.7/subprocess.py", line 390, in __init__
        errread, errwrite)
      File "/opt/anaconda2/lib/python2.7/subprocess.py", line 1024, in _execute_child
        raise child_exception
    TypeError: execve() arg 3 contains a non-string value
[E 171208 11:28:15 web:2063] 500 POST /api/kernels (9.98.25.12) 21.20ms
```

kevin-bates: @liukun1016 - this issue looks familiar and something I believe I addressed in a recent change.  Are you sure you're running against the latest version?  At least the version that adds support for impersonation and authorization?

Issue #242, opened 2017-12-05
Spark R client mode kernel's app name not same as kernel ID


liukun1016:
Currently if a new Spark R client mode notebook is opened, and the SparkContext gets created, on the YARN RM UI side, the application name would be `SparkR` instead the kernel ID.

Here is an example: http://charger-master.fyre.ibm.com:8088/cluster/app/application_1512070656800_0072

akchinSTC:
PR open

PR #248, opened 2017-12-12
Fix R-kernel to show kernel ID as app name in YARN RM

akchinSTC:
Updated the R launcher script to include Kernel ID instead of "SparkR" as application name in YARN resource manager


kevin-bates:
Fixes #242

PR #268, opened 2018-02-09
ISSUE-263 - Initial changes of itests to use pytest [WIP]

akchinSTC:
-Switched test framework from untittest to pytest, still retains backwards compatibility with unittest format tests. 
- Replaced much of nb_entity.NBCodeEntity with a run_cell() method, so you can run individual tests as needed. 
- Elyra client is now static class
- Removed some unneeded comments from the test notebooks
- to run use following 
`pytest -s <python testcase> --host <gateway host:port> --username <user> --impersonation <true|false>`

Things to do:
- Make the test assert regex more accurate and finer grain for each test
- add impersonation tests


enterprise_gateway/itests/src/R_client_testcase_pytest.py
```diff
@@ -0,0 +1,47 @@
+from elyra_client import ElyraClient
```
kevin-bates: Is the suffix of `_testcase_pytest` a convention used by the framework?  If not, I suggest shortening it.  Seems like `_pytest` is a little more redundant than `_testcase`, so I suppose the latter could be retained.

requirements.txt
```diff
@@ -9,3 +9,5 @@ paramiko>=2.1.2
 yarn-api-client>=0.2.3
 pexpect>=4.2.0
 pycrypto>=2.6.1
+pytest>=3.3.2
```
kevin-bates: I thought test dependencies were supposed to go into `requirements-test.txt`.  Is there a compelling reason for having these here?  They will end up getting loaded as part of the general pip installs otherwise.

enterprise_gateway/itests/src/R_client_testcase_pytest.py
akchinSTC: Negative, just a name I came up with to distinguish between older unittest files and newer pytest files. Will change with shorter and more succinct names.

requirements.txt
akchinSTC: Will move dependency changes to test requirements..thanks for letting me know.

PR #270, opened 2018-02-21
Add support for specifying port range to use for kernel and EG ports

### LINKED ISSUE #288: Add ability to specify port ranges for kernels, as well as Enterprise Gateway
Some companies impose policies on applications running within their networks as to which ports are available for a given application.  Similarly, firewall configuration may also need to conform to such policies.  As a result, Enterprise Gateway should provide a mechanism by which port-ranges can be specified both globally and on a per-kernelspec basis.

lresende:
Add --kernel-port-range parameter that takes a range of
ports on the format xxx..yyy where xxx is the lower bound
port number and yyy is the upper bound port number and when
choosing the ports to use, the kernel launcher will consider
those boundaries to select the port

Fixes #288

### DIFF
--- a/docs/source/config-options.md
+++ b/docs/source/config-options.md
@@ -186,6 +186,13 @@ EnterpriseGatewayApp options
 --EnterpriseGatewayApp.port=<Integer>
     Default: 8888
     Port on which to listen (KG_PORT env var)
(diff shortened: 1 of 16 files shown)


lresende:
PR Merged.

PR #283, opened 2018-03-21
Initial deliver for Conductor integration

charlieeeeeee:
Changeset include,
1) Add conductor_endpoint to enterprisegatewayapp.py
2) Add conductor.py processproxy
3) Run script with debug removed from Spark-submit
4) kernel.json for conductor. Somewhat empty, all the parameters will be dynamically filled during runtime.

### DIFF
--- a/enterprise_gateway/enterprisegatewayapp.py
+++ b/enterprise_gateway/enterprisegatewayapp.py
@@ -70,6 +70,16 @@ class EnterpriseGatewayApp(KernelGatewayApp):
     def yarn_endpoint_default(self):
         return os.getenv(self.yarn_endpoint_env, self.yarn_endpoint_default_value)
 
+    # Conductor endpoint
+    conductor_endpoint_env = 'EG_CONDUCTOR_ENDPOINT'
+    conductor_endpoint_default_value = None
+    conductor_endpoint = Unicode(conductor_endpoint_default_value, config=True,
+        help="""The http url for accessing the Conductor REST API. (EG_CONDUCTOR_ENDPOINT env var)""")
+
+    @default('conductor_endpoint')
+    def conductor_endpoint_default(self):
+        return os.getenv(self.conductor_endpoint_env, self.conductor_endpoint_default_value)
+
     _log_formatter_cls = LogFormatter
 
     @default('log_format')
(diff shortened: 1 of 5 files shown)


enterprise_gateway/enterprisegatewayapp.py
```diff
@@ -70,6 +70,16 @@ def remote_hosts_default(self):
...
+    # Conductor endpoint
+    conductor_endpoint_env = 'EG_CONDUCTOR_ENDPOINT'
+    conductor_endpoint_default_value = None
+    conductor_endpoint = Unicode(conductor_endpoint_default_value, config=True,
+        help="""The http url for accessing the Conductor REST API. (EG_CONDUCTOR_ENDPOINT env var)""")
+
+    @default('conductor_endpoint')
+    def conductor_endpoint_default(self):
+        return os.getenv(self.conductor_endpoint_env, self.conductor_endpoint_default_value)
+
```
lresende: @kevin-bates should we have merge these configurations into a single RESOURCE_MANAGER_ENDPOINT type of parameter?

kevin-bates: @lresende. I'd rather we move the various *process proxy* configuration items to the specific classes, then perhaps those classes could have a callable *main()* method that displays information about the class including the configurable parameters.  This will take some research, so I think its best we leave the parameters separated for now and defined at more of the global scope.

The goal would be to have process proxy classes completely self-contained.  I'm just not sure these configuration items would be automatically *documented* when running `jupyter enterprisegateway --help-all` or not - although so long as there's a way for someone to get at the *help* content for the process proxy I don't view entries in `--help-all` as a requirement.

kevin-bates: Please see my comment in the conversation.  Let's leave this as-is for now.

enterprise_gateway/services/processproxies/conductor.py
```diff
@@ -0,0 +1,374 @@
...
+            Conductor node so that its in place when the kernel is started.  This step is skipped if pull or socket modes 
+            are configured, which results in the kernel process determining ports and generating encoding key.
+            Once started, the method will poll the application (after discovering the application ID via the
+            kernel ID) until host is known.  Note that this polling may timeout and result in a 503 Http error (Service 
+            unavailable).
+            Once the host is determined the connection file is retrieved. If pull mode is configured, the remote file is 
+            copied locally and member variables are loaded based on its contents.  If socket mode is configured, the
+            kernel launcher sends the connection information - which is then written out upon its reception.  If push
+            mode is configured, the kernel manager's IP is updated to the selected node.
+        """
```
kevin-bates: @charlieeeeeee - please remove all but perhaps the first sentence.  This comment is old and primarily incorrect and we need to update the YarnClusterProcessProxy comment as well.  (Feel free to submit that same update if you like!)
If you want, you could add some comment about how the command is generated via the update method.

PR #284, opened 2018-03-21
Add a trouble shooting section for Kerberos authentication

ajbozarth:
Including a short trouble shooting section for the most common issue I found while
testing Kerberos impersonation that was already covered in other documentation.

### DIFF
--- a/docs/source/getting-started-security.md
+++ b/docs/source/getting-started-security.md
@@ -79,9 +79,10 @@ The recommended approach for performing user impersonation when the kernel is la
 is to configure the `spark-submit` command within 
 [run.sh](https://github.com/jupyter-incubator/enterprise_gateway/blob/master/etc/kernelspecs/spark_python_yarn_cluster/bin/run.sh)
 to use the `--proxy-user ${KERNEL_USERNAME}` option.  This YARN option  requires that kerberos be configured within the
-cluster.  Please refer to the
+cluster.  In addition the gateway user (`elyra` by default) needs to be set up as a `proxyuser` in hadoop configs.
+Please refer to the
 [Hadoop documentation](https://hadoop.apache.org/docs/current/hadoop-project-dist/hadoop-common/Superusers.html) 
-regarding the proper configuration of `--proxy-user`.
+regarding the proper configurations.
 
 ##### Impersonation in Standalone or YARN Client Mode
 Impersonation performed in standalone or YARN cluster modes tends to take the form of using `sudo` to perform the 
--- a/docs/source/troubleshooting.md
+++ b/docs/source/troubleshooting.md
@@ -139,8 +139,7 @@ but it failed with a "Kernel error" and a SSHException.**
     This should be done on the host running Enterprise Gateway as well as all the remote hosts
     on which the kernel is launched.
 
-- **I'm trying to use a notebook with user impersonation on a Kerberos authenticated
-cluster but it fails to authenticate.**
+- **I'm trying to use a notebook with user impersonation on a Kerberos enabled cluster but it fails to authenticate.**
 
     When using user impersonation in a YARN cluster with Kerberos authentication, if Kerberos is not
     setup properly you will usually see the following warning that will keep a notebook from connecting:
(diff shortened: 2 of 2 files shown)


ajbozarth:
@kevin-bates @lresende I've added a doc section, it's short and sweet, but it addresses the one issue I ran into not already addressed elsewhere in the current docs. Feel free to give grammar feedback, I tried to match the rest of the docs in wording.

docs/source/troubleshooting.md
```diff
@@ -138,3 +138,17 @@ but it failed with a "Kernel error" and a SSHException.**
 
     This should be done on the host running Enterprise Gateway as well as all the remote hosts
     on which the kernel is launched.
+
+- **I'm trying to use a notebook with user impersonation on a Kerberos authenticated
```
lresende: I would say 'Kerberos enabled'

docs/source/troubleshooting.md
```diff
@@ -138,3 +138,17 @@ but it failed with a "Kernel error" and a SSHException.**
...
+    When using user impersonation in a YARN cluster with Kerberos authentication, if Kerberos is not
+    setup properly you will usually see the following warning that will keep a notebook from connecting:
+
+    ```bash
+    WARN Client: Exception encountered while connecting to the server : javax.security.sasl.SaslException: GSS initiate failed [Caused by GSSException: No valid credentials provided (Mechanism level: Failed to find any Kerberos tgt)]
+    ```
+
+    The most common cause for this WARN is when the user that started Enterprise Gateway is not
+    authenticated with Kerberos. This can happen when the user has either not run `kinit` or their
+    previous ticket has expired.
```
lresende: should we also add the command to properly init the user (e.g. if this was elyra user)

lresende:
Thanks @ajbozarth, how about also adding a section on the security page with some steps required to enable Kerberos? I believe the user that start EG needs to be listed in the yarn configuration so it can impersonate the actual user that is trying to start a kernel.

docs/source/troubleshooting.md
ajbozarth: Will be updated in my next commit

docs/source/troubleshooting.md
ajbozarth: The command is `kinit` so I'm not sure what to add, how a user authenticates is dependent on their server setup, `kinit` is just runs the default authentication for the user who runs it.

ajbozarth:
@lresende I'll add a section to the security page saying that the gateway user (typically elyra) must be added to the proxy user configs as described in the Hadoop Docs (already linked in the corresponding Security page section).

ajbozarth:
I've pushed my changes mentioned in my previous comments, but I'm not sure they fully address you're issues. I would love feedback on how to make things better, Kerberos setup is a very case by case issue and makes documenting it difficult.

docs/source/getting-started-security.md
```diff
@@ -79,9 +79,10 @@ The recommended approach for performing user impersonation when the kernel is la
 is to configure the `spark-submit` command within 
 [run.sh](https://github.com/jupyter-incubator/enterprise_gateway/blob/master/etc/kernelspecs/spark_python_yarn_cluster/bin/run.sh)
 to use the `--proxy-user ${KERNEL_USERNAME}` option.  This YARN option  requires that kerberos be configured within the
-cluster.  Please refer to the
+cluster.  In addition the gateway user (`elyra` by default) needs to be set up as a `proxyuser` in hadoop configs.
```
kevin-bates: When reading the doc reference, it seems like the docs refer to this as a *superuser*.  As a result, I think it might be more clear to say 
>... needs to be set up as a `hadoop.proxyuser` *superuser* in hadoop configs.

IMO, this makes `proxyuser` more distinct from `--proxy-user` since the two are only a typo away from each other but mean very different things.

ajbozarth:
@lresende given you merged this are @kevin-bates issues not a problem then?

kevin-bates:
@ajbozarth - yeah, @lresende added *superuser* following `proxyuser` so I'm good to go.  Thanks!

ajbozarth:
Ah, didn't catch that he edited the commit, looks good

PR #289, opened 2018-03-29
Disable SSH tunneling by default

### LINKED ISSUE #258: Streamline EG deployment and security requirements
Currently, we have SSH tunneling enable by default which will make deployments to fail in Yarn cluster mode in case of any of the security requirements/configurations are not properly handled in the cluster. 

What I would like to propose is the following:
* SSH tunneling is DISABLED by default
* On security section of our documentation, describe how to configure SSH tunneling together with the configuration requirements (maybe with options for strict versus lazy security).

In summary, my main goal is to reduce adoption barriers for folks that are just experimenting/trying EG, but still maintaining a good set of enterprise/security functionality available. 

Note that this will require us to update the code and remove the "StrictHostKeyChecking=no " while creating the tunnels in cluster mode as well as remove the reliance on AutoAddPolicy and moving to RejectPolicy on client support.

Note that, for our integration test builds, we should continue to run with ssh tunneling enable to make sure we don't regress on the existing capabilities

lresende:
Disable SSH tunneling and describe configuration requirements necessary before enabling it.

Fixes #258

### DIFF
--- a/etc/kernel-launchers/scala/toree-launcher/src/main/scala/launcher/utils/SocketUtils.scala
+++ b/etc/kernel-launchers/scala/toree-launcher/src/main/scala/launcher/utils/SocketUtils.scala
@@ -1,0 +1,86 @@
+/**
+  * Copyright (c) Jupyter Development Team.
+  * Distributed under the terms of the Modified BSD License.
+  */
+
+package launcher.utils
+
+import java.io.PrintStream
+import java.net.{InetAddress, ServerSocket, Socket}
+
+import scala.util.Random
+
+
+object SocketUtils {
+
+  val random: Random = new Random (System.currentTimeMillis)
+
+  def writeToSocket(socketAddress : String, content : String): Unit = {
+    val ipPort = socketAddress.split(":")
+    if (ipPort.length == 2) {
+      println("Sending connection info to gateway at %s\n%s".format(socketAddress, content)) // scalastyle:off
+      val ip = ipPort(0)
+      val port = ipPort(1).toInt
+      val s = new Socket(InetAddress.getByName(ip), port)
+      val out = new PrintStream(s.getOutputStream)
+      try {
+        out.append(content)
+        out.flush()
+      } finally {
+        s.close()
+      }
+    } else {
+      println("Invalid format for response address '%s'!".format(socketAddress)) // scalastyle:off
+    }
+  }
+
+  def findPort(portLowerBound: Int, portUpperBound: Int): Int = {
+
+    val socket = findSocket(portLowerBound, portUpperBound)
+    val port = socket.getLocalPort
+    println("port %s is available".format(port)) // scalastyle:off
+
+    // now Close the socket/port
+    socket.close()
+    println("Port %s closed...".format(port)) // scalastyle:off
+
+    port
+  }
+
+  def findSocket(portLowerBound: Int, portUpperBound: Int): ServerSocket = {
+
+    var foundAvailable: Boolean = false
+    var socket: ServerSocket = null
+
+    while (foundAvailable == false) {
+
+      val candidatePort = getCandidatePort(portLowerBound, portUpperBound)
+
+      // try candidatePort - only display 'Trying...' if in range
+      if ( candidatePort > 0 )
+        println("Trying port %s ...".format(candidatePort)) // scalastyle:off
+
+      try {
+        socket = new ServerSocket(candidatePort)
+        // return the socket to be used
+        foundAvailable = true
+      } catch {
+        case _ : Throwable => println("port %s is in use".format(candidatePort)) // scalastyle:off
+        socket = null
+      }
+    }
+
+    socket
+  }
+
+  private def getCandidatePort(portLowerBound: Int, portUpperBound: Int): Int = {
+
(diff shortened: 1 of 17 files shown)


lresende:
Not sure why this is still showing multiple commits from master. The only committs being added are the last two [cc149349a68f0d265c3a9f6f40e5819efc118e53](https://github.com/jupyter-incubator/enterprise_gateway/pull/289/commits/cc149349a68f0d265c3a9f6f40e5819efc118e53) and [d25d4e52261545474a4e6330c9ac96c6f6a97aa1](https://github.com/jupyter-incubator/enterprise_gateway/pull/289/commits/d25d4e52261545474a4e6330c9ac96c6f6a97aa1)

etc/kernel-launchers/scala/toree-launcher/src/main/scala/launcher/utils/SocketUtils.scala
```diff
@@ -0,0 +1,86 @@
...
+
+    // now Close the socket/port
+    socket.close()
+    println("Port %s closed...".format(port)) // scalastyle:off
+
+    port
+  }
+
+  def findSocket(portLowerBound: Int, portUpperBound: Int): ServerSocket = {
+
```
sanjay-saxena: Same comment as above.

etc/kernel-launchers/scala/toree-launcher/src/main/scala/launcher/utils/SocketUtils.scala
```diff
@@ -0,0 +1,86 @@
...
+    var foundAvailable: Boolean = false
+    var socket: ServerSocket = null
+
+    while (foundAvailable == false) {
```
sanjay-saxena: Perhaps, this can be:

```
while (foundAvailable) {
```

etc/kernel-launchers/scala/toree-launcher/src/main/scala/launcher/utils/SocketUtils.scala
```diff
@@ -0,0 +1,86 @@
...
+    socket
+  }
+
+  private def getCandidatePort(portLowerBound: Int, portUpperBound: Int): Int = {
+
+    val portRange = portUpperBound - portLowerBound
+    if ( portRange <= 0 )
+        return 0
+
+    val port = portLowerBound + random.nextInt(portRange)
```
sanjay-saxena: I am not sure whether we need to make this code thread-safe. If thread-safety is needed, perhaps we should do this:

```
val port = portLowerBound + ThreadLocalRandom.current().nextInt(portRange)
```

etc/kernel-launchers/scala/toree-launcher/src/main/scala/launcher/utils/SocketUtils.scala
```diff
@@ -0,0 +1,86 @@
...
+      } finally {
+        s.close()
+      }
+    } else {
+      println("Invalid format for response address '%s'!".format(socketAddress)) // scalastyle:off
+    }
+  }
+
+  def findPort(portLowerBound: Int, portUpperBound: Int): Int = {
+
```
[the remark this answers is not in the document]
kevin-bates: @sanjay-saxena - this is a good idea from a completeness standpoint.  I'll write up an issue for v0.9 to apply RFC 793.  I'm thinking this checkout could be made in EG, then we don't have to proliferate similar code across the launchers (although one could argue that would be best as well).  At any rate, let's wait until v0.9.  Thanks.

etc/kernel-launchers/scala/toree-launcher/src/main/scala/launcher/utils/SocketUtils.scala
kevin-bates: @sanjay-saxena do you mean `!foundAvailable`?

etc/kernel-launchers/scala/toree-launcher/src/main/scala/launcher/utils/SocketUtils.scala
kevin-bates: Probably a good idea, but not necessary since this portion of the launcher is single-threaded.

etc/kernel-launchers/scala/toree-launcher/src/main/scala/launcher/utils/SocketUtils.scala
sanjay-saxena: Yep. Sorry about missing the `!`.

etc/kernel-launchers/scala/toree-launcher/src/main/scala/launcher/utils/SocketUtils.scala
sanjay-saxena: Sounds good.

etc/kernel-launchers/scala/toree-launcher/src/main/scala/launcher/utils/SocketUtils.scala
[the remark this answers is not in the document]
sanjay-saxena: Sounds good.

PR #297, opened 2018-04-04
Fix tests to support Tornado 5.0

akchinSTC:
Initial PR to support tornado 5.0+


akchinSTC:
Restricting tornado to < 5.x causes pip (at least in my environment) to resolve to 4.5.3, but the unit tests do pass with pyzmq>= 17.x, so it'll work, ill just remove the 5.x ceiling from tornado
jupyter-kernel-gateway <2.x has a dependency on notebook<5.x, but in EG we specifically want notebook>5.1, so figure if its going to always resolve to 2.x, may as well update it
notebook was bumped to 5.3 since pre 5.3 version don't play well with tornado 5+
jupyter-client was bumped to 5.2 because notebook 5.3 has a requirement of jupyter-client 5.2+

willingc:
Hi @akchinSTC, Thanks for the explanation and updates.

Quick question: any particular reason that we are using `requirements-conda.txt` file for conda instead of an `environment.yml`? It's technically fine for merging the PR. I'm curious if there is something in the build for EG that prefers the txt file over the more common yml file for conda.

lresende:
@willingc I believe this is mostly used in the makefile to create a conda environment and was created to remove dependencies that were not available as conda packages. I will create an issue to investigate the possibility to migrate to the different approach based on conda best practices.

kevin-bates:
@akchinSTC - hmm, I think these version updates also need to be expressed in setup.py, otherwise wheel installations won't install the right stuff - reopening.

PR #301, opened 2018-04-11
Add spark-submit REST RESPONSE parser. Change application reference t…

charlieeeeeee:
…o driver reference because if application failed to created, JEG lose reference of driver and will not cleanup.

1) Remove redundant conditional checks
2) Add REST parser for Spark submit response
3) Change "Application" reference to "Driver" reference
4) Expose KERNEL_SPARK_OPTS - user configurable Spark configuration list to append to spark-submit

### DIFF
--- a/enterprise_gateway/services/processproxies/conductor.py
+++ b/enterprise_gateway/services/processproxies/conductor.py
@@ -11,6 +11,7 @@ import logging
 import subprocess
 import errno
 import socket
+import re
 from tornado import web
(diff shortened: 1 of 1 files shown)


charlieeeeeee:
@kevin-bates Yea, after you explained about the generic usage of the run.sh, I figured enabling stderr would obtain some useful info in conductor_processproxy, in our case the REST response from the run.sh process.

PR #318, opened 2018-04-26
Remove code that terminates launcher prematurely

### LINKED ISSUE #317: The R launcher is terminating itself on shutdown when it shouldn't
The gateway listener code running in the R launcher monitors two requests, a `signum` request in which case it passes the signal number on to the kernel via `kill` and a `shutdown` request in which it should exist the listener's loop because EG has told it to shutdown.  However, in this case, its also signalling itself to terminate (via `SIGTERM`) when it shouldn't.  These lines should be removed and just let a positive shutdown value exit the listener loop: https://github.com/jupyter-incubator/enterprise_gateway/blob/master/etc/kernel-launchers/R/scripts/gateway_listener.py#L98-L100

kevin-bates:
The launcher was terminating itself on shutdown requests when it
should terminate its listener loop.  This was resulting in KILLED
applications on Conductor (and YARN).  Note however, that after
this change, while Conductor applications are gracefully terminated
and indicate FINISHED (confirmed by @charlieeeeeee), YARN 
applications still appear as KILLED.

We'll need to look into why YARN applications are still not shutting
down gracefully.

Fixes #317

### DIFF
--- a/etc/kernel-launchers/R/scripts/gateway_listener.py
+++ b/etc/kernel-launchers/R/scripts/gateway_listener.py
@@ -95,9 +95,7 @@ def gateway_listener(sock, parent_pid):
                 os.kill(int(parent_pid), signum)
             elif request.get('shutdown'):
                 shutdown = bool(request.get('shutdown'))
-                if shutdown:
-                    print("Listener received shutdown request, terminating parent (pid: {}).".format(parent_pid))
-                    os.kill(int(parent_pid), signal.SIGTERM)
+
         else:  # check parent
             try:
                 os.kill(int(parent_pid), 0)


etc/kernel-launchers/R/scripts/gateway_listener.py
```diff
@@ -95,9 +95,7 @@ def gateway_listener(sock, parent_pid):
                 os.kill(int(parent_pid), signum)
             elif request.get('shutdown'):
                 shutdown = bool(request.get('shutdown'))
-                if shutdown:
-                    print("Listener received shutdown request, terminating parent (pid: {}).".format(parent_pid))
```
lresende: Do we need to actually keep the log message around?

kevin-bates: I don't believe so.  This behavior is consistent with the python launcher.  However, in checking the Scala launcher just now, it does produce the following: `Stopping gateway listener.`

Since the launcher is going to terminate, I think its fine to not have a message here.  If anything, it would probably be appropriate to log as the last line in the main launcher script.

PR #320, opened 2018-04-27
Add support for local ip blacklist

### LINKED ISSUE #319: Some network interfaces cause issues with IP used for response address
Multiple network interfaces are common.  In some situations, they can interfere with how the juptyer framework determines the set of public IPs.  For example, the framework determines that the default IP for the `docker0` interface (`172.17.0.1`) is in the list of public IPs, but when used to build the response-address for a remote kernel to convey its connection information, the remote kernel will get a connection failure.

Rather than just take the first in the list, which is the current behavior, there should be a way to better determine the correct IP to use.

kevin-bates:
Introduce the environment variable `EG_LOCAL_IP_BLACKLIST` that takes a
comma-separated list of local IPv4 addresses (can be regular expressions)
to NOT be used when building the response address information on which
the kernel launcher returns its connection information.

Fixes #319

### DIFF
--- a/enterprise_gateway/services/processproxies/processproxy.py
+++ b/enterprise_gateway/services/processproxies/processproxy.py
@@ -4,6 +4,7 @@
 
 import os
 import sys
+import re
 import signal
 import abc
 import json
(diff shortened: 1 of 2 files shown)


enterprise_gateway/services/processproxies/processproxy.py
```diff
@@ -53,7 +53,24 @@
...
+# This env can be used to indicate such IPs.
+local_ip_blacklist = os.getenv('EG_LOCAL_IP_BLACKLIST','').split(',')
+
+
+def get_local_ip():
+    """
+    Honor the blacklist-ed IPs, locating the first not in the list.
+    """
+    for ip in localinterfaces.public_ips():
+        if ip not in local_ip_blacklist:
```
lresende: I am not sure this would be flexible enough for virtual/containerized environments. If we could use a IP class, or something like a regEx, it would give the flexibility we need (e.g. 172.* instead of a specific ip as it might change from one node versus another particularly when we start launching EG on K8s, Docker, or in HA). Thoughts ?

kevin-bates: Having each item of the comma separated list be a regex is a good idea. However, even in the case being handled 172.* wasn't sufficient, but 172.17.* would be.
I think we should be careful not to over engineer this since use of this env should be rare. Spectrum happens to have these hybrid-visible networks.
I can add regex support tomorrow. Setting an env with regex characters may be tricky, but I'm sure it can be made to work.

kevin-bates: @lresende - Just pushed a commit (amending the original) with support for regular expressions.  Thanks for pointing that out - that's a great change.

PR #334, opened 2018-05-09
Disable pip --upgrade when building docker images

lresende:
Pip 10 seems to be causing issues when building the docker
images, so continue on what is available on the platform
until those issues are resolved.


kevin-bates:
I think the issue is that anaconda's pip wrapper still contains a reference to `pip.main()` - which isn't in v10.  The upgrade to pip v10 succeeds, but subsequent commands - that use the anaconda pip wrapper - fail.

Not sure if this applies to the nb2kg image since I don't think it uses anaconda.

PR #340, opened 2018-05-11
Updates to Make for Integration Tests

akchinSTC:
Makefile has updated 'itest' profile to invoke pytest instead of unittest as
its testing framework. Default settings are set to test all Cluster and Client
modes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -117,28 +117,24 @@ docker-clean docker-clean-enterprise-gateway docker-clean-nb2kg docker-clean-yar
 # itest should have these targets up to date: bdist kernelspecs docker-image-enterprise-gateway 
 
 # itest configurable settings
-# indicates which host (gateway or notebook) to connect to...
+# indicates which host (gateway) to connect to...
 ITEST_HOST:=localhost:8888
 # indicates the user to emulate.  This equates to 'KERNEL_USERNAME'...
 ITEST_USER:=bob
 # indicates the other set of options to use.  At this time, only the python notebooks succeed, so we're skipping R and Scala.
-ITEST_OPTIONS:=--notebook_files=../notebooks/Python_Client1.ipynb,../notebooks/Python_Cluster1.ipynb
+ITEST_OPTIONS:=
 
 ENTERPRISE_GATEWAY_TAG:=dev
 
-# here's a complete example of the options (besides host and user) with their expected values ...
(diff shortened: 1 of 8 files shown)


kevin-bates:
Thanks @akchinSTC - the changes look good.

Makefile
```diff
@@ -117,28 +117,24 @@ docker-clean docker-clean-enterprise-gateway docker-clean-nb2kg docker-clean-yar
...
 PREP_DOCKER:=1
 itest: ## Run integration tests (optionally) against docker container
 ifeq (1, $(PREP_DOCKER))
 	make docker-prep
 endif
-	@rm -rf build/itests
-	@mkdir -p build/itests
-	@cp -r enterprise_gateway/itests build
-	(cd build/itests/src/; $(SA) $(ENV); python main.py --host=$(ITEST_HOST) --username=$(ITEST_USER) $(ITEST_OPTIONS))
+	(cd enterprise_gateway/itests/src/; pytest *_testcase.py --host=$(ITEST_HOST) --username=$(ITEST_USER) $(ITEST_OPTIONS))
```
lresende: you should not remove `$(SA) $(ENV);` which is responsible for using the proper conda environment that has all dependencies installed. Without that, I get issues with websockets package not available, but this has been installed into the conda env based on the requirements-test.txt

Makefile
```diff
@@ -117,28 +117,24 @@ docker-clean docker-clean-enterprise-gateway docker-clean-nb2kg docker-clean-yar
 # itest should have these targets up to date: bdist kernelspecs docker-image-enterprise-gateway 
 
 # itest configurable settings
-# indicates which host (gateway or notebook) to connect to...
+# indicates which host (gateway) to connect to...
 ITEST_HOST:=localhost:8888
```
lresende: You probably want to enable overriding these by using `ITEST_HOST?=localhost:8888`, this is true for at least `ITEST_HOST`, `ITEST_USER`, `PREP_DOCKER`, you probably have a better idea if there are use cases where the other configurations should be also overridable.

kevin-bates: I agree with the idea to override the variables.  I do find, however, that `:=` still honors command-line overrides.  That said, I don't see that behavior documented anywhere while `?=` is documented to do the same - so we should probably make that change.

PR #351, opened 2018-05-18
Consolidate requirements into conda yml file

### LINKED ISSUE #304: Investigate moving from requirements-conda.txt to conda yml metadata files
See the discussion on PR #297 suggesting to move to use conda yml files for dependency declaration

akchinSTC:
Fixes #304


Makefile
```diff
@@ -18,12 +18,7 @@ help:
 
 build:
 env: ## Make a dev environment
-	-conda create -y -n $(ENV) -c conda-forge --file requirements-conda.txt \
-		--file requirements-test.txt
-	$(SA) $(ENV) && \
-		pip install -r requirements.txt && \
-		pip install -r requirements-doc.txt && \
-		pip install -e .
+	-conda env create --file requirements.yml --name $(ENV)
```
kevin-bates: Is there a reason `-y` and `-c conda-forge` were not preserved.  

I'm not a python or conda expert by any means but this [SO post](https://stackoverflow.com/questions/39857289/should-conda-or-conda-forge-be-used-for-python-environments) seems to lean toward keeping the channel at `conda-forge`. 

`-y` just strikes me as a good idea from an automation standpoint.

akchinSTC: Removed -y because during my installs, anaconda never prompted me for a response when creating an environment from a .yml configuration file. This might be unique to my own env though, but to be safe I added it back, cant hurt. The packages also seems to be pulling from conda-forge by default, at least in my env, as a safeguard though, I have added channels into the yml. Anaconda should pull from conda-forge first as long as anaconda's configuration setting "channel_priority" is set to true.

kevin-bates: 😄

Issue #356, opened 2018-05-24
Unauthorized_users lists are not honored when defined in process-proxy stanza


kevin-bates:
Administrators wishing to deny use of certain kernels to users can do so in two ways. 
1. The set of allowed users can be explicitly added to the `authorized_users` list within the process-proxy config stanza.  The target user will be denied access so long as they are not listed in this set of users.
2. The target user can be added to the `unauthorized_users` list within the process-proxy config stanza.  This option is useful when the set of authorized users is large.

The second approach is not working.  As a result, the only way to deny a user the use of a specific kernel is to explicitly authorize all other users.

PR #358, opened 2018-05-25
Add initial authorization tests

kevin-bates:
Added tests to validate the use of the authorized_users and unauthorized_users
lists that can be configured in the process-proxy:config stanza of a kernelspec.

This change also uses local kernelspec definitions that are mounted into the
docker image via the byok (bring-your-own-kernelspec) capabilities recently
added to image.  The idea is that different test-specific kernelspecs can be
added when necessary.

An actual bug (Issue #356) was identified during development of these changes.

Note: The travis builds associated with this PR will fail until PR #357 has
been merged since that contains the fix for the unauthorized_users list that
is tested as part of these changes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -135,5 +135,5 @@ endif
 docker-prep: 
 	@-docker rm -f itest >> /dev/null
 	@echo "Starting enterprise-gateway container (run \`docker logs itest\` to see container log)..."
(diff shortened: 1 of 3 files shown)


kevin-bates:
I've checked the travis failures and they're all related to this failure (as expected):
`test_unauthorized_users (enterprise_gateway.itests.test_authorization.TestAuthorization) ... FAIL`
As a result, I believe this PR can be merged following #357 and this failure should be resolved.

PR #368, opened 2018-06-01
Update docs to include kubernetes support information

### LINKED ISSUE #365: Integrate Kubernetes.md contents into the documentation hierarchy
The initial support for kubernetes documented items in a single file in the root of the repo [kubernetes.md](https://github.com/jupyter-incubator/enterprise_gateway/blob/master/kubernetes.md) to prevent difficult merges later.  We need to split up and merge the appropriate sections of that file into our documentation hierarchy.  This will likely require some additional documentation refactoring - a good thing!

This task should also include the necessary markdown files for each of the images that are then used on dockerhub.

kevin-bates:
Updated the online documentation to include the necessary information
for Kubernetes support.  This consisted of extending our set of process
proxy implementations and adding descriptions for the various images.

I also updated the elyra dockerhub kubernetes images with the contents
of their respective README.md files (added in this change).

A general walk through of the entire online documentation set was performed
and I updated the version-specific references to the latest, although
that isn't always necessary.  I felt like we should be more recent that
0.6.0 though.

I added a placeholder for the IBM Spectrum Conductor process proxy since
it was odd not have anything there at this point.  That can be filled in
later.  Hoping @charlieeeeeee can take a crack at that.

Fixes #365

### DIFF
--- a/docs/source/devinstall.md
+++ b/docs/source/devinstall.md
@@ -27,25 +27,38 @@ Entering `make` with no parameters yields the following:
 
 ```
 activate                       eval `make activate`
+clean-docker                   Remove docker images
+clean-enterprise-gateway       Remove elyra/enterprise-gateway:dev docker image
+clean-kubernetes-enterprise-gateway Remove elyra/kubernetes-enterprise-gateway:dev docker image
+clean-kubernetes-kernel-py     Remove elyra/kubernetes-kernel-py:dev docker image
+clean-kubernetes-kernel-r      Remove elyra/kubernetes-kernel-r:dev docker image
+clean-kubernetes-kernel-scala  Remove elyra/kubernetes-kernel-scala:dev docker image
+clean-kubernetes-kernel-tf-py  Remove elyra/kubernetes-kernel-tf-py:dev docker image
+clean-kubernetes               Remove kubernetes docker images
+clean-nb2kg                    Remove elyra/nb2kg:dev docker image
+clean-yarn-spark               Remove elyra/yarn-spark:2.1.0 docker image
 clean                          Make a clean source tree
 dev                            Make a server in jupyter_websocket mode
-dist                           Make binary and source distribution to dist folder
-docker-clean-enterprise-gateway Remove elyra/enterprise-gateway:dev docker image
-docker-clean-nb2kg             Remove elyra/nb2kg:dev docker image
-docker-clean-yarn-spark        Remove elyra/yarn-spark:2.1.0 docker image
-docker-clean                   Remove docker images
-docker-image-enterprise-gateway Build elyra/enterprise-gateway:dev docker image
-docker-image-nb2kg             Build elyra/nb2kg:dev docker image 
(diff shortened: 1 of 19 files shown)


docs/source/devinstall.md
```diff
@@ -73,7 +86,7 @@ otherwise the command will use the default environment.
 Build a wheel file that can then be installed via `pip install`
 
 ```
-make dist
```
lresende: Make dist will do all, do we need to call `bdist` explicitly here?

docs/source/getting-started-cluster-mode.md
```diff
@@ -17,17 +17,17 @@ SPARK_HOME:/usr/hdp/current/spark2-client                            #For HDP di
 EG_YARN_ENDPOINT=http://${YARN_RESOURCE_MANAGER_FQDN}:8088/ws/v1/cluster #Common to YARN deployment
 ``` 
 
-**Configuring Kernels for YARN Cluster mode**
+### Configuring Kernels for YARN Cluster mode
 
 For each supported Jupyter Kernel, we have provided sample kernel configurations and launchers as part of the release
-[e.g. jupyter_enterprise_gateway_kernelspecs-0.6.0.tar.gz](https://github.com/jupyter-incubator/enterprise_gateway/releases/download/v0.6.0/jupyter_enterprise_gateway_kernelspecs-0.6.0.tar.gz).
+[e.g. jupyter_enterprise_gateway_kernelspecs-0.9.1.tar.gz](https://github.com/jupyter-incubator/enterprise_gateway/releases/download/v0.9.1/jupyter_enterprise_gateway_kernelspecs-0.9.1.tar.gz).
```
lresende: Is there a way to have the version be something like a macro? In this case, it should actually be the current master version and not a specific release.

docs/source/getting-started-cluster-mode.md
```diff
@@ -17,17 +17,17 @@ SPARK_HOME:/usr/hdp/current/spark2-client                            #For HDP di
...
 
-Considering we would like to enable the iPython Kernel that comes pre-installed with Anaconda to run on Yarn Cluster mode, we
+Considering we would like to enable the IPython Kernel that comes pre-installed with Anaconda to run on Yarn Cluster mode, we
```
lresende: Good :)

docs/source/getting-started-cluster-mode.md
kevin-bates: @lresende - no idea.  We can look into this.  However, these artifacts are manually uploaded to the releases directory when tied to a given release - so I'm not sure how someone would pull "today's build".

docs/source/devinstall.md
kevin-bates: @lresende This section is for dev workflow.  I figured `bdist` was more useful since devs typically just build the wheel file when iterating.  The description needs to change to include what the target does if we left it as `dist` and we'd want to remove the next section (for building the kernelspecs).

I added a fully explained `make dist` section a couple items below.

PR #369, opened 2018-06-03
Update kubernetes kernelspecs and launchers with initialization mode

### LINKED ISSUE #366: Update kubernetes kernelspecs and launchers to use spark-context-initialization-mode
One of the areas that changed since initial kubernetes development was the introduction of the `--RemoteProcessProxy.spark-context-initialization-mode` parameter to specify one of `lazy`, `eager` or `none`.  Prior to that, the kubernetes branch had introduced a boolean flag `--RemoteProcessProxy.no-spark-context` that was used by those (vanilla) kernels not expecting the use of Spark.

This issue tracks the conversion to use initialization mode form of the option.  Kernelspecs using the boolean flag will convert to using a value of `none`, while the others will, for now, convert to `lazy` since that's the default mode.

kevin-bates:
One of the changes that occurred while kubernetes was on a separate branch
was the introduction of spark-context-initialization-mode.  These changes
include that update as well as the following found during testing:

1. Spark scala kubernetes kernels must create their spark context using the
eager mode.  For whatever reason, the Toree lazy initialization change does
not work via the spark-on-k8s fork.  However, eager works fine.  We can revist
once we switch to the Spark 2.3 code.

2. There weren't any Toree assembly files in the enterprise-gateway image. As
a result, the code that allows us to avoid pinning to specific versions doesn't
work because it requires that the assembly be present.  The Dockerfile has been
updated accordingly.

3. Added the changes from the other scala kernels that checks for the existence
of the toree assembly and launcher jar.

4. Updated handle_timeout() in processproxy.py since it wasn't using the changes
made to throw better error messages.

Fixes #366

### DIFF
--- a/enterprise_gateway/services/processproxies/processproxy.py
+++ b/enterprise_gateway/services/processproxies/processproxy.py
@@ -835,9 +835,8 @@ class RemoteProcessProxy(with_metaclass(abc.ABCMeta, BaseProcessProxyABC)):
             error_http_code = 500
             reason = "Waited too long ({}s) to get connection file".format(self.kernel_launch_timeout)
             timeout_message = "KernelID: '{}' launch timeout due to: {}".format(self.kernel_id, reason)
-            self.log.error(timeout_message)
             self.kill()
-            raise tornado.web.HTTPError(error_http_code, timeout_message)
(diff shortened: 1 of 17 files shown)


lresende:
I have restarted the build as the previous one failed.

PR #372, opened 2018-06-07
Create proper vanilla R kernel image for use on k8s

### LINKED ISSUE #370: Create separate R kernel images for spark and vanilla modes
Currently, we share the same image for both k8s Spark and Vanilla (non-spark) modes.  This issue splits this out into two images to ensure data scientists have a complete set of tools in both environments.

The R kernel image will be based on `juptyer/r-notebook` image since that has been commonly used in jupyter envs.

The image names should be (for now): `elyra/kubernetes-kernel-r` and `elyra/kubernetes-spark-kernel-r`.

This issue is related to #367

kevin-bates:
Base the R kernel image on the pre-existing jupyter/r-notebook image.
This contains a vetted list of R-packages so all we add are the kernelspec
files, the bootstrap command and any (one) R-package necessary for the
kernel launcher.  This image will use the existing name: elyra/kubernetes-kernel-r

The previous image, based on the k8s-on-spark fork continues as before but
has been renamed to: elyra/kubernetes-spark-kernel-r

The R launch script changed the inclusion of the SparkR package from
using the library() function to using the require() since the latter does
not throw an exception if the package is not present.  This is more compatible
with spark-context-initialization-mode values of 'none' - used by the by
vanilla images.

Fixes #370


etc/kernel-launchers/R/scripts/launch_IRkernel.R
```diff
@@ -1,7 +1,7 @@
-library(SparkR)
 library(argparser)
 library(jsonlite)
 
+require("SparkR")
```
lresende: Is this going to be ok with the vanilla kernel?

kevin-bates: Yes. See last paragraph of PR comment. Use of `library` is more binding than `require`.  Seems like `require` Is poorly named IMHO.

PR #377, opened 2018-06-26
Update REST Authorization header schema to Spectrum Conductor endpoint.

charlieeeeeee:
Spectrum Conductor endpoint REST authorization header only takes in value.


charlieeeeeee:
Hi Kevin,

Sorry for the confusion. This schema is not newly introduced (therefore, it's not version specific) It was my bad on the testing with Spark Master Authorization enabled scenario. After turning on authorization, I found that conductor_processproxy could not successfully make REST calls to the conductor_endpoint, it turned out that the header schema is incorrect. Thus, inducing this change. Conductor do not usually change the REST schema, at least we try not to. Sorry for the inconvenience.

Thanks,
Charlie

PR #379, opened 2018-07-03
utf-8 decode stdout from curl subprocess in YarnClusterProcessProxy

### LINKED ISSUE #378: Python 3 encoding bug in YarnClusterProcessProxy (and yarn_api_client)
Python 3 http response bodies are returned as bytes, not strings.  This causes json.load(s) to raise an exception in YarnClusterProcessProxy's curl subprocess calls, and also in yarn_api_client.

`TypeError: the JSON object must be str, not 'bytes'`

https://github.com/toidi/hadoop-yarn-api-python-client/pull/13 addresses this in yarn_api_client, but YarnClusterProcessProxy has several manual `curl` subprocess calls to also get responses from the ResourceManager.  The stdout output should also be have `.decode('utf-8')` called before being passed to `json.loads`.

A quick code search for `json.loads` also results in some other places that should probably be decoded to utf-8, and even some that already are (in tests).

ottomata:
Fixes #378

### DIFF
--- a/enterprise_gateway/services/processproxies/yarn.py
+++ b/enterprise_gateway/services/processproxies/yarn.py
@@ -292,7 +292,7 @@ class YarnClusterProcessProxy(RemoteProcessProxy):
         try:
             process = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, close_fds=True)
             output, stderr = process.communicate()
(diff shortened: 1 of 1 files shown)


kevin-bates:
Thank you for your contribution!

I could only test this against a YARN cluster using Python 2.7, but confirmed there were no regressions and the change makes sense.

PR #384, opened 2018-07-13
Configure non-root user for kubernetes images

### LINKED ISSUE #382: Kubernetes: Ensure images run as non-root user
By default, docker images run as the root user.  We should ensure that the images used in kubernetes run as non-root users.

The proposal is to use a different user for the enterprise-gateway container (e.g., `eg-svc`) than for the kernel containers (`eg-kernel`).

kevin-bates:
Since docker containers run as the root user by default, its a best practice
to run containers as non-root users.  This PR does that by configuring the
user `eg-svc` (UID 619) for the kubernetes-enterprise-gateway image and user
`eg-kernel` (UID 620) for the kubernetes-kernel images.  A correspondingly
named group is associated with each as well.

All files and working directories also have their ownership and permissions
changed approriately.

Please note that general testing has been performed.  We may find additional
permission changes necessary as we begin to use persistent volumes and other
'touch points'.

Fixes #382


akchinSTC:
Merged.
Commit ID # 2f94a66a7ef9c2b1511ca8f4561da005f51151ee

Issue #383, opened 2018-07-13
Kubernetes: Ensure enterprise-gateway and kernels use appropriately configured roles


kevin-bates:
Because enterprise-gateway needs to manage kernel-based pods and create namespaces and their service accounts, the service account associated with enterprise-gateway will need to be a cluster-bound role.  As a result, we need to ensure only the minimally viable roles are configured for that service account.

We should also do similar for the service accounts that are used in the kernel pods - i.e., associate the service accounts to the minimum roles necessary to perform their task.  These roles, however, will be relegated to the namespace.

PR #385, opened 2018-07-13
Configure use of namespaces for Enterprise Gateway and kernels

### LINKED ISSUE #381: Kubernetes: Per Kernel Namespaces
As a means of isolating kernel activities, it seems the right approach would be to create the kernel pod in its own namespace.  Currently, kernel pods are created in the same namespace as Enterprise Gateway unless the environment variable `KERNEL_NAMESPACE` is provided - at which time that value is used as the namespace.

To address default scenarios (where `KERNEL_NAMESPACE` is not provided), the Enterprise Gateway service should create a namespace using the k8s API and then delete that namespace upon the kernel's termination.  This results in two use cases...

1. Long-lived namespaces
Applications wishing to configure their own namespace for a kernel can do so using the aforementioned `KERNEL_NAMESPACE` approach.  When provided, EG will simply use that value, creating the pod within that namespace for vanilla (non-spark-on-k8s) kernels, and passing it to spark-submit via `--kubernetes-namespace` for spark-on-k8s kernels.
Such namespaces will NOT be deleted upon deletion of the kernel pod.

2. Short-lived namespaces
Applications not requiring long-lived namespaces will still want to achieve isolation for each kernel to address things like quota limits, etc.  In such cases, `KERNEL_NAMESPACE` will not provided.  In these cases, the namespace name will be generated from `KERNEL_USERNAME` and the kernel-to-be-launched's ID (e.g., `alice-fd677011-b651-4f10-9ef9-50ba7eef2cb6`).  The namespace will be created (prior to pod creation) and EG will record the fact that this namespace must be destroyed at the kernel's termination.  In fact, since deletion of a namespace deletes all associated objects, the kernel termination logic can simply delete the namepace when using short-lived namespaces, or simply delete the pod, when using long-lived namespaces.

Open questions arise relative to what kinds of roles are required for the default service account within the namespace.  At a minimum, it looks like spark-on-k8s requires pod list and create functionality since the driver pod must create executor pods.  For long-lived namespace support, we will need to document these rules, while for the short-lived namespaces, we will need to create the appropriate bindings upon creation of the namespace.  As a result, these objects may require explicit cleanup on the kernel's termination as well.

Any questions or suggestions are welcome!

kevin-bates:
This moves the kubernetes-enterprise-gateway.yaml from the location in
the docker hierarchy to kubernetes and combines it with enterprise-gateway-acl.yaml.
The file is also renamed to enterprise-gateway.yaml.

The yaml is extended to create a primary namespace named `enterprise-gateway` in
which all EG-based objects are created.  The service account in the cluster role
binding has been updated to be the default account in the enterprise-gateway
namespace.

A namespace will be created for each requested kernel if the kernel creation
request does NOT provide a namespace name in the `KERNEL_NAMESPACE` environment
variable.  If provided, the namespace is assumed to exist.  If no KERNEL_NAMESPACE
value is provided, a namespace of the name <kernel-username>-<kernel-id> will be
created.  In addition, a role binding object will be created that references
the default service account for the new namespace.

When a namespace is created during kernel startup, EG will record that fact and
is responsible for deleting the namespace and all associated objects upon the
kernel's termination.

Currently, there is no means for running kernels within the same namespace as
enterprise-gateway unless KERNEL_NAMESPACE is set to 'enterprise-gateway'.  We
could easily add that option (such that setting the env is not required) at a
later time.

We will tighten up service accounts and role-based access control via #383

Fixes #381


akchinSTC:
Merged.
Commit ID # 0561f63b42a7f2427457c0e44542b03d68751c8e

PR #386, opened 2018-07-16
Allow access from remote hosts

kevin-bates:
[Notebook PR 3714](https://github.com/jupyter/notebook/pull/3714/) in 5.6.0 has code that prevents remote host access.  Since
the entire purpose of Enterprise Gateway is to allow remote host access, we need
to enable the web setting `allow_remote_access` to prevent the 403 exception.
This was discovered due to recent Travis build failures since it uses updated resources.

This led to other dependency related issues, namely in the docker image used for
testing leading to the update of the jupyter framework files in the `Dockerfile`.

Also increased the timeout for waiting for the docker image startup from 30
to 60 seconds and added the ability to override on the command line using
`PREP_TIMEOUT=n`.


lresende:
LGTM

PR #414, opened 2018-08-14
Remove obsolete code that downloads Toree assembly

lresende:
Now that Toree assembly is already part of the kernelspecs
distribution we don't need to manually add the jar into
lib folder.


kevin-bates:
Thanks Luciano!

Issue #408, opened 2018-08-09
Investigate adding code coverage for EG


lresende:
Investigate adding a code coverage badge similar to what [IPython](https://github.com/ipython/ipython) does.

PR #415, opened 2018-08-16
code coverage addition

akchinSTC:
PR for testing addition of codecov


akchinSTC:
This should resolve issue #408

PR #417, opened 2018-08-24
Typo's in content change

sudheer0553:
Considering we would like to enable the Scala Kernel to run on YARN Cluster and Client mode
we would have to copy the sample configuration folder **spark_scala_yarn_client** and
**spark_scala_yarn_client** to where the Jupyter kernels are installed  --> instead of ** spark_scala_yarn_cluster **  **spark_scala_yarn_client** is mentioned twice


kevin-bates:
Thank you for your contribution.

PR #427, opened 2018-08-30
Remove Kubernetes prefix from docker images

akchinSTC:
- removed kubernetes prefix from docker images and from directories under /etc/docker
- updated respective k8s kernel.json files with the changes
- updated default image in k8s process proxy with change
- updated travis build with changes


kevin-bates:
This is excellent @akchinSTC - it cleans things up nicely!  

Would you mind applying the same changes to the references in the `docs/source` files?  I think a `grep -n kubernetes- *` in that directory will be sufficient in identifying the changes.

Also, many of the changes are in `devinstall.md` where we issue the results of just typing `make`.  If you could replace that entire output, that would be great since it could probably use a general update anyway.

Thank you.

kevin-bates:
@akchinSTC - thank you for the update.  This leads me to ask one more favor (please?)... 

Since we're renaming kernel images, I think `spark-kernel-r` should be named `kernel-spark-r`.  That way, all kernel-relative images start with `kernel` and end with their language indicator.  What, if anything, is in between merely better describes what else is provided (above vanilla).

PR #429, opened 2018-09-07
Add ability to use kernelspec.env for default values

kevin-bates:
Prior to this change, any values in the env stanza of a kernelspec would
override values provided by the caller in the json's env section - thereby
eliminating the ability to let those values defined in the kernelspec to
be used as defaults and the user-provided values used as overrides. Following
this change any json env's prefixed with KERNEL_ or appearing in the
KG_ENV_WHITELIST or KG_PROCESS_ENV_WHITELIST will be added back into the
kernel's env after it has been seeded with values from the kernelspec.  The
only exception to this is the internally defined KERNEL_GATEWAY variable which
is set to '1' to indicate the kernel was launched by the gateway.

Added test to validate this behavior.  This test requires use of locally
defined kernelspec so I added the ability to define such kernelspecs and
leverage setting of JUPYTER_PATH to the resources directory where the
kernelspecs reside.

Also removed close of ioloop that was (I believe) causing `loop is closed`
exceptions that side-affected test output.  This should not be necessary
since the ioloop is closed on process termination anyway.

### DIFF
--- a/docs/source/config-options.md
+++ b/docs/source/config-options.md
@@ -351,3 +351,16 @@ access at the kernel level_.  These values apply to **all** process-proxy kernel
 * `port_range`: This process proxy configuration entry can be used to override `--EnterpriseGatewayApp.port_range`.
 Any values specified in the config dictionary override the globally defined values.  These apply to all
 `RemoteProcessProxy` kernels.
+
+### Per-kernel Environment Overrides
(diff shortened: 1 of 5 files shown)


lresende:
LGTM, could you add/update docs describing how this would be used by users.

PR #430, opened 2018-09-07
Update DistributedProcessProxy to not use ssh if host is local server

kevin-bates:
DistributedProcessProxy is now "optimized" for the case when the target
host is the local server.  In such cases, use of ssh is bypassed and
kernel management is like that of the local case with the exception that
message-based interrupts still go through the communication port (in loopback).
Another advantage is that ZMQ port creation will occur much closer to the
invocation of the kernel - thereby significantly reducing the window where
port conflicts can arise.

Added example kernelspec - python_distributed - which demonstrates how the
native python kernel could be reconfigured to work in a distributed manner.

Updated integration test to exercise the distributed python kernel similar
to how the local python kernel is tested.

Found issues with the EG and Yarn images relative to the R Yarn Client
kernelspec.  Removed the addition of the PATH entry in that kernelspec's
env stanza (no other kernelspecs were doing this, can't recall why this
was necessary in the first place).  And updated the spark-conf to force
the location of the metastore_db directory and derby.log files to /tmp.
Otherwise, failures would occur (only for the R Yarn Client kernel) because
the working directory (/usr/local/share/jupyter) was not writable.

### DIFF
--- a/enterprise_gateway/itests/test_python_kernel.py
+++ b/enterprise_gateway/itests/test_python_kernel.py
@@ -111,6 +111,28 @@ class TestPythonKernelLocal(unittest.TestCase, PythonKernelBaseTestCase):
         cls.gatewayClient.shutdown_kernel(cls.kernel)
 
 
+class TestPythonKernelDistributed(unittest.TestCase, PythonKernelBaseTestCase):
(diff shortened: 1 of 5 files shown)


lresende:
LGTM

PR #439, opened 2018-09-10
Fix up remaining items related to recent image rename change.

kevin-bates:
Many of the README.md files used on DockerHub still referenced kubernetes-
prefixed image names, so I dropped the prefix.

The kernel-spark-r image couldn't be built because the old directory was
still in use.

Added a note to the enterprise-gateway README.md instructing those looking
for the YARN-based image to go to elyra/enterprise-gateway-demo.

Updated contents of enterprise-gateway-demo README.md to reference the new
image name.

I've also published the new images and seeded their DockerHub entries with
these README files.  I've also added notices of deprecation to the previous
kubernetes-kernel-* descriptions on DockerHub.


lresende:
LGTM

PR #447, opened 2018-09-17
Allow enterprise gateway to ignore the SIGHUP signal

tf3193:
We noticed that on python 3.x python would not ignore the SIGHUP by default. We made the similar change in kernel gateway, you can see the discussion here 
https://github.com/jupyter/kernel_gateway/pull/304


lresende:
LGTM

PR #449, opened 2018-09-19
Set Sphinx to use v1.7.9

akchinSTC:
Documentation builds were failing after Sphinx was bumped up to 1.8.0. Was unable to recreate the error locally. This PR will test to see if reverting to 1.7.9 resolves the issue.


lresende:
LGTM

PR #450, opened 2018-09-20
Initial implementation for launching docker-based kernels

kevin-bates:
- Introduced new abstract base class `ContainerProcessProxy` with the idea that container-based implementations can derive from that class.  This results in leaf classes implementing methods to get status and terminate resources.

- Updated `k8s.py` to use new base class. Created docker_pp.py for `DockerProcessProxy` and `DockerSwarmProcessProxy`. A second process proxy class was created since the implementations for traditional docker differed significantly from that for swarm.  This also enables the ability to launch either with the same installation, as well as the ability to launch the other from an Enterprise Gateway running in traditional or swarm.

- Introduces `KERNEL_NAME` and `KERNEL_SPECS_PATH` env variables to reflect the kernel (spec) name and its location.  Each of the bootstrap-kernel.sh scripts have been updated to use these values.  This allows us to use the same image for both docker and kubernetes configurations (for the most part - spark support will likely change this).

- Moved the bootstrap-kernel.sh script to language specific subdirectories under kernel-launchers/<lang>/bootstrap since these were redundant.

- Added `launch_docker.py` script that is referenced by docker kernelspecs. This script (similar to `launch_kubernetes.py`) is responsible for launching the docker service (swarm) or container (traditional) corresponding to the target kernel's process proxy.  If the process proxy is relative to docker swarm a 'docker mode' variable will indicate that fact since startup of the corresponding entity differs.  For swarm a once-only service is created and can reside anywhere in the swarm cluster. For traditional docker a regular container is created and will reside on the same host as Enterprise Gateway.

- Added itest-docker target (and renamed existing itest target to itest-yarn) for testing docker-based images.  Currently, only local tests occur since Spark isn't configured.  Because building 4 additional docker images appear to have serious side affects on travis, only itest-yarn is performed during builds. 

- Updated Makefile targets - removed kubernetes-specific targets and generalized others since k8s and docker images are shared.

- Updated docs with initial docker swarm and traditional docker references. While in there, I replaced URL instances of `jupyter-incubator` with `juptyer`.

- Initial update of dockerhub README.md files so that kubernetes and docker swarm deployments are referenced.  Generalized the kernel-based container entries to reference the enterprise-gateway deployment usage to eliminate redundancy.

- For the enterprise-gateway container to create and launch other containers, it it needs to mount /var/run/docker.sock and run as UID 0.  We need to look into this - most likely documenting that the EG container user needs to belong to the docker group and admins need to permit 'rw' access on /var/run/docker.sock to that group.  (This is commonly documented to enable 'docker in docker'
operations).

Open items:
- Need to look into session [affinity settings](https://docs.docker.com/ee/ucp/interlock/usage/sessions/) so replica count can be increased.
- Currently -u 0 (run as UID 0) is required for EG image to create kernel images.
We need to revisit build of kernel-r image once jupyter image is fixed.  Currently
using older image.
- It would be nice to name the docker_pp.py file docker.py but python import issues
occur due to conflict with docker api package.
- Currently, only local tests are performed against docker images via itest-docker
due to lack of Spark support.
- Need to look at increasing the Travis job timeout (50 minutes) or address length or
rework what is perfornmed via public travis.
- Should also look into "log folding" for the script section of the travis.yml.


kevin-bates:
A couple observations about the travis tests.
1. The itest-docker tests don't run on Python 2.7.  Not entirely sure why.  Perhaps its related to the docker image being based on python 3.x?  
2. Have found that the tests now take quite a while longer to run because at least 4 additional docker images need to be rebuilt and there's a 50 minute job timeout imposed by Travis.  I have asked for, and been granted, a timeout increase (to 90 minutes), but they only applied that to my fork.  I've since requested all forks be increased, but, in lieu of that, asked for the our primary and a couple dev forks to be increased.  

In addition, since I was in discussion with them, I asked if the resources could be increased based on fork - since I believe that's our bigger issue anyway.   

Its nice to see the python 3.6 tests passing within the time limit though.

PR #452, opened 2018-09-24
Set EG to use yarn-api-client dependency <0.3.0

akchinSTC:
kerberos issues with the new release of yarn-api-client


akchinSTC:
Need to look into further. Installing kerberos libraries into each respective docker image resolves the build issues, but some tests seem to hang and timeout using the new client.

lresende:
LGTM

Issue #395, opened 2018-07-23
Update test docker images to use latest spark to address security issue


akchinSTC:
Referencing CVE-2018-1334: Apache Spark local privilege escalation vulnerability
Some of the docker images we use for testing may still be using an older version of spark (2.1.0 - https://hub.docker.com/r/elyra/yarn-spark), suggest we update to patch.

```
Severity: High

Vendor: The Apache Software Foundation

Versions affected:

Spark versions through 2.1.2
Spark 2.2.0 to 2.2.1
Spark 2.3.0
Description: In Apache Spark up to and including 2.1.2, 2.2.0 to 2.2.1, and 2.3.0, when using PySpark or SparkR, it’s possible for a different local user to connect to the Spark application and impersonate the user running the Spark application.

Mitigation:

1.x, 2.0.x, and 2.1.x users should upgrade to 2.1.3 or newer
2.2.x users should upgrade to 2.2.2 or newer
2.3.x users should upgrade to 2.3.1 or newer
Otherwise, affected users should avoid using PySpark and SparkR in multi-user environments.
Credit:

Nehmé Tohmé, Cloudera, Inc.
```

akchinSTC:
Issue is partially resolved with PR $454. 
Kubernetes docker images are still running with Spark 2.2.0

PR #454, opened 2018-10-01
Update non-K8s docker images to Apache Spark 2.3.1

akchinSTC:
-Rebuilt the base image (aghorbani/spark:2.1.0) to use apache spark 2.3.1 and publish to docker under: akchin/spark:2.3.1
-Updated yarn-spark dockerfile to use 2.3.1 image
-Updated integration tests to reflect new version
-Updated R launcher to explicitly use toJSON method from jsonlite package

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -91,7 +91,7 @@ release: bdist sdist ## Make a wheel + source release on PyPI
 # Here for doc purposes
 docker-images:  ## Build docker images
 enterprise-gateway-demo:  ## Build elyra/enterprise-gateway-demo:dev docker image
-yarn-spark:  ## Build elyra/yarn-spark:2.1.0 docker image
+yarn-spark:  ## Build elyra/yarn-spark:2.3.1 docker image
 nb2kg:  ## Build elyra/nb2kg:dev docker image
 kubernetes-images: ## Build kubernetes docker images
 enterprise-gateway: ## Build elyra/enterprise-gateway:dev docker image
@@ -114,7 +114,7 @@ docker-image-enterprise-gateway: $(WHEEL_FILE)
 clean-docker: ## Remove docker images
 clean-enterprise-gateway-demo: ## Remove elyra/enterprise-gateway-demo:dev docker image
 clean-nb2kg: ## Remove elyra/nb2kg:dev docker image
-clean-yarn-spark: ## Remove elyra/yarn-spark:2.1.0 docker image
+clean-yarn-spark: ## Remove elyra/yarn-spark:2.3.1 docker image
 clean-kubernetes: ## Remove kubernetes docker images
 clean-enterprise-gateway: ## Remove elyra/enterprise-gateway:dev docker image
 clean-kernel-py: ## Remove elyra/kernel-py:dev docker image
--- a/Makefile
+++ b/Makefile
@@ -135,6 +135,10 @@ kubernetes-publish: ## Push kubernetes docker images to docker hub
 # itest configurable settings
 # indicates which host (gateway) to connect to...
 ITEST_HOST?=localhost:8888
+# indicates two things:
+# this prefix is used by itest to determine hostname to test against, in addtion,
+# if itests will be run locally with docker-prep target, this will set the hostname within that container as well
+export ITEST_HOSTNAME_PREFIX?=itest.local
 # indicates the user to emulate.  This equates to 'KERNEL_USERNAME'...
 ITEST_USER?=bob
 # indicates the other set of options to use.  At this time, only the python notebooks succeed, so we're skipping R and Scala.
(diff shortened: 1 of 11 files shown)


enterprise_gateway/itests/test_python_kernel.py
```diff
@@ -86,7 +86,7 @@ def test_get_deploy_mode(self):
 
     def test_get_host_address(self):
         result = self.kernel.execute("print(sc.getConf().get('spark.driver.host'))")
-        self.assertRegexpMatches(result, '\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}')
+        self.assertRegexpMatches(result, 'itest')
```
kevin-bates: We're going to want to be more flexible than this, since we should be able (and do) point at any cluster to run these tests against.  Since it appears spark changed the `spark.driver.host` property to be the actual hostname, I guess there are two options (perhaps others):
1. See if there's an applicable `spark.driver.ip` property and keep the result regex the same.
2. Modify the regex to accept any valid hostname, which is probably close to `[A-z,0-9,_,-,.]`.

enterprise_gateway/itests/test_r_kernel.py
```diff
@@ -85,7 +85,7 @@ def test_get_deploy_mode(self):
 
     def test_get_host_address(self):
         result = self.kernel.execute('unlist(sparkR.conf("spark.driver.host"))')
-        self.assertRegexpMatches(result, '\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}')
+        self.assertRegexpMatches(result, 'itest')
```
kevin-bates: See previous.

enterprise_gateway/itests/test_scala_kernel.py
```diff
@@ -85,11 +85,11 @@ def test_get_deploy_mode(self):
 
     def test_get_host_address(self):
         result = self.kernel.execute('sc.getConf.get("spark.driver.host")')
-        self.assertRegexpMatches(result, '\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}')
+        self.assertRegexpMatches(result, 'itest')
```
kevin-bates: Ditto.

enterprise_gateway/itests/test_python_kernel.py
kevin-bates: Ok - thanks.  How about this?  Let's define a test-based env, say, `ITEST_HOST_PREFIX` that represents the _expected_ hostname prefix.  This variable should default to the value of `itest`.  Then update the assertions to ensure this prefix (which should be treated as a regular expression string) appears as a prefix to the returned value.

This way we can run against entirely different clusters (outside the images used for integration tests) and, assuming each node of the cluster has the same prefix, should be able to pass the tests by conveying the appropriate value for `ITEST_HOST_PREFIX`.

Makefile
```diff
@@ -151,11 +155,11 @@ ifeq (1, $(PREP_DOCKER))
...
+	@-docker rm -f $(ITEST_HOSTNAME_PREFIX) >> /dev/null
 	@echo "Starting enterprise-gateway container (run \`docker logs itest\` to see container log)..."
-	@-docker run -itd -p 8888:8888 -h itest --name itest -v `pwd`/enterprise_gateway/itests:/tmp/byok elyra/enterprise-gateway-demo:$(ENTERPRISE_GATEWAY_TAG) --elyra
-	@(r="1"; attempts=0; while [ "$$r" == "1" -a $$attempts -lt $(PREP_TIMEOUT) ]; do echo "Waiting for enterprise-gateway to start..."; sleep 2; ((attempts++)); docker logs itest |grep 'Jupyter Enterprise Gateway at http'; r=$$?; done; if [ $$attempts -ge $(PREP_TIMEOUT) ]; then echo "Wait for startup timed out!"; exit 1; fi;)
+	@-docker run -itd -p 8888:8888 -h $(ITEST_HOSTNAME_PREFIX) --name $(ITEST_HOSTNAME_PREFIX) -v `pwd`/enterprise_gateway/itests:/tmp/byok elyra/enterprise-gateway-demo:$(ENTERPRISE_GATEWAY_TAG) --elyra
+	@(r="1"; attempts=0; while [ "$$r" == "1" -a $$attempts -lt $(PREP_TIMEOUT) ]; do echo "Waiting for enterprise-gateway to start..."; sleep 2; ((attempts++)); docker logs $(ITEST_HOSTNAME_PREFIX) |grep 'Jupyter Enterprise Gateway at http'; r=$$?; done; if [ $$attempts -ge $(PREP_TIMEOUT) ]; then echo "Wait for startup timed out!"; exit 1; fi;)
```
kevin-bates: Sorry, but I find it confusing to use a "prefix" for the host and container name.  It just feels wrong and a bit clunky.  Instead, just add a -e option (e.g., `-e ITEST_HOSTNAME_PREFIX=${ITEST_HOSTNAME_PREFIX}`) to the docker run command, leaving the previous -h and --name options intact.

enterprise_gateway/itests/test_python_kernel.py
```diff
@@ -84,9 +84,9 @@ def test_get_deploy_mode(self):
         result = self.kernel.execute("sc.getConf().get('spark.submit.deployMode')")
         self.assertRegexpMatches(result, '(cluster|client)')
 
-    def test_get_host_address(self):
-        result = self.kernel.execute("print(sc.getConf().get('spark.driver.host'))")
-        self.assertRegexpMatches(result, 'itest')
+    def test_get_hostname(self):
+        result = self.kernel.execute("import subprocess; subprocess.check_output(['hostname'])")
+        self.assertRegexpMatches(result, os.environ['ITEST_HOSTNAME_PREFIX'] + "*")
```
kevin-bates: Thanks for updating the test method.  Since this test is now nolonger dependent on Spark, we should probably move these into the xxxKernelBaseTestCase classes.

akchinSTC:
Fixes #395

PR #458, opened 2018-10-08
Enhance NB2KG image to be used in JupyterHub

lresende:


etc/docker/nb2kg/Dockerfile
```diff
@@ -1,24 +1,22 @@
...
-# Install Lab and NB2KG. Enable NB2KG extension.  Lab extension is optionally enabled at runtime.
-RUN pip install --upgrade jupyter_contrib_nbextensions jupyterlab nb2kg && \
-    jupyter contrib nbextension install --sys-prefix && \
-    jupyter serverextension enable --py nb2kg --sys-prefix
-
-ADD start-nb2kg.sh /usr/local/share/jupyter/
+# Install Lab and NB2KG. Enable NB2KG extension. Lab is enabled by default
+RUN pip install --upgrade jupyterlab nb2kg && \
+    jupyter serverextension enable --py nb2kg --sys-prefix && \
+    jupyter serverextension enable --py jupyterlab --sys-prefix
```
kevin-bates: Looks like this installs the lab extension unconditionally.  Previously, lab would only be enabled if the command line of the image indicated that it should be enabled.  I think we should preserve the ability to use this image in Noteb ook or Lab if possible.

etc/docker/nb2kg/start-nb2kg.sh
```diff
@@ -2,32 +2,17 @@
...
-elif [[ "${CMD}" != "notebook" ]];
-then
-	echo ""
-	echo "usage: <docker run arguments> [notebook | lab]"
-	echo "Entering shell..."
-	/bin/bash
-	exit 0
-fi
+echo "${@: -1}"
+
```
kevin-bates: Here's the code relative to the previous comment.  May need to add something equivalent.

etc/docker/nb2kg/Dockerfile
lresende: So, currently, if you go to http://host:8888 it will use the default notebook ui, otherwise, if you use http://host:8888/lab it will use jupyterlab ui.

PR #463, opened 2018-10-10
Add the Enterprise Gateway version stamp to the startup message.

kevin-bates:
This will help us during times of support.  It would be nice to capture all the components, but we can ask for a `pip freeze` for those.

Before:
```
[I 2018-10-10 12:53:16.677 EnterpriseGatewayApp] Jupyter Enterprise Gateway at http://0.0.0.0:8888
```

After:
```
[I 2018-10-10 12:53:16.677 EnterpriseGatewayApp] Jupyter Enterprise Gateway 1.1.0 is available at http://0.0.0.0:8888
```


lresende:
LGTM, Thanks Kevin.

PR #471, opened 2018-10-19
Update project versioning and release processes

akchinSTC:
- All version and tag references are set by the VERSION variable in the parent Makefile in the project root directory
  -  publish-images target - added a bash helper script to retag images to latest and push to repo. This was done due to a limitation with Make where it would see the second colon in the docker tag line and assume its another target. 
- removed explicit “dev” tag from all kernel.json files - wonder if we should keep some kind of token in there to easily sed replace it with something else


Makefile
```diff
@@ -107,7 +107,7 @@ clean-images clean-enterprise-gateway-demo clean-nb2kg clean-yarn-spark clean-ke
 publish-images: ## Push docker images to docker hub
 	make WHEEL_FILE=$(WHEEL_FILE) VERSION=$(VERSION) -C etc $@
 
-ENTERPRISE_GATEWAY_TAG?=dev
+ENTERPRISE_GATEWAY_TAG?=$(VERSION)
```
lresende: We could do something like in [this makefile](https://github.com/apache/incubator-toree/blob/master/Makefile), to use dev tag when version contains .dev and use a specific version when it's a release version (without tag) ?

akchinSTC: Is ENTERPRISE_GATEWAY_TAG used any other than the itest-yarn-prep target? seems like the its the only place its used..if so then I think we should just remove it and use whats in VERSION. 

if so, did we want to use just “dev”?  it seems not very informative if someone were to develop on multiple versions simultaneously(not likely but maybe). having something like 2.0.0-dev would be more helpful...or we would adopt toree’s release nomenclature and use a pseudo-maven SNAPSHOT reference?

kevin-bates: I don't think that specific tag is necessary.  +1 for removal.

Makefile
```diff
@@ -10,7 +10,7 @@ SA:=source activate
 ENV:=enterprise-gateway-dev
 SHELL:=/bin/bash
 
-VERSION:=2.0.0.dev0
+export VERSION:=2.0.0.dev0
```
kevin-bates: Why is `export` necessary here?

akchinSTC: Removed.

Makefile
akchinSTC: +1 to removal

etc/Makefile
```diff
@@ -30,6 +30,10 @@ clean: ## Make a clean source tree
 KERNELSPECS := kernelspecs_all kernelspecs_yarn kernelspecs_conductor kernelspecs_kubernetes kernelspecs_docker
 kernelspecs: $(KERNELSPECS)
 
+# Specify which version of docker images you would like to use in your kernelspecs in IMAGE_VERSION.
+# This value will be baked into the kernelspecs
+export IMAGE_VERSION:=$(TAG)
+
```
kevin-bates: I don't think the export is necessary.   Instead, I think you just need to use a double-dollar `$$` when referencing this variable from a make-defined function.

What is the purpose of IMAGE_VERSION?  If its to override the value of $TAG - which is the value of $VERSION, then I believe the declaration should be `IMAGE_VERSION?=$(TAG)`.  However, we need to be careful because the publish targets use $(TAG).  As a result, I think I'd prefer we remove IMAGE_VERSION and change the declaration of TAG to `TAG?=$(VERSION)`.  Then one could build/push their own set of custom-tagged images using:

`make TAG=myCustomTag dist docker-images publish-images`

etc/Makefile
```diff
@@ -65,7 +69,7 @@ define BUILD_KERNELSPEC
 $1: $$(FILE_$1)
 $$(FILE_$1): ../build/kernelspecs
 	rm -f $$(FILE_$1)
-	@( cd ../build/kernelspecs; tar -pvczf "../$$(FILE_$1)" $$(PATTERN_$1) )
+	@( cd ../build/kernelspecs; find . -name 'kernel.json' -print0 | xargs -0 sed -i.bak "s/VERSION/$(IMAGE_VERSION)/g"; find . -name *.bak -print0 | xargs -0 rm -f; tar -pvczf "../$$(FILE_$1)" $$(PATTERN_$1) )
```
kevin-bates: Please move the VERSION replacement to the `../build/kernelspecs` target (to the second to last statement just after the second cp command).  Otherwise we're performing the replacement multiple times when only the first attempt does anything and all other _massaging_ takes place above.  This target should be purely for applying the pattern to produce the various tar files (i.e., a read-only step).

etc/Makefile
akchinSTC: Removed IMAGE_VERSION, and set TAG to ?=$(VERSION)

etc/Makefile
akchinSTC: relocated sed replacement to ../build/kernelspecs target to avoid multiple iterations

etc/docker/enterprise-gateway-demo/Dockerfile
```diff
@@ -1,4 +1,7 @@
-FROM elyra/yarn-spark:dev
+ARG HUB_ORG
+ARG VERSION
+
+FROM $HUB_ORG/yarn-spark:$VERSION
```
lresende: Nice ...

akchinSTC:
Should be good to merge

kevin-bates:
Merged - thank you @akchinSTC!

PR #473, opened 2018-10-26
get the assigned host ip for pod

SolarisYan:
We should use host_ip to get the assigned host ip for pod in k8s,not the same pod_ip, if my understanding is correct

### DIFF
--- a/enterprise_gateway/services/processproxies/docker_swarm.py
+++ b/enterprise_gateway/services/processproxies/docker_swarm.py
@@ -167,7 +167,7 @@ class DockerProcessProxy(ContainerProcessProxy):
                         self.log.warning("Docker network '{}' could not be located in container attributes - "
                                          "using assigned_ip '{}'.".format(docker_network, self.assigned_ip))
 
-                    self.assigned_host = self.assigned_ip
+                    self.assigned_host = self.container_name
 
         if iteration:  # only log if iteration is not None (otherwise poll() is too noisy)
             self.log.debug("{}: Waiting to connect to docker container. "
--- a/enterprise_gateway/services/processproxies/k8s.py
+++ b/enterprise_gateway/services/processproxies/k8s.py
@@ -57,7 +57,8 @@ class KubernetesProcessProxy(ContainerProcessProxy):
                 if pod_status == 'Running' and self.assigned_host == '':
                     # Pod is running, capture IP
                     self.assigned_ip = pod_info.status.pod_ip
-                    self.assigned_host = pod_info.status.host_ip
+                    self.assigned_host = pod_info.status.pod_ip
+                    self.assigned_node_ip = pod_info.status.host_ip
 
         if iteration:  # only log if iteration is not None (otherwise poll() is too noisy)
             self.log.debug("{}: Waiting to connect to k8s pod in namespace '{}'. "
--- a/enterprise_gateway/services/processproxies/k8s.py
+++ b/enterprise_gateway/services/processproxies/k8s.py
@@ -57,7 +57,7 @@ class KubernetesProcessProxy(ContainerProcessProxy):
                 if pod_status == 'Running' and self.assigned_host == '':
                     # Pod is running, capture IP
                     self.assigned_ip = pod_info.status.pod_ip
-                    self.assigned_host = pod_info.status.pod_ip
+                    self.assigned_host = pod_info.status.host_ip
 
         if iteration:  # only log if iteration is not None (otherwise poll() is too noisy)
             self.log.debug("{}: Waiting to connect to k8s pod in namespace '{}'. "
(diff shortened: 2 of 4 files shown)


SolarisYan:
@kevin-bates Thank you for the explanation. It seems that I understand it wrong. As you mentioned that ```The purpose of self.assigned_host is primarily to put a more human-readable name to self.assigned_ip```,assigned_ip should obviously be pod's own,then assigned_host should also be pod's own.So option 2 is the best, because pod name is the hostname for k8s pod:
```
kubectl -n  enterprise-gateway exec enterprise-gateway-6f96c889d7-l8qjb -- hostname
enterprise-gateway-6f96c889d7-l8qjb
```
But we have get the pod name [at here](https://github.com/jupyter/enterprise_gateway/blob/4e1dadb14f962d2ad028df0d27de674e17e0a371/enterprise_gateway/services/processproxies/k8s.py#L54), andi want to figure out which host its running on directly,so option 1 is more suitable for me.

In summary, I think there is ambiguity here, my suggestion is to change the assigned_host to assigned_name.

kevin-bates:
@solaris - great discussion - thanks for your response.

I completely agree that there's ambiguity here, but that ambiguity is only relative to the container-based process proxies.  `assigned_host` is perfectly meaningful to the YARN, Distributed and Conductor proxies.  I should have clarified my statement regarding the purpose of `assigned_host`.  Its intention, besides being more user-firendly, is to possibly help with re-establishment of a kernel's communication, should a _different_ EG instance replace the original (part of our longer-term _session persistence_ story).  This is true of any values referenced in the {get,load}_process_info() methods.  For example, should the ability to re-connect not succeed using `assigned_ip`, then we may need to fall back on `assigned_host`.  That's hypothetical, but that's another intention of `assigned_host`.

I'd prefer we keep `assigned_host` rather than change to `assigned_name` for a couple reasons.  1) `assigned_host` is distinct in non-container proxies.  2) Although unlikely, anyone that may have created their own process proxy as a standalone module will break if `assigned_host` is now `assigned_name`.  I know that's unlikely, but that's something we need to consider when working in the process proxy base classes.

What about we add a new _assigned_ member that gets introduced on `ContainerProcessProxy` and call it `self.assigned_node_ip`?  (I was first thinking `assigned_node` but since k8s provides `host_ip` and you probably want the more helpful `ip`, we should avoid another level of ambiguity and go with `assigned_node_ip`.)

Since this value is something that should get persisted, we'd need to extend `ContainerProcessProxy` with the `{get,load}_process_info()` methods to reference that member.  You could reference the pair in k8s.py as an example, but define in container.py.

We'll need to figure out what to do for the docker_swarm process proxy's, but that field could be unused for now in that file - unless you know how to navigate the network metadata provided by docker.  If not, we can do that later.

Would adding `self.assigned_node_ip` work for you?  If so, could you update your PR with these changes and let's also go with option 2 in the k8s.py file for how to set `self.assigned_host`.  That way, it will be identical to `DockerSwarmProcessProxy`.  (I just noticed that `DockerProcessProxy` takes option 3 approach - sigh.  It should probably get updated to also take option 2 so that all container-bassed proxies are the same.  Would you mind updating [this line](https://github.com/jupyter/enterprise_gateway/blob/master/enterprise_gateway/services/processproxies/docker_swarm.py#L170) to be like [this line](https://github.com/jupyter/enterprise_gateway/blob/master/enterprise_gateway/services/processproxies/docker_swarm.py#L82)?)

This is a good discussion - thanks for raising it.

I assume this is for your monitoring functionality that you're building.  I'm curious how you're exposing the process proxy class information to your applications?  Just a thought, but you might want to consider leveraging the `process_info` persistence stuff.  Currently it just writes all kernel session information to a file, but the idea is that we'll eventually use a lightweight DB (preferrably pluggable) that can be shared across a cluster of EG instances.  This information is then removed when the kernel is shutdown - so it represents an active snapshot.  Of course, that facility could also be used to produce historical metrics, auditing, etc.

enterprise_gateway/services/processproxies/k8s.py
```diff
@@ -57,7 +57,8 @@ def get_container_status(self, iteration):
                 if pod_status == 'Running' and self.assigned_host == '':
                     # Pod is running, capture IP
                     self.assigned_ip = pod_info.status.pod_ip
-                    self.assigned_host = pod_info.status.host_ip
+                    self.assigned_host = pod_info.status.pod_ip
```
kevin-bates: I think we should be consistent across the container sub-classes and set `self.assigned_host` to `self.container_name`

enterprise_gateway/services/processproxies/docker_swarm.py
```diff
@@ -167,7 +167,7 @@ def get_container_status(self, iteration):
                         self.log.warning("Docker network '{}' could not be located in container attributes - "
                                          "using assigned_ip '{}'.".format(docker_network, self.assigned_ip))
 
-                    self.assigned_host = self.assigned_ip
+                    self.assigned_host = self.container_name
```
kevin-bates: Thank you.

enterprise_gateway/services/processproxies/k8s.py
SolarisYan: Yeah, i can't agree more

kevin-bates:
@SolarisYan - I went ahead and added the container_name change and merged.  Thank you for your contribution!

PR #483, opened 2018-11-02
Properly set PYTHONUSERBASE relative to user home folder

lresende:


lresende:
I believe we don't need this for client mode (let me update the pr) and regarding ${HOME}, that is always empty on the YARN environment where the application is being executed.

But, in summary, there is a need to set a directory to where pip will install --user packages.

lresende:
I have tested all kernels (client/cluster modes) and all except R seems to be working, but the R issue might be unrelated to these changes. I will investigate it further.

kevin-bates:
LGTM - thanks.  I confirmed that R YARN client mode works - so your env must have an unrelated issue.

PR #486, opened 2018-11-05
Python kernel image based on jupyter/scipy-notebook image

### LINKED ISSUE #367: Update vanilla python kernel docker image to be based on Anaconda
Current we have all kernel docker images based on the spark images. We should continue to have these, which would be used by Spark on Kubernetes support, but the vanilla ones we should base on anaconda to make it much simpler for data scientists to use these as bases for their work.

kevin-bates:
The kernel-py image is now using jupyter/scipy-notebook pinned to tag 61d8aaedaeaf.

The existing image based on the spark-on-k8s fork is now referenced in
the `kernel-spark-py` kernelspec. Makefiles were updated to handle this target.

Updated the Tensorflow images to use version 1.11 since 1.09 is no longer
available.  This version is also what the jupyter/tensorflow-notebook image
is based on.

Updated the kernel image builds to only pull in the kernelspec tar files for
the kubernetes and docker kernelspecs.  However, the enterprise-gateway image
is built will all kernelspecs.

Fixes #367


akchinSTC:
LGTM

PR #501, opened 2018-11-26
Update image startup script locations

### LINKED ISSUE #480: Update images to be more in line with Jupyter community images
We should update our images to be more like those found in the Jupyter community.  This is primarily for consistency purposes, which has a significant benefit for adoption and easier support.

There are two primary areas that I can immediately see:  USER and CMD

USER:
Our images currently (and correctly) run under a non-root (non-zero) user (uid).  The kernel images run using user `eg-kernel` (UID 620) while the enterprise-gateway image runs as user `eg-svc` (UID 619).  We should adopt the _Jupyter Community_ approach of using user `jovyan` (UID 1000) for all images, both kernel and EG.  In addition, we should ensure the user is a member of the `users` group (100).  I have found that our spark-k8s images already have `users (100)` defined, so that's good.

For some of these images, those built on jupyter community images, the jovyan user is likely already in place.

By doing this, most users will already be familiar with that idea, but also, for those that aren't, they will more easily evolve within the community.

CMD:
Most of our container commands are placed in `/etc` and/or `/usr/local/share/jupyter`.  Jupyter community images place their commands in `/usr/local/bin` - which seems more correct anyway, but also serves as a single location.

As a result, we should move our various `bootstrap-kernel.sh` scripts from /etc to /usr/local/bin, along with any other things (like `spark-exec.sh`).  Similarly, the `start-enterprise-gateway.sh` script that's in `/usr/local/share/jupyter` should also be moved `/usr/local/bin`.

This change will require our k8s and docker scripts be updated as well.

akchinSTC:
Update the location of various bootstrap and startup scripts to be
more in line with the Jupyter community as well as default user to
"jovyan"

Fixes #480

### DIFF
--- a/docs/source/getting-started-kubernetes.md
+++ b/docs/source/getting-started-kubernetes.md
@@ -370,7 +370,7 @@ spec:
       value: $spark_context_initialization_mode
     image: $docker_image
     name: $kernel_username-$kernel_id
-    command: ["/etc/bootstrap-kernel.sh"]
+    command: ["/usr/local/bin/bootstrap-kernel.sh"]
 ```
 There are a number of items worth noting:
 1. Kernel pods can be identified in three ways using `kubectl`: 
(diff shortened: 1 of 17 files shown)


kevin-bates:
I had the exact same thought relative to the kernel bootstraps when I was looking at the kernel-id parameter change to the launchers.  In fact, I was going to slip that change in, but forgot about it.  Instead, I'll do the right thing and write up an issue. 😃

PR #505, opened 2018-11-29
Consolidate the kernel bootstrap scripts into one

### LINKED ISSUE #503: Consolidate kernel image bootstrap scripts into single file
Currently, we have three separate `bootstrap-kernel.sh` scripts - one for each language.  These scripts all do essentially the same thing...
1. Set some global values 
2. Validate the expected language
3. Launch their appropriate launcher

These scripts reside in the language-specific subdirectory in a `bootstrap` directory under `kernel-launchers`.

We should combine these scripts into a single `bootstrap-kernel.sh` file and promote the `bootstrap` subdirectory to be directly under `kernel-launchers` (sibling to the language-specific directories). 

In `bootstrap-kernel.sh` we should set the common globals, then, per-language, make shell function calls to each language-specific boot sequence.  This will yield a clean top half of the file.  I would recommend we make the scala bootstrap function last in the file since the python and R functions will be relatively short.

Of course, with this change, the _expected language_ enforcement becomes just a check on a _valid language_, but that's probably fine.

kevin-bates:
To simplify maintenance of the kernel image bootstrap files, all have
been combined into a single file where each language-specific action
resides in its own function.

In addition to this change, the following fixes were applied...
1. The new kernel-id parameter (replacement of the dummy {connection_file})
have been added to the launch scripts for docker and kubernetes and
the connection file has been removed from the kernel-pod.yaml file.
2. The bootstrap files for enterprise-gateway and enterprise-gateway-demo
had their `--elyra` command options restored (from `--jovyan`) since these
are command indicators and the docker/k8s launchers were still using `--elyra`.
3. The kernelspecs build was producing bad tar filenames and image tags within
the kernel.json files. These were fixed by propagating both VERSION and TAG
to the etc/Makfile on all make commands.
4. Fixed up the demo image so that it can be invoked with a command other than
`--elyra` or `--yarn` and that command will be invoked w/o having to wait for
YARN to startup.  This enables commands like `/bin/bash` to be used to gain
immediate access to the container.

Fixes #503


etc/docker/enterprise-gateway-demo/bootstrap-enterprise-gateway.sh
```diff
@@ -5,13 +5,12 @@
...
 CMD=${1:-"--help"}
-if [[ "$CMD" == "--help" ]];
-then
+if [[ "$CMD" == "--help" ]]; then
 	echo ""
 	echo "usage: docker run -it[d] --rm -h <container-hostname> -p 8888:8888 [-p 8088:8088 -p 8042:8042] <docker-opts> <docker-image> <command>"
 	echo ""
 	echo "where <command> is:"
-	echo "    --jovyan ... Invokes Enterprise Gateway as user 'jovyan' directly.  Useful for daemon behavior."
+	echo "    --elyra ... Invokes Enterprise Gateway as user 'jovyan' directly.  Useful for daemon behavior."
```
lresende: what if we name this --gateway ?

etc/docker/enterprise-gateway/bootstrap-enterprise-gateway.sh
```diff
@@ -11,17 +11,15 @@ then
 	echo "usage: docker run <docker-opts> <docker-image> <command>"
 	echo ""
 	echo "where <command> is:"
-	echo "    --jovyan ... Invokes Enterprise Gateway directly.  Useful for daemon behavior."
+	echo "    --elyra ... Invokes Enterprise Gateway directly.  Useful for daemon behavior."
```
lresende: assuming this is the only option, and it's not shared with the other docker, do we need the `--elyra` option here? If it's just the default, I would say remove it...

etc/kernel-launchers/docker/scripts/launch_docker.py
```diff
@@ -81,21 +79,24 @@ def launch_docker_kernel(connection_file, response_addr, spark_context_init_mode
 
 if __name__ == '__main__':
     """
-        Usage: launch_docker_kernel [connection_file] [--RemoteProcessProxy.response-address <response_addr>]
+        Usage: launch_docker_kernel 
+                    [--RemoteProcessProxy.kernel-id <kernel_id>]
+                    [--RemoteProcessProxy.response-address <response_addr>]
```
lresende: is there a case where we need the original key from the connection document?

kevin-bates: The connection file is not used - only the kernel-id is pulled from its name.  The launchers generate a key and send it back via the communication port.

etc/docker/enterprise-gateway-demo/bootstrap-enterprise-gateway.sh
kevin-bates: That's fine - I'll make this change.  It will also affect the makefile.

etc/docker/enterprise-gateway/bootstrap-enterprise-gateway.sh
kevin-bates: I'm incline to keep the command placeholder - renamed to `--gateway` - but if no command is provided act as if `--gateway` was provided rather than today's `--help` default.  This still allows us to specific a specific CMD in the k8s yaml and docker launch scripts, which I believe avoids ambiguity.
Are you okay with that?

PR #507, opened 2018-11-29
Update IRkernel to pull from CRAN instead of Github

akchinSTC:
Updated pulls for IRkernel v0.8.14 via CRAN.
Updated CRAN mirrors to be consistent across dockerfiles
Removed conda installation of IRkernel so all dockerfiles pull IRkernel from CRAN

### DIFF
--- a/etc/docker/demo-base/Dockerfile
+++ b/etc/docker/demo-base/Dockerfile
@@ -37,12 +37,10 @@ RUN wget --quiet https://repo.anaconda.com/miniconda/Miniconda2-latest-Linux-x86
     ln -s /opt/anaconda2/etc/profile.d/conda.sh /etc/profile.d/conda.sh && \
     echo ". /opt/anaconda2/etc/profile.d/conda.sh" >> ~/.bashrc && \
     echo "conda activate base" >> ~/.bashrc
-RUN conda install -v --yes --quiet pip jupyter r-irkernel r-devtools && \
-	Rscript -e 'install.packages("argparser", repos="https://cran.rstudio.com")' \
-            -e 'devtools::install_github("apache/spark@v2.3.1", subdir="R/pkg")' \
-            -e "options(unzip = 'internal')" \
-            -e "Sys.setenv(TAR = '/bin/tar')" \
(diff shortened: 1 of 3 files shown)


etc/docker/demo-base/Dockerfile
```diff
@@ -37,12 +37,10 @@ RUN wget --quiet https://repo.anaconda.com/miniconda/Miniconda2-latest-Linux-x86
...
     echo "conda activate base" >> ~/.bashrc
-RUN conda install -v --yes --quiet pip jupyter r-irkernel r-devtools && \
-	Rscript -e 'install.packages("argparser", repos="https://cran.rstudio.com")' \
-            -e 'devtools::install_github("apache/spark@v2.3.1", subdir="R/pkg")' \
-            -e "options(unzip = 'internal')" \
-            -e "Sys.setenv(TAR = '/bin/tar')" \
-            -e "devtools::install_github('IRkernel/IRkernel@0.8.14')"
+RUN conda install -v --yes --quiet pip jupyter r-devtools && \
+	Rscript -e 'install.packages("argparser", repos="http://cran.cnr.berkeley.edu")' \
+	        -e 'install.packages("IRkernel", repos="http://cran.cnr.berkeley.edu", lib="/opt/anaconda2/lib/R/library")' \
```
kevin-bates: Is there a reason the IRkernel can't be included in the same list of packages like that used in the kernel-spark-r/Dockerfile below?  Seems like a good simplification.  Also, since the previous installation didn't require the lib parameter for the argparser package, how did you know the IRkernel package required it? Just curious.

akchinSTC: It was a weird bug, maybe it just me, but if i combined the two into a single install packages line, it would fail with a cannot read/write to a R library directory. Combining the two and explicitly defining the library directory results in the Rscript succeeding, but with an error stating that IRkernel was not installed, bashing into the docker container confirms it wasn't installed. Separating the two was the only way I was able to to get it install properly, strange.

kevin-bates: Bizarre.  Thanks for the explanation.

PR #517, opened 2018-12-05
[WIP] Updating to use OpenJDK instead of oracle

akchinSTC:
1.x builds were failing due to broken link to oracle JDK in image used for testing. Removed and replaced with OpenJDK from system packages (via yum) to avoid having to update the oracle link too frequently.

### DIFF
--- a/etc/docker/yarn-spark/Dockerfile
+++ b/etc/docker/yarn-spark/Dockerfile
@@ -4,10 +4,11 @@ FROM aghorbani/spark:2.1.0
 USER root
 
 ENV ANACONDA_HOME=/opt/anaconda2
-ENV PATH=$ANACONDA_HOME/bin:$PATH
+ENV JAVA_HOME /usr/lib/jvm/jre-1.8.0
+ENV PATH=$ANACONDA_HOME/bin:$JAVA_HOME/bin:$PATH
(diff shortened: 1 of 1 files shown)


kevin-bates:
These changes look good although the test failures are consistent relative to the scala cluster kernel, so there may be a side-affect here relative to cluster-mode.

akchinSTC:
Looks like scala cluster kernel doesnt like the JDK8 flavor, testing with 7 now

kevin-bates:
We need to revisit this since the oracle version referenced is no longer available and they've changed their licensing.

@lresende may have had to address the scala issue in toree so he might have some insight.

kevin-bates:
It might be easier to just use the demo-base image - which already uses openjdk and spark 2.4.1, etc.

kevin-bates:
Hi @akchinSTC - do we still need to spend time on this PR?  Just asking since 1.x is not that active and master demo-base was [updated to use openjdk](https://github.com/jupyter/enterprise_gateway/blob/master/etc/docker/demo-base/Dockerfile#L51) 16 months ago.

akchinSTC:
nope, im closing it

kevin-bates:
RIght on - thanks.

PR #529, opened 2018-12-13
Add Dask on YARN support

jcrist:
First, I was surprised how easy this was to add. The flexibility of the kernelspecs for launching kernels was quite nice.

This adds support for running [Dask](http://dask.pydata.org/) on YARN, which allows creating a remote kernel on YARN as part of a running Dask cluster.

To accomplish this I copied and modified the `launch_ipykernel.py` file. These changes/cleanups could be merged with the original file, adding a flag to create a dask cluster instead of a spark context. 
There's also no tests so far (not even sure if this PR is interesting to y'all). Happy to work further on this if desired.

To setup:

- Install `dask-yarn` (currently requires master branch, planning to release tomorrow) and the dependencies of `launch_ipykernel.py` (I think this is just `ipykernel`, `pycrypto`, `jupyter_client`) in a python environment on every node
- Install the `dask_python_yarn_remote` kernelspec as you would other kernelspecs
- Update the kernelspec appropriately. Fields of note would be `DASK_YARN_EXE`, the path to the `run.sh` file, and the python environment (path to the python environment setup above).
- Maybe configure the size of the workers/scheduler/client nodes (see [the dask-yarn docs](http://yarn.dask.org/en/latest/cli.html#dask-yarn-submit) for available fields).

I have this working locally in my [dockerized test cluster](https://github.com/jcrist/hadoop-test-cluster) and everything works fine, including kerberos authentication and impersonation.

Upon startup the kernel namespace is initialized with a `cluster` object corresponding to the existing YARN cluster. Users can then scale the cluster as needed using the scale methods or widget.


kevin-bates:
@jcrist - this is fantastic - thank you!  I've thought a Dask integration would be a good thing to have.  Great example of _good things come to those who wait_! :smiley:

There are a couple things that I think we should tweak and I'll add some comments via a review for that.

Thanks again!

etc/kernelspecs/dask_python_yarn_remote/scripts/launch_ipykernel.py
```diff
@@ -0,0 +1,285 @@
...
+
+        write_connection_file(fname=connection_file, ip=ip, key=key, shell_port=ports[0], iopub_port=ports[1],
+                              stdin_port=ports[2], hb_port=ports[3], control_port=ports[4])
+        if response_addr:
+            gateway_socket = return_connection_info(connection_file, response_addr, lower_port, upper_port)
+            if gateway_socket:  # socket in use, start gateway listener thread
+                gateway_listener_thread = Thread(target=gateway_listener, args=(gateway_socket,))
+                gateway_listener_thread.start()
+
+    namespace = initialize_namespace()
```
kevin-bates: Here's the other portion of [dynamic config determination](https://github.com/jupyter/enterprise_gateway/blob/master/etc/kernel-launchers/python/scripts/launch_ipykernel.py#L314-L324).

etc/kernelspecs/dask_python_yarn_remote/kernel.json
```diff
@@ -0,0 +1,23 @@
+{
+  "language": "python",
+  "display_name": "Dask - Python (YARN Remote Mode)",
+  "metadata": {
+    "process_proxy": {
+      "class_name": "enterprise_gateway.services.processproxies.yarn.YarnClusterProcessProxy"
```
kevin-bates: Have you confirmed that cells can be interrupted correctly?  To confirm, try running the following cell and attempt to interrupt within 60 seconds.  You should get `KeyboardInterrupt` exception.
```
import time
print('begin')
time.sleep(60)
print('end')
```
Also, is the kernel's shutdown happening gracefully?  You'll probably need to check the EG log for that. If Dask/Yarn simply honors the YARN REST api, then it probably will shutdown cleanly.  Given you're able to use the notebook probably means we're good here.  (That's really cool!)

jcrist: Yes, this works fine as is.

jcrist:
Thanks for the review, I believe all needs have been addressed. I added a new flag to the launcher `--RemoteProcessProxy.cluster-type`. This takes either `spark` (default), `dask`, or `none`. This *could* deprecate the need for `--RemoteProcessProxy.spark-context-initialization-mode`, as `none` for that is equivalent to `none` for cluster-type, and both `eager` and `lazy` are equivalent to `spark`. I left that alone for now.

This also cleans up how namespace initialization happens for spark. Currently the kernels share the same namespace as the launcher script, which means that the initial namespace is full of unrelated imports. I've changed this to use a specified namespace, and cleaned up how that namespace is initialized in the background. Note that I haven't tested this locally as I lack a spark setup, but could set one up if needed.

Finally, I added some basic flake8 configuration in setup.cfg, and ensured that `launch_ipykernel.py` passes flake8 with this configuration. I find that flake8 helps me catch common errors before testing, and figured you wouldn't mind this addition.

Please let me know if there's anything else needed here.

etc/Makefile
```diff
@@ -44,7 +44,7 @@ TOREE_LAUNCHER_FILES:=$(shell find kernel-launchers/scala/toree-launcher/src -ty
     # Seed the build tree with initial files
 	cp -r kernelspecs ../build
     # Distribute language and config-sensitive files.
-	@echo ../build/kernelspecs/{spark_,}python_* | xargs -t -n 1 cp -r kernel-launchers/python/scripts
+	@echo ../build/kernelspecs/{spark_,dask_}python_* | xargs -t -n 1 cp -r kernel-launchers/python/scripts
```
kevin-bates: We need to preserve the trailing comma, otherwise directories beginning with `python_` won't get their launchers.

jcrist: Thanks, fixed.

etc/kernelspecs/dask_python_yarn_remote/kernel.json
```diff
@@ -0,0 +1,25 @@
...
+    }
+  },
+  "env": {
+    "DASK_YARN_EXE": "/home/testuser/miniconda/envs/dask-env/bin/dask-yarn",
+    "DASK_OPTS": "--name ${KERNEL_ID:-ERROR__NO__KERNEL_ID} --environment python:///opt/anaconda2/bin/python --temporary-security-credentials --deploy-mode remote",
+    "LAUNCH_OPTS": ""
+  },
```
lresende: Do you need to submit some YARN specific configuration similar to what we have for [Spark on YARN ](https://raw.githubusercontent.com/jupyter/enterprise_gateway/master/etc/kernelspecs/spark_R_yarn_cluster/kernel.json)?

etc/kernelspecs/dask_python_yarn_remote/bin/run.sh
```diff
@@ -0,0 +1,25 @@
...
+echo
+echo "Starting IPython kernel for Dask ${USER_CLAUSE}"
+echo
+
+PROG_HOME="$(cd "`dirname "$0"`"/..; pwd)"
+
+set -x
+eval exec \
+     "${DASK_YARN_EXE}" submit \
+     "${DASK_OPTS}" \
```
lresende: Minimal, but EXE seems like Windows, how about _SCRIPT ?

etc/kernelspecs/dask_python_yarn_remote/kernel.json
jcrist: The CLI executable here is the [`dask-yarn` CLI](http://yarn.dask.org/en/latest/cli.html) and is already YARN specific.

etc/kernelspecs/dask_python_yarn_remote/bin/run.sh
jcrist: It's not really a script. Perhaps `DASK_YARN_CLI` or `DASK_YARN`?

etc/kernelspecs/dask_python_yarn_remote/kernel.json
lresende: Let's leave like exe then for now.

etc/kernelspecs/dask_python_yarn_remote/bin/run.sh
lresende: Let's leave like it is for now

PR #530, opened 2018-12-24
Adding the ability to use custom version of the KernelSessionManager class

lamOrigin007:
Hello! I made a small fix to add the ability to connect through the parameters of the command line a custom version of the KernelSessionManager class. For example, in order to be able to realize saving sessions not only in files, but also in SQL and NOSQL databases.


lamOrigin007:
That's all that i wanted to add at the moment. I will be glad to add an example of a separate MongoDB class as soon as it will be ready.

kevin-bates:
Sounds good.  Thanks!

PR #536, opened 2019-01-07
give availability to explicitly specify host for backconnect

xhumanoid:
required for run EG in docker on separate host

problem:
EG always announce local_ip, which doesn't match with host ip

example of installation:
1) docker compose with EG on one host
2) kernels running on separate hosts

solution:
use the same techniques as used in kafka and ElasticSearch
provide way to explicitly define hostname which will be announcement 

EG_ADVERTISED_HOST_LISTENER may require specify EG_PORT_RANGE and port mapping for this range

### DIFF
--- a/enterprise_gateway/services/processproxies/processproxy.py
+++ b/enterprise_gateway/services/processproxies/processproxy.py
@@ -40,6 +40,7 @@ poll_interval = float(os.getenv('EG_POLL_INTERVAL', '0.5'))
 socket_timeout = float(os.getenv('EG_SOCKET_TIMEOUT', '5.0'))
 tunneling_enabled = bool(os.getenv('EG_ENABLE_TUNNELING', 'False').lower() == 'true')
 ssh_port = int(os.getenv('EG_SSH_PORT', '22'))
+advertised_host = os.getenv('EG_ADVERTISED_HOST_LISTENER', None)
 
 # Minimum port range size and max retries
 min_port_range_size = int(os.getenv('EG_MIN_PORT_RANGE_SIZE', '1000'))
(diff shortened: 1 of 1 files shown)


enterprise_gateway/services/processproxies/processproxy.py
```diff
@@ -40,6 +40,7 @@
 socket_timeout = float(os.getenv('EG_SOCKET_TIMEOUT', '5.0'))
 tunneling_enabled = bool(os.getenv('EG_ENABLE_TUNNELING', 'False').lower() == 'true')
 ssh_port = int(os.getenv('EG_SSH_PORT', '22'))
+advertised_host = os.getenv('EG_ADVERTISED_HOST_LISTENER', None)
```
kevin-bates: Please change the env variable name to `EG_RESPONSE_IP` and add an entry about it to https://github.com/jupyter/enterprise_gateway/blob/master/docs/source/config-options.md#addtional-supported-environment-variables.  
Please mention this should be used when kernels reside in a different logical network than the EG container or pod.

It would also be helpful to describe how the appropriate IP address is determined - public IP, etc?

enterprise_gateway/services/processproxies/processproxy.py
```diff
@@ -593,7 +594,7 @@ def _prepare_response_socket(self):
         self.log.debug("Response socket launched on {}, port: {} using {}s timeout".format(local_ip, port, socket_timeout))
```
kevin-bates: The debug statement should reflect the proper address.  Probably best to move the debug statement and reformat (i.e., remove the `, port:{}` entry).

kevin-bates:
Hi Alexey - did you want to make the minor changes?  I'm happy to apply the rename pieces during the commit and can commit the doc update in another step if you like.

kevin-bates:
Merged with minor changes. Will update docs with EG_RESPONSE_IP entry separately.

Thank you for the contribution @xhumanoid.

PR #546, opened 2019-01-14
Convert spark python kernel to python 3

### LINKED ISSUE #545: Add support for Python3 to kernel-spark-py image
Currently, the `kernel-spark-py` image only supports Python 2 despite the fact that the underlying `spark-py` image installs both Python 2 and Python 3.  Since all other python-based images are Python 3, we should make `kernel-spark-py` default to Python 3 and (optionally) create a kernelspec that configures `kernel-spark-py` to run Python 2 (or build an explicit python2 image if necessary).

kevin-bates:
Update kernel-spark-py Dockerfile to install pip3 and necessary
files required by kernel.  Updated the corresponding kernel.json
to set the appropriate spark option to use python version 3.

This image can also be used for python 2.  Just create a new (or
update the existing) kernel.json to set version to 2.

Fixes #545


etc/docker/kernel-spark-py/Dockerfile
```diff
@@ -1,7 +1,10 @@
 FROM elyra/spark-py:v2.4.0
 
-RUN apk add --no-cache build-base libffi-dev openssl-dev python-dev && \
-    pip install cffi ipykernel ipython jupyter_client pycrypto
+RUN apk del python3 && \
```
akchinSTC: Did we need to remove python3? The next line we install python3-dev in which python3 is a dependency, so I think we would end up re-downloading it anyway

akchinSTC: thanks for the explanation, LGTM

PR #554, opened 2019-01-18
rewrite username to follow naming conversion

### LINKED ISSUE #553: Failed to start kernel if username contains .
As name of namespace only allows lower case alphanumber and -, it has to rewrite user name by change L164 to:

namespace = re.sub('[^0-9a-z]+', '-', kernel_username).lower() + '-' + self.kernel_id

https://github.com/jupyter/enterprise_gateway/blob/69807950e4bff399bb73ac41f74cf6be4b4b055a/enterprise_gateway/services/processproxies/k8s.py#L164

ktong:
fix #553.

### DIFF
--- a/enterprise_gateway/services/processproxies/k8s.py
+++ b/enterprise_gateway/services/processproxies/k8s.py
@@ -41,7 +41,7 @@ class KubernetesProcessProxy(ContainerProcessProxy):
         # Kubernetes relies on many internal env variables.  Since EG is running in a k8s pod, we will
         # transfer its env to each launched kernel.
         kwargs['env'] = dict(os.environ, **kwargs['env'])  # FIXME: Should probably use process-whitelist in JKG #280
-        self.kernel_pod_name = self._determine_POD_name(**kwargs)
+        self.kernel_pod_name = self._determine_pod_name(**kwargs)
         self.kernel_namespace = self._determine_kernel_namespace(**kwargs)  # will create namespace if not provided
 
         return super(KubernetesProcessProxy, self).launch_process(kernel_cmd, **kwargs)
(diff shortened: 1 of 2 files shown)


enterprise_gateway/services/processproxies/k8s.py
```diff
@@ -161,7 +161,7 @@ def _create_kernel_namespace(self, kernel_username, service_account_name):
         # the namespace's default ServiceAccount.  Since this is always done when creating a namespace, we can
         # delete the RoleBinding when deleting the namespace (no need to record that via another member variable).
 
-        namespace = kernel_username + '-' + self.kernel_id
+        namespace = re.sub('[^0-9a-z]+', '-', kernel_username).lower() + '-' + self.kernel_id
```
kevin-bates: I think the `.lower()` should be inside the substitution method so that username is lower-cased prior.  Otherwise a user name of `Foo.bar` will yield `-oo-bar` instead of `foo-bar`.

However, the kubernetes docs state that namespace names are a bit more restrictive.  From their error message text...
> a DNS-1123 label must consist of lower case alphanumeric characters or '-', and must start and end with an alphanumeric character

So a KERNEL_USERNAME of "-foo" isn't valid.  As a result, I think we should just drop the offending characters, so "-Foo.bar-dud" will yield: "foobardud" via 
```python
re.sub('[^0-9a-z]+', '', kernel_username.lower())
```
Ideally, we should allow embedded hyphens to be preserved (e.g., `foobar-dud`) but I wasn't able to figure out an RE to do that.  If you can, that would be great but I wouldn't knock yourself out.

In my local testing, I needed to `import re`.  Did you not find this necessary?

ktong:
```
        namespace = re.sub('[^0-9a-z]+', '-', kernel_username.lower()) + '-' + self.kernel_id
        while namespace.startswith('-'):
            namespace = namespace[1:]
```
would remove heading -s and without deleting other -s.

enterprise_gateway/services/processproxies/k8s.py
```diff
@@ -122,6 +125,16 @@ def terminate_container_resources(self):
                              "not been terminated.".format(self.kernel_namespace, self.container_name, self.kernel_id))
         return result
 
+    def _determine_kernel_pod_name(self, **kwargs):
+       pod_name = kwargs['env'].get('KERNEL_POD_NAME')
+        if pod_name is None:
```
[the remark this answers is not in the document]
ktong: I did not find good place to put the doc for KERNEL_POD_NAME. The best place probably is section Per-kernel Environment Overrides in config-options.md. I left it to you to add it.

ktong:
I will test it tomorrow

ktong:
I tested the _determine_kernel_pod_name method on python 3:

_determine_kernel_pod_name(env = {'KERNEL_POD_NAME': '.k.tong-fdsafa-'})
_determine_kernel_pod_name(env = {'KERNEL_USERNAME': '.k.tong'}) when .kernel_id='fdsfdsa-'

and both execution return 'k-tong-fdsfdsa'

I did not test it in k8s cluster as make bdist throws error:

make bdist
make dist/jupyter_enterprise_gateway-2.0.0.dev1-py2.py3-none-any.whl
make[1]: *** No rule to make target `venv/lib/python3.7/site-packages/nbconvert/preprocessors/tests/files/Disable', needed by `dist/jupyter_enterprise_gateway-2.0.0.dev1-py2.py3-none-any.whl'.  Stop.
make: *** [bdist] Error 2

ktong:
It's probably as I did not install conda in my env.  I used python3 -m venv venv && source venv/bin/activate && pip install -e . && make dist.

kevin-bates:
Yeah, I suspect this is due to the env.  You might try `make ENV=venv dist`.  At any rate, thanks for checking things out.  I'll approve and merge.

PR #558, opened 2019-01-21
Add support for run-as-UID/GID in containers

### LINKED ISSUE #548: Add support for running k8s kernel pods as different UIDs
We should look at supporting the following in the `kernel-pod.yaml` file:
```yaml
spec:
  securityContext:
    runAsUser: ${kernel_uid}
    fsGroup: ${kernel_gid}
```
to allow users to specify the user to run as by setting `KERNEL_UID` (and `KERNEL_GID`).

EnterpriseGateway should have a rule that, by default, prevents UID values of 0 (root).  This could be done via a `EG_UID_BLACKLIST` that specifies a list of UID values to disallow. Similar for GID.
 
KERNEL_UID would default to 1000, KERNEL_GID to 100.  Creation requests encountering a blacklisted UID/GID would fail with 403.

Installations wishing to allow root access could set their EG_UID_BLACKLIST to -1 - but this requires access to the EG instance creation - which is assumed secure.

I'm not sure this is doable for spark-based pods.

kevin-bates:
Added basic support for running as user id and group id values.  These values
can be specified via KERNEL_UID and KERNEL_GID respectively.  The defaults for
these values are 1000 (jovyan) and 100 (users) since that's how the containers
are built.  However, note that on k8s, use of `runAsGroup` requires that
feature-gate `RunAsGroup` first be enabled.  Failure to enable the feature-gate,
yet use either `runAsUser` or `runAsGroup` will result in the 0 (root) group
being the effective group of the container.  As a result, use of these attributes
are commented out in the `kernel-pod.yaml` file with an appropriate comment.

The docker launcher has also been extended for these applicable values in docker.

Although Kubernetes PodSecurityPolicy objects can be used to ensure that the
root user/group is not used, a PSP affects ALL pods in the cluster, so we do
not want to introduce those in our yaml as that will break other applications.
UID and GID blacklists are sepecified via EG_{UID,GID}_BLACKLIST - both defaulting
to 0 (root).

Fixes #548

### DIFF
--- a/docs/source/config-options.md
+++ b/docs/source/config-options.md
@@ -283,28 +283,35 @@ JupyterWebsocketPersonality options
 ```
 
 ### Addtional supported environment variables
-```
+```text
   EG_DEFAULT_KERNEL_SERVICE_ACCOUNT_NAME=default
-    Kubernetes only.  This value indicates the default service account name to use for kernel
-    namespaces when the Enterprise Gateway needs to create the kernel's namespace and
-    KERNEL_SERVICE_ACCOUNT_NAME has not been provided.
+    Kubernetes only.  This value indicates the default service account name to use for
+    kernel namespaces when the Enterprise Gateway needs to create the kernel's namespace
+    and KERNEL_SERVICE_ACCOUNT_NAME has not been provided.
 
   EG_DOCKER_NETWORK=enterprise-gateway or bridge
-    Docker only. Used by the docker deployment and launch scripts, this indicates the name of the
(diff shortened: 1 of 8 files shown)


kevin-bates:
We should transfer the uid and gid values to the NB_ values and possibly run fix_permissions from the command scripts at container startup.

kevin-bates:
It isn't going to help to transfer the uid/gid to NB values since the containers already run as non-root.  We can't change permissions.  Instead, we make sure to add `users` (100) to the supplemental groups - which enables `/home/jovyan` to be writable.

PR #568, opened 2019-02-01
Handling session persistence and added env config for session persistence

Gsbreddy:


kevin-bates:
Hi Sai,

Please amend your commit with a rebase of the remote master branch.  We merged a few PRs yesterday, so there's unnecessary noise showing up.

I've gone ahead and added some in-line comments that I'm hoping can be addressed during the amend.  We can proced once I can clearly see what has changed.

Thank you!

enterprise_gateway/services/sessions/kernelsessionmanager.py
```diff
@@ -30,8 +30,15 @@ class KernelSessionManager(LoggingConfigurable):
         state consistent.
     """
 
-    enable_persistence = Bool(default_value=False, config=True,
-        help="""Enable kernel session persistence.  Default = False""")
+    # Session Persistence
+    session_persistence_env = 'EG_SESSION_PERSISTENCE'
```
kevin-bates: I'd like to include `KERNEL` in the config option's env variable so that it is clear this only pertains to _kernel sessions_ and not the more common _notebook sessions_.  This difference is conveyed in the command line option due to the class prefix of `KernelSessionManager`.  I think using `EG_KERNEL_SESSION_PERSISTENCE` is appropriate.

Gsbreddy: Sure thing.

docs/source/config-options.md
```diff
@@ -249,6 +249,10 @@ EnterpriseGatewayApp options
     Default: False
     Is YARN Kerberos/SPNEGO Security enabled (True/False).
     (EG_YARN_ENDPOINT_SECURITY_ENABLED env var)
+--EnterpriseGatewayApp.KernelSessionManager.enable_persistence=<Bool>
```
kevin-bates: Does this work?  Will this show up when running `juptyer-enterprisegateway --help-all`?

I know that using `KernelSessionManager.enabled_persistence=True` works.  Not sure how this changes things unless I'm missing another change or wasn't aware that internal classes could be appended to the app name for this.

kevin-bates:
Note that this PR was manually closed and "replaced" with #569.

PR #569, opened 2019-02-01
added EG_KERNEL_SESSION_PERSISTENCE env and modified kernel persisten…

Gsbreddy:
…ce handling changes from #568

### DIFF
--- a/enterprise_gateway/services/sessions/kernelsessionmanager.py
+++ b/enterprise_gateway/services/sessions/kernelsessionmanager.py
@@ -8,9 +8,9 @@ import json
 import os
 import threading
 
-from ipython_genutils.py3compat import (bytes_to_str, str_to_bytes)
+from ipython_genutils.py3compat import (unicode_to_str, str_to_unicode)
 from jupyter_core.paths import jupyter_data_dir
-from traitlets import Bool
+from traitlets import Bool, default
 from traitlets.config.configurable import LoggingConfigurable
 
 kernels_lock = threading.Lock()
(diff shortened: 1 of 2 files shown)


enterprise_gateway/services/sessions/kernelsessionmanager.py
```diff
@@ -181,7 +188,7 @@ def _pre_save_transformation(sessions):
                 info = session['connection_info']
                 key = info.get('key')
                 if key:
-                    info['key'] = bytes_to_str(key)
+                    info['key'] = unicode_to_str(key)
```
kevin-bates: The `key` should be handled as bytes.  Is there a reason you found this change to be necessary?

Gsbreddy: Yes, I was getting error saying `AttributeError: 'str' object has no attribute 'decode'` from ~/anaconda2/envs/ana35/lib/python3.5/site-packages/ipython_genutils/py3compat.py:157

Gsbreddy:
Also, I don't understand why the builds are failing with error  `AttributeError: 'bool' object has no attribute 'lower'`  because lower is actually written inside of the `bool` block. Can you please check this.

kevin-bates:
Ok. I'll try to take a look into both of the issues (str_to_bytes and bool/lower). Do you have the line of code relative to the bool/lower issue?

kevin-bates:
Strange.  I updated the requirements.yml file to build on 3.5 and changed the default value to True - so that the code in question would be executed, and [the build passed](https://travis-ci.org/kevin-bates/enterprise_gateway/builds/487945857?utm_source=github_status&utm_medium=notification).

I would suggest trying to create a fresh conda env using the requirements.yml file and try again.

Btw, using `unicode_to_str()` produces a no-op for python 3.x code.  `key` should be of type `bytes` when that code is invoked.

Gsbreddy:
> using `unicode_to_str()` produces a no-op for python 3.x code

What do you suggest @kevin-bates . Because I just tested it im PY35.

enterprise_gateway/services/sessions/kernelsessionmanager.py
Gsbreddy: Anything regarding this error?

kevin-bates:
Have you tried creating a fresh conda env using only [requirements.yml](https://github.com/jupyter/enterprise_gateway/blob/master/requirements.yml)?  That is, emulate what I did using travis by enabling persistence via the env you added?  I was able to confirm via a debugger that the value of `key` at that line was `bytes`.  

I think you need to confirm that `key` is indeed a `str`, then figure out why its a string in your case.  You can then also use the nosetests to confirm everything works when `bytes_to_str()` is used.

Gsbreddy:
> Have you tried creating a fresh conda env using only [requirements.yml](https://github.com/jupyter/enterprise_gateway/blob/master/requirements.yml)? That is, emulate what I did using travis by enabling persistence via the env you added? I was able to confirm via a debugger that the value of `key` at that line was `bytes`.
> 
> I think you need to confirm that `key` is indeed a `str`, then figure out why its a string in your case. You can then also use the nosetests to confirm everything works when `bytes_to_str()` is used.

Hi @kevin-bates , I debugged and noticed following:
So what was happening is, in version 1.1.1 the key conversion was being handled in `remotemanager.py `(conversion to bytes and str line:209 and line:217). So after this step, the key is converted to str. That's the reason key type is str in KernelSessionManager.

But in the latest master code i.e., dev, this handling is not being done in remotemanager. Hence, the key type you see in KernelSessionManager is bytes and the conversion is being handled in KernelSessionManager itself.

kevin-bates:
@Gsbreddy - I thought there was a change that switched things back to use `bytes_to_str`, but I don't see that commit.  Did a commit get removed - or am I just seeing things in my old age? :smile:

PR #572, opened 2019-02-07
Pull correct version of spark entrypoint script in spark based kernel images

akchinSTC:
- Fix spark kernel images so it pulls entrypoint.sh from githubs $SPARK_VER tag
- Fix issue with kernel-spark-r image where it could not locate ssl lib
- Change kernel-py Dockerfile to install packages with conda instead of pip


kevin-bates:
I made a slight change to combine the linking to the anaconda tini with the previous RUN statement to eliminate a layer and made sure the corresponding images ran correctly.

PR #580, opened 2019-02-19
[DEPENDENCY][WIP] Add support for concurrent kernel startup (and restarts)

### LINKED ISSUE #86: Remove blocking Kernel Startup
Single-threaded kernel startup causing notebook front end to lose connection to gateway as a lot of kernels are being started--so that notebook has to re-pull kernel specs from gateway after Elyra finishes starting kernels.

kevin-bates:
Updated start_kernel and restart_kernel methods to coroutines.

Converted various `time.sleep()` to `yield gen.sleep()` which enables tornado to "slice-out" to a different request. As a result, converting `confirm_remote_startup()` and `handle_timeout()` to coroutines was key.

**NOTE: Installation of this PR requires that BOTH related [Jupyter Client](https://github.com/jupyter/jupyter_client/pull/425) and [Notebook](https://github.com/jupyter/notebook/pull/4412) PRs be merged.  Once merged, we need to update our dependencies to depend on those versions - thus the bracketed values in the title.**

Fixes #86


kevin-bates:
@esevan - great question.  

This change needs to be back-ported to our 1.x branch - so that's one reason for not moving to the 3.x constructs.  In addition, we probably shouldn't make that switch until Notebook and jupyter_client have done the same.  I don't know if using `async/await` in one class will work with `coroutine/yield` in a superclass or vice versa.  Do you know if that works?

For the time being, I believe we should continue using backward compatible code until we know we're going to not go back or the benefit of using the more native constructs is critical to the implementation's success.

The most important aspect to this PR is that the dependent PRs get merged AND are part of releases, so we can then update our dependencies in `setup.py`.

Thanks.

kevin-bates:
The dependent areas of jupyter_client and notebook are in the process of getting reworked due to compatibility issues of the original PRs.  A subclassed approach is being looked at and we should have good information by the end of this week.  Stay tuned.

kevin-bates:
I've updated the description (along with the referenced PRs) to reflect the subclassed based approach introduced in the referenced PRs.

The other thing to note is that we probably lose support for `http-mode` that comes from JKG since we're now skipping those classes.  Should we decide otherwise, then we'll need to create async-related classes in JKG and update EG to reference the async parent.  Not a big deal, but since we're thinking about implementing our own web handlers for HA support, it might be good separate the two.

kevin-bates:
If anyone wants to take things for a spin, here's a zip of `whl` files with relative changes in jupyter_client, notebook and enterprise_gateway.   

[AsyncKernelManagement-wheels.zip](https://github.com/jupyter/enterprise_gateway/files/2973266/AsyncKernelManagement-wheels.zip)

I'll update the EG image shortly.

kevin-bates:
An EG image built on the referenced PRs can be found at [`elyra/enterprise-gateway`](https://hub.docker.com/r/elyra/enterprise-gateway) and pulled via the following:
```
docker pull elyra/enterprise-gateway:dev_async
```

meeseeksmachine:
This pull request has been mentioned on **Jupyter Community Forum**. There might be relevant details there:

https://discourse.jupyter.org/t/scalable-enterprise-gateway/2014/2

PR #583, opened 2019-02-21
Fix k8s kernel restarts when EG creates kernel namespace

### LINKED ISSUE #578: Kubernetes: Kernel restarts fail due to erroneous namespace creation attempt
When EG needs to create a kernel's namespace, restarts for that kernel fail because EG wants to create a namespace for the restarted kernel but gets an existence error (409).  The process proxy can determine a kernel is restarting by checking `self.kernel_manager.restarting` and should probably skip the creation attempt.

Here's the applicable stack:
```
      File "/usr/lib/python2.7/site-packages/jupyter_client/manager.py", line 251, in start_kernel
        self.kernel = yield gen.maybe_future(self._launch_kernel(kernel_cmd, env=env, **kw))
      File "/usr/lib/python2.7/site-packages/enterprise_gateway/services/kernels/remotemanager.py", line 279, in _launch_kernel
        return self.process_proxy.launch_process(kernel_cmd, **kwargs)
      File "/usr/lib/python2.7/site-packages/enterprise_gateway/services/processproxies/k8s.py", line 45, in launch_process
        self.kernel_namespace = self._determine_kernel_namespace(**kwargs)  # will create namespace if not provided
      File "/usr/lib/python2.7/site-packages/enterprise_gateway/services/processproxies/k8s.py", line 160, in _determine_kernel_namespace
        namespace = self._create_kernel_namespace(service_account_name)
      File "/usr/lib/python2.7/site-packages/enterprise_gateway/services/processproxies/k8s.py", line 208, in _create_kernel_namespace
        self.log_and_raise(http_status_code=500, reason=reason)
      File "/usr/lib/python2.7/site-packages/enterprise_gateway/services/processproxies/processproxy.py", line 624, in log_and_raise
        raise web.HTTPError(status_code=http_status_code, reason=reason)
    HTTPError: HTTP 500: Error occurred creating namespace 'itest21744-06c11849-e66d-44b0-b0d0-15a261712d22': (409)
    Reason: Conflict
    HTTP response headers: HTTPHeaderDict({'Date': 'Mon, 18 Feb 2019 19:57:25 GMT', 'Content-Length': '309', 'Content-Type': 'application/json'})
    HTTP response body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"object is being deleted: namespaces \"itest21744-06c11849-e66d-44b0-b0d0-15a261712d22\" already exists","reason":"AlreadyExists","details":{"name":"itest21744-06c11849-e66d-44b0-b0d0-15a261712d22","kind":"namespaces"},"code":409}
```

kevin-bates:
When EG creates the kernel pod's namespace (default behavior), restarts
would fail because the restart encouters an existing (but terminating)
namespace during the start portion of the restart (with the shutdown portion
having triggered the namespace's termination).  This change only deletes
the kernel pod during shutdowns associated with restarts, then tolerates
the namespace's existence after creation.  Note we still attempt the
namespace creation in case it was deleted outside of EG and a kernel
restart was attempted to rectify the situation.  This results in restart
behaviors identical to that if the user brought their own namespace.

Fixes #578

### DIFF
--- a/enterprise_gateway/services/processproxies/k8s.py
+++ b/enterprise_gateway/services/processproxies/k8s.py
@@ -79,13 +79,13 @@ class KubernetesProcessProxy(ContainerProcessProxy):
     def terminate_container_resources(self):
         """Terminate any artifacts created on behalf of the container's lifetime."""
         # Kubernetes objects don't go away on their own - so we need to tear down the namespace
-        # or pod associated with the kernel.  If we created the namespace, then that's our target,
-        # else just delete the pod.
+        # or pod associated with the kernel.  If we created the namespace and we're not in the
+        # the process of restarting the kernel, then that's our target, else just delete the pod.
 
         result = False
         body = client.V1DeleteOptions(grace_period_seconds=0, propagation_policy='Background')
 
-        if self.delete_kernel_namespace:
+        if self.delete_kernel_namespace and not self.kernel_manager.restarting:
             object_name = 'namespace'
         else:
             object_name = 'pod'
@@ -93,11 +93,11 @@ class KubernetesProcessProxy(ContainerProcessProxy):
         # Delete the namespace or pod...
         try:
             # What gets returned from this call is a 'V1Status'.  It looks a bit like JSON but appears to be
-            # intentionally obsfucated.  Attempts to load the status field fail due to malformed json.  As a
+            # intentionally obfuscated.  Attempts to load the status field fail due to malformed json.  As a
             # result, we'll see if the status field contains either 'Succeeded' or 'Failed' - since that should
             # indicate the phase value.
 
-            if self.delete_kernel_namespace:
+            if self.delete_kernel_namespace and not self.kernel_manager.restarting:
                 v1_status = client.CoreV1Api().delete_namespace(name=self.kernel_namespace, body=body)
             else:
                 v1_status = client.CoreV1Api().delete_namespaced_pod(namespace=self.kernel_namespace,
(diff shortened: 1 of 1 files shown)


enterprise_gateway/services/processproxies/k8s.py
```diff
@@ -197,15 +197,19 @@ def _create_kernel_namespace(self, service_account_name):
...
             self._create_role_binding(namespace, service_account_name)
         except Exception as err:
-            if self.delete_kernel_namespace:
-                reason = "Error occurred creating role binding for namespace '{}': {}".format(namespace, err)
-                # delete the namespace since we'll be using the EG namespace...
-                body = client.V1DeleteOptions(grace_period_seconds=0, propagation_policy='Background')
-                client.CoreV1Api().delete_namespace(name=namespace, body=body)
-                self.log.warning("Deleted kernel namespace: {}".format(namespace))
+            if isinstance(err, client.rest.ApiException) and err.status == 409 and self.kernel_manager.restarting:
+                self.delete_kernel_namespace = True  # okay if ns already exists and restarting, still mark for delete
```
esevan: What is the case of this line?
As your fix lands, there's no namespace creation while restarting kernel, is there?

kevin-bates: @esevan - correct, we will not delete a kernel pod's namespace during the shutdown phase of a kernel restart so we will tolerate its existence (409) during the start phase of a kernel restart.

You may have noticed that Issue #578 references a `309` error code, not `409`.  Actually, I found both to occur during my analysis - due to timing.  `409` occurs if the object being created already exists, while `309` occurs if the object being created happens to be _terminating_.  I believe there was a race condition occurring relative to the namespace deletion that used to occur during the shutdown phase.  Since that deletion is no longer attempted, then `409` (already exists) is the error code we should tolerate if the kernel is restarting.

So the net result is that a kernel pod's namespace behaves identical to if the user brought their own namespace via `KERNEL_NAMESPACE` relative to kernel restarts.  However, we still want to _ultimately_ delete the namespace when the kernel is finally shutdown - so we set the instance's member `self.delete_kernel_namespace = True` - since we're in the process of creating a new instance of this process proxy.

esevan: Okay I thought it was better to check if kernel is restarting before creating namespace, but I agree with timing issue can happen.

+1, Is there any contribution rule to approve PR?

kevin-bates: There's no rule for approval.  Your review is appreciated!

lresende: @esevan just do a comment with LGTM, or even better, review and approve the PR via the Github UI.

PR #587, opened 2019-02-27
Remove support for Python 2.x

lresende:
Todo:
- [x] Update demo docker image to use Python 3.x #590
- [x] Update EG docker image to derive from minimal-notebook rather than spark base image #589
- [x] Update kernelspec python version references #590

Issue #573, opened 2019-02-11
Remove Python 2.x support on EG 2.0.0 release


lresende:
Various other Jupyter projects are only supporting Python 3.5 and above, and Enterprise Gateway should also move to that direction.

kevin-bates:
It will also be nice to remove the flakey python 3.4 from travis.  When doing so, we should see how 3.7 works (it wasn't quite ready 6 months ago or so) and should probably add python 3.5, leaving us with 3.5, 3.6 and 3.7(?).

lresende:
@kevin-bates Yes to all !!!

lresende:
And another support for going into this direction: https://python3statement.org/

PR #590, opened 2019-03-01
Update to python3 and use version agnostic conda paths

akchinSTC:
- Update miniconda / python to version 3
- Use /opt/conda as installation prefix for miniconda

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,7 @@ sudo: required
 language: python
 
 python:
-    - 2.7
-    - 3.4
+    - 3.5
     - 3.6
 
 services:
--- a/docs/source/getting-started-client-mode.md
+++ b/docs/source/getting-started-client-mode.md
@@ -50,9 +50,9 @@ After that, you should have a kernel.json that looks similar to the one below:
   },
   "env": {
     "SPARK_HOME": "/usr/hdp/current/spark2-client",
-    "PYSPARK_PYTHON": "/opt/anaconda3/bin/python",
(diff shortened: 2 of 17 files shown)


lresende:
@akchinSTC Do you want to merge #587 and take issue #573 ?

akchinSTC:
@lresende - copy that, assigned the issue to me and will merge your changes in with my PR and close both.

kevin-bates:
FWIW, I setup a python 3 env on my python 2.7 YARN cluster today.  

My base (python 2) configuration was located in `/opt/anaconda2`.  I then created a python 3 environment:
```
conda create --name py3 python=3.6
```
I then switched users (su) to the gateway user and switched envs to `py3`:
```
source /opt/anaconda2/bin/activate py3  (I could also use `conda activate py3`)
```
I then updated the start script for EG to remove the fully-qualified path prefix on the `jupyter` command so that it used that from the current env.  Then launched EG.

For those kernels that require python 2, I added the following to their `run.sh` scripts prior to the spark-submit call to switch back to python 2
```
source /opt/anaconda2/bin/activate base
```
I also added `source /opt/anaconda2/bin/deactivate` to the end of `run.sh`.

I was then able to launch a python kernel in YARN cluster mode.  Had the environment not be switched, python syntax errors will be produced from `/bin/hdp-select` since it doesn't use parentheses on its print statements.

lresende:
This has been merged to the [remove-python-2.x](https://github.com/jupyter/enterprise_gateway/tree/remove-python-2.x) branch for team collaboration.

PR #606, opened 2019-03-15
Increase shutdown wait time for YARN kernels

kevin-bates:
On YARN systems, there's a tendency for the shutdown times to exceed the default wait time of 5 seconds.  Rather than require customers that need to use YARN to set another command-line option, this change will increase the timeout only when `YarnClusterProcessProxy` instances are in play.  The increase only occurs if the current value is less than the desired value.   This value defaults to 15 seconds and can be adjusted via env `EG_YARN_SHUTDOWN_WAIT_TIME`.

If debug is enabled, the following will be logged at kernel startup and when adjustment is performed:
```
[D 2019-03-15 14:33:17.943 EnterpriseGatewayApp] YarnClusterProcessProxy shutdown wait time adjusted to 15.0 seconds.
```


lresende:
@kevin-bates, BTW, I should also merge this to 1.x before the release goes out?

kevin-bates:
@lresende - yes, it's safe to take for 1.x.  Thank you!

Issue #595, opened 2019-03-08
Add support for multiple YARN masters


kevin-bates:
When YARN is configured for HA, it requires that multiple masters be specified.  As a result, admins of EG should be able to specify multiple masters or none at all (in which case, the yarn-api-client library uses local Hadoop config files).

Currently EG requires a yarn_endpoint be configured `--EnterpriseGatewayApp.yarn_endpoint=http://hostname:port/ws/v1/cluster`, yet, now that the underlying library has been enhanced, EG only uses the hostname from that value.  Instead, we should add a configuration option of `--EnterpriseGatewayApp.yarn_masters` which is a _list-valued_ property, whose default value is the empty list.

If empty, the underlying library will use local configuration files.  If single entry, EG will use that host.  If multiples, we could either be optimistic and use the first one (with some recovery to the others) or we could ensure the one used is an active master via the [`ws/v1/cluster/info`](https://hadoop.apache.org/docs/current/hadoop-yarn/hadoop-yarn-site/ResourceManagerRest.html#Cluster_Information_API) REST API.  We should also retain the selected (or single) master in a static variable relative to the scope (remember that yarn hosts could appear in kernelspec configs) so that we start with that _verified_ value for the next applicable kernel.  (Unconditionally verifying the master when multiple exist, is probably the easiest and most stable approach.)

We should also try to switch from downed masters for already running kernels, otherwise things like termination, might fail - assuming the scope of that effort is not too great.

saipradeepkumar:
@kevin-bates 

as you said we can take the list of yarn endpoints.when ever we have to pick the yarn end point we can start connecting in the round robin fashion ,and pick the active from the list and assign it to the variable which we can use while getting the status.
for this we need two functions.
1) selecting the active from RM from the list
2)assigning the active yarn end point to the variable.

am i correct?

kevin-bates:
Yes.  However, we _may_ want to do this in the yarn-api-client layer since it already [uses the `/cluster` endpoint for the case when no address is specified](https://github.com/toidi/hadoop-yarn-api-python-client/blob/master/yarn_api_client/hadoop_conf.py#L47-L59).

That said, I like the idea with dealing with the list in EG because, ultimately it would be nice not to have to rely on local hadoop config files (although spark might need those regardless).  I'm not that familiar the spark configuration relative to YARN.

Adding @lresende and @akchinSTC in case they have input here.

lresende:
[...]

Issue #29, opened 2017-06-07
Create docker infrastructure to run integration tests


lresende:
Enables bootstrapping a Elyra environment using docker for integration test purposes

kevin-bates:
@lresende - we can start looking at this once PR #208 is merged.

PR #607, opened 2019-03-19
Add support for multiple YARN resource managers

kevin-bates:
Add support for specifying multiple YARN resource managers for HA-configured environments.  A new configuration option `alt_yarn_endpoint` compliments `yarn_endpoint` when YARN HA configurations are in use.  When both values are provided, the `yarn-api-client` library will the original config value for an active RM.  If not active, the library will check the alternate.

Note that both of these configuration values default to `None` (`yarn_endpoint` has been updated for this).  In cases where `yarn_endpoint` is `None`, the `yarn-api-client` library will read from the `yarn-site.xml` file located in `HADOOP_CONF_DIR` and choose the active resource manager - both HA and non-HA configurations.

This pull request is dependent on the [yarn-api-client PR #29](https://github.com/toidi/hadoop-yarn-api-python-client/pull/29).  Once merged and an updated version is available, [setup.py](https://github.com/jupyter/enterprise_gateway/blob/master/setup.py#L54) will be updated with the appropriate version.

Fixes: #595


kevin-bates:
Removed dependency and WIP status from description.  I suppose we still need to wait for yarn-api-client 0.3.5 in pypi.org.  Until then builds will fail (sorry, just realized this!)

PR #614, opened 2019-03-23
Increase write permissions on SPARK_HOME/work-dir

kevin-bates:
When changing the spark-based images to jupyter-based images, we dropped chown to `jovyan` for the SPARK_HOME/work-dir.  As a result, the jupyter-based images would get failures running spark jobs that required executors.  This change increases the permissions of the work-dir, not the owners, since we can run images as non-jovyan users today.

An image to test this with exists at: `kbates/kernel-spark-py:workdir`.


lresende:
And make the same change (numbers) on the R kernel image

kevin-bates:
Yeah, that makes sense.  It will need 0777 since the directory should be writeable by all as well as executable.  I believe these are already 0755.

PR #615, opened 2019-03-23
[HOTFIX] Fix 'make clean-images' build target

lresende:


etc/Makefile
```diff
@@ -99,7 +99,7 @@ kernel_image_files: ../build/kernel_image_files
 # Docker image build section ***********************************************
 #
 
-KERNEL_IMAGES := kernel-py kernel-spark-py kernel-r kernel-spark-r kernel-scala kernel-tf-py
+KERNEL_IMAGES := kernel-py kernel-spark-py kernel-r kernel-spark-r kernel-scala kernel-tf-py kernel-tf-gpu-py
```
kevin-bates: Adding this image back to this tag is going to trigger its publish via the publish target and include it in builds.  Anyone using that "hidden" image needs to use `make clean-kernel-tf-gpu-py` manually to clean that image.

kevin-bates: hmm - are you finding that even the manual attempt is not working - that's probably true.  So if we don't want this image published, then PUBLISHED_IMAGES shouldn't reference KERNEL_IMAGES.

lresende: I made some further tests, and because we have already previously removed the `publish-kernel-tf-gpu-py` from `publish-images` then it's not getting published. You can easily verify that by issuing the following command:
```
cd etc
make --dry-run publish-images
docker push elyra/enterprise-gateway-demo:
docker push elyra/nb2kg:
docker push elyra/enterprise-gateway:
docker push elyra/kernel-py:
docker push elyra/kernel-spark-py:
docker push elyra/kernel-tf-py:
docker push elyra/kernel-r:
docker push elyra/kernel-spark-r:
docker push elyra/kernel-scala:
```

Also, to be on the safe side, I could add the following as well
```
$(filter-out kernel-tf-gpu-py,$(KERNEL_IMAGES))
``` 

Thoughts?

kevin-bates: Sorry, you're right.  I was focused on KERNEL_IMAGES and forget the publish op used a different set of targets.

PR #616, opened 2019-03-24
Fix/debug restart tests

kevin-bates:
The travis builds have been failing due to timeouts lately.  Some timeouts are due to system issues, but many appear related to the restart tests, which were randomly triggering 10 minute travis timeouts.  As a result, the following changes have been made:
1. Added `itest-yarn-debug` and `itest-docker-debug` targets which set debug options for the nosetests.  Currently that sets `--nocapture --nologcapture` so that log statements from the test code are produced.  (GatewayClient and the kernel class have been updated to use loggers.)
1. Added some more log statements to produce messages regarding reception and process of a message (and its type).
1. After some analysis, it seems that error status's are not returned always returned, but the appropriate error message is embedded in the `execute_reply`, so I updated the test to look for the error response in `execute_reply` in addition to the straight `error` responses.  All other `execute_reply` messages (and `execute_input`) go unhandled.
1. The default timeout for requests was set to 1 hour!  We should have this value be less than 10 minutes, otherwise Travis will stop the test when, it could be that the message processing is hung up.  With the additional debug options in place, the timeout has been set to 5 minutes.  Now the only travis stops should be due to system-related issues.

### DIFF
--- a/enterprise_gateway/client/gateway_client.py
+++ b/enterprise_gateway/client/gateway_client.py
@@ -1,5 +1,6 @@
 
 import os
+import logging
 import requests
 import time
 from uuid import uuid4
@@ -12,10 +13,13 @@ try:  # prefer python 3, fallback to 2
 except ImportError:
     import Queue as queue
 
-DEFAULT_TIMEOUT = 60 * 60
+REQUEST_TIMEOUT = int(os.getenv("REQUEST_TIMEOUT", 120))
+log_level = os.getenv("LOG_LEVEL", "INFO")
+
+logging.basicConfig(format='[%(levelname)1.1s %(asctime)s.%(msecs).03d %(name)s] %(message)s')
 
 
-class GatewayClient:
+class GatewayClient(object):
     """
     *** E X P E R I M E N T A L *** *** E X P E R I M E N T A L ***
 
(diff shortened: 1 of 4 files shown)


kevin-bates:
Briefly switched to WIP to add a couple more changes to help with analysis.

enterprise_gateway/client/gateway_client.py
```diff
@@ -12,7 +12,7 @@
 except ImportError:
     import Queue as queue
 
-DEFAULT_TIMEOUT = 60 * 60
+DEFAULT_TIMEOUT = 60 * 5  # Use 5 minute timeout (should remain < 10 min to avoid travis timeouts
```
lresende: We should not set the default based on Travis, but enable iTests to pass their desired timeout otherwise, this will not work on real use cases where a big data invocation might take a little while to respond.

kevin-bates: I agree this should be more easily configured.  Upon closer inspection, this timeout is really a connection-oriented timeout and does not correspond to the overall length of a cell's processing.  In fact, its default should be on the order of 60 seconds.  I'll probably reset to 120 just to be on the safe side.

kevin-bates: In the latest update, this has been changed to `REQUEST_TIMEOUT` (with a default of 120) and can be updated via env `REQUEST_TIMEOUT`.

PR #623, opened 2019-04-01
Set YARN endpoint default to None

kevin-bates:
This pull request is a split of PR #607 that changes the default value for the YARN endpoint configuration option to `None`.  By doing this, 90% of customers can take advantage of built-in behaviors in `yarn-api-client` which determine the YARN Resource Manager endpoint from local configuration files.  This is because 90% of customers install Enterprise Gateway on an edge node of their YARN cluster.  In addition, that logic will also handle the case where YARN is configured for HA.

Installations on which Enterprise Gateway is not installed on an edge node, will still need to provide a YARN endpoint parameter.  In addition, these installations cannot take advantage of HA configurations until [yarn-api-client PR #29](https://github.com/toidi/hadoop-yarn-api-python-client/pull/29) is merged.

### DIFF
--- a/enterprise_gateway/services/processproxies/yarn.py
+++ b/enterprise_gateway/services/processproxies/yarn.py
@@ -41,34 +41,23 @@ class YarnClusterProcessProxy(RemoteProcessProxy):
         self.yarn_endpoint \
             = proxy_config.get('yarn_endpoint',
                                kernel_manager.parent.parent.yarn_endpoint)
-        self.alt_yarn_endpoint \
-            = proxy_config.get('alt_yarn_endpoint',
-                               kernel_manager.parent.parent.alt_yarn_endpoint)
-
         self.yarn_endpoint_security_enabled \
             = proxy_config.get('yarn_endpoint_security_enabled',
                                kernel_manager.parent.parent.yarn_endpoint_security_enabled)
 
-        yarn_master = alt_yarn_master = None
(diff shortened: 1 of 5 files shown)


enterprise_gateway/services/processproxies/yarn.py
```diff
@@ -37,22 +37,27 @@ class YarnClusterProcessProxy(RemoteProcessProxy):
...
-        else:
-            self.resource_mgr = ResourceManager(address=yarn_master,
-                                                port=yarn_port)
+
+        yarn_master = None
+        yarn_port = 8088
+        if self.yarn_endpoint:
+            yarn_url = urlparse(self.yarn_endpoint)
+            yarn_master = yarn_url.hostname
+            yarn_port = yarn_url.port
```
lresende: It seems that when yarn_master is none, the port will never be used. In this case, should we actually initialize the yarn_port as None?

kevin-bates: It doesn't matter since it will be pulled from the config. I was following the defaulting used in yarn client. That said it probably makes more sense to follow suit and pass none in these cases. This will be okay as long as yarn client only uses the port parameter when the address parameter is not none. I'll change this tomorrow.

kevin-bates: @lresende - update pushed.

Issue #625, opened 2019-04-03
Helm chart


witten:
It would be great to have an officially supported [Helm](https://helm.sh/) chart for the Kubernetes version of Jupyter Enterprise Gateway. For Kubernetes + Helm users, that would unlock a more structured and on-the-rails mechanism for packaging and distributing the gateway, setting individual variables like `EG_NAMESPACE`, deploying the gateway, etc. At least compared to the current approach of checking out source, hacking directly on YAML template files, and running `kubectl apply`.

kevin-bates:
@witten - this is a great idea.  Would you be willing to take stab at this?  

I'm not familiar with Helm - but it would probably be good to model _conventions_ (if that's applicable) after the charts used for JupyterHub.

witten:
Possibly.. It depends whether my organization ends up adopting Jupyter Enterprise Gateway! If we do, we'll need a Helm chart. And if there isn't one by then, we'd be happy to create one and share it here.

kevin-bates:
Excellent!  So how can we help your organization adopt Enterprise Gateway? 😃

Please feel free to ask questions and discuss various topics on our [discourse category](https://discourse.jupyter.org/c/enterprise-gateway) or [gitter room](https://gitter.im/jupyter/enterprise_gateway).  For more sensitive discussions we would be happy to have a call in order to help move things forward for you and your organization.

witten:
I think it's about two things: 1. We need to free up the internal development bandwidth to drive adoption. That's strictly on us. 2. Once we do that, we can try to integrate Enterprise Gateway with our stack and workflow. That's where I think it's likely that we'll have questions or feature requests that you may be able to help out with or at least weigh in on.

For number 2, here's an example. We use HDFS, with uses Kerberos, which requires a custom Docker image for use with Spark. If/when we get to the point of trying that out, we may request help with with using custom Spark Docker images with Enterprise Gateway.

Thanks though for your offer to help out! That's much appreciated.

kevin-bates:
Sounds good.

[...]

Issue #626, opened 2019-04-03
Kubernetes ingress


witten:
Based on the [documentation](https://jupyter-enterprise-gateway.readthedocs.io/en/latest/getting-started-kubernetes.html#enterprise-gateway-deployment), it sounds like the current approach for reaching the Jupyter Enterprise Gateway running on Kubernetes is via a NodePort. However, on most clusters, this does not actually make the port available to the outside world, but rather just to internal cluster IPs.

So this ticket is a request to add support for including an optional [Kubernetes Ingress](https://kubernetes.io/docs/concepts/services-networking/ingress/) resource when deploying Jupyter Enterprise Gateway. The idea is that if the cluster is configured for ingress, then the Ingress resource will thereby make the Gateway accessible from the outside world. This makes it much more convenient to reach the Gateway, say from a developer's local laptop. (Without ingress, a developer may need to maintain a `kubectl port-forward` proxy or similar.)

Note that making ingress optional may be easier to do once a Helm chart is available (#625).

lresende:
You can find an ingress template at our [ansible scripts](https://github.com/lresende/ansible-kubernetes-cluster/blob/master/roles/jupyter-enterprise-gateway/templates/enterprise-gateway-ingress.yaml.j2). The issue that I see with providing the yaml at eg repo is that there are different versions of ingress implementations and from playing with couple ones they all require different configurations and I don't think we can easily maintain/test on the multiple available ingress implementations, but I am open to suggestions... maybe just add this example to the documentation and users would translate/configure based on the ingress they are using ?

witten:
One way I've seen around this, besides making ingress itself optional, is to support configurable ingress annotation keys/values. Here's an example of an arbitrary Helm template that takes this approach: https://github.com/helm/charts/blob/master/incubator/istio/templates/deployment/ingress.yaml .. Note the `.Values.ingress.annotations` that are injected into the template.

That way, it's up to the user deploying (in this case, deploying the Helm chart) to make the Ingress resource work with their own ingress implementation.

But you make a fair point that maybe it's not worth trying to make this work for the variety of ingress implementations. As soon as you support Ingress resources, the next request will be for Certificate resources.. Documentation examples may be sufficient, especially if there's a base Helm chart (as per #625) that users can derive from to add their own custom resources like Ingresses and Certificates.

witten:
Some additional thoughts from @kevin-bates on the topic of ingress from #625:

[...]

PR #628, opened 2019-04-04
#625: Helm chart

witten:
##### Changes

This PR includes an optional Helm chart for deploying Enterprise Gateway, and the corresponding documentation updates to describe using it.

I started by copying the existing `etc/kubernetes/enterprise-gateway.yaml` and splitting it into separate per-resource template files as per Helm convention. Then, I factored out any apparent customization points into documented Helm values that get plumbed into the templates. That way, modifying the Helm chart should generally not be necessary on a per-deployment basis, and customization can happen directly via overrides on the Helm command-line.

In a few cases, I opted to support specification of the values as native YAML types rather than as strings. This includes `true`/`false`, various integer values, and the Kernel whitelist as a list. This made it easier to deal with these values in Helm templates. And they are converted to strings before being passed into the Enterprise Gateway pod as environment variables, so there should be no functional difference.

I also added metadata labels that are usually found on Helm chart resources.

##### Testing

```bash
$ helm upgrade --install --atomic --namespace enterprise-gateway enterprise-gateway etc/kubernetes/helm
Release "enterprise-gateway" has been upgraded. Happy Helming!
...
```

I ensured things appeared to be running:

```
$ kubectl get all --all-namespaces -l app=enterprise-gateway
NAMESPACE            NAME                        DESIRED   CURRENT   UP-TO-DATE   AVAILABLE   AGE
enterprise-gateway   deploy/enterprise-gateway   1         1         1            1           1h

NAMESPACE            NAME                               DESIRED   CURRENT   READY     AGE
enterprise-gateway   rs/enterprise-gateway-64fb7c779c   1         1         1         1h
enterprise-gateway   rs/enterprise-gateway-6db45fb858   0         0         0         1h
enterprise-gateway   rs/enterprise-gateway-6fb884bb5f   0         0         0         1h
enterprise-gateway   rs/enterprise-gateway-bc5b4676b    0         0         0         1h

NAMESPACE                                    NAME                                                                 READY     STATUS    RESTARTS   AGE
enterprise-gateway                           po/enterprise-gateway-64fb7c779c-jrxkq                               1/1       Running   0          1h

NAMESPACE            NAME                     TYPE       CLUSTER-IP      EXTERNAL-IP   PORT(S)          AGE
enterprise-gateway   svc/enterprise-gateway   NodePort   100.66.108.35   <none>        8888:30913/TCP   1h
```

I started a `kubectl port-forward` to Enterprise Gateway on port 8888, and then I started a notebook locally to connect, selecting the Python Spark Kubernetes kernel. Then I printed `sc` within the notebook to confirm that the Spark context was working.

I also tried overriding a few of the configurable values to make sure that worked. For instance, including
`--set kernel_whitelist={spark_python_kubernetes}` with the Helm upgrade ensured that only a single kernel showed up in the Jupyter UI. Setting `--set replicas=2` ensured that two Enterprise Gateway pods got spun up. (Although I had to include the `--recreate-pods` Helm option.) Including `--set mirror_working_dirs=true` ensured that "True" was passed to the Enterprise Gateway pod for that option.

Here's the full pod YAML from one of these tests:

```
$ kubectl get pods -n enterprise-gateway -o yaml
apiVersion: v1
items:
- apiVersion: v1
  kind: Pod
  metadata:
    creationTimestamp: 2019-04-04T19:32:19Z
    generateName: enterprise-gateway-64fb7c779c-
    labels:
      app: enterprise-gateway
      component: enterprise-gateway
      gateway-selector: enterprise-gateway
      pod-template-hash: "2096373357"
    name: enterprise-gateway-64fb7c779c-jrxkq
    namespace: enterprise-gateway
    ownerReferences:
    - apiVersion: apps/v1
      blockOwnerDeletion: true
      controller: true
      kind: ReplicaSet
      name: enterprise-gateway-64fb7c779c
      uid: a83ebe09-570f-11e9-838d-06c5023a10da
    resourceVersion: "56510328"
    selfLink: /api/v1/namespaces/enterprise-gateway/pods/enterprise-gateway-64fb7c779c-jrxkq
    uid: 5c4fab58-5710-11e9-838d-06c5023a10da
  spec:
    containers:
    - args:
      - --gateway
      env:
      - name: EG_NAMESPACE
        value: enterprise-gateway
      - name: EG_KERNEL_CLUSTER_ROLE
        value: kernel-controller
      - name: EG_SHARED_NAMESPACE
        value: "False"
      - name: EG_MIRROR_WORKING_DIRS
        value: "False"
      - name: EG_CULL_IDLE_TIMEOUT
        value: "3600"
      - name: EG_LOG_LEVEL
        value: DEBUG
      - name: EG_KERNEL_LAUNCH_TIMEOUT
        value: "60"
      - name: EG_KERNEL_WHITELIST
        value: '["spark_python_kubernetes"]'
      image: elyra/enterprise-gateway:dev
      imagePullPolicy: IfNotPresent
      name: enterprise-gateway
      ports:
      - containerPort: 8888
        protocol: TCP
      resources: {}
      terminationMessagePath: /dev/termination-log
      terminationMessagePolicy: File
      volumeMounts:
      - mountPath: /var/run/secrets/kubernetes.io/serviceaccount
        name: enterprise-gateway-sa-token-4c484
        readOnly: true
    dnsPolicy: ClusterFirst
    nodeName: REDACTED.internal
    priority: 0
    restartPolicy: Always
    schedulerName: default-scheduler
    securityContext: {}
    serviceAccount: enterprise-gateway-sa
    serviceAccountName: enterprise-gateway-sa
    terminationGracePeriodSeconds: 30
    tolerations:
    - effect: NoExecute
      key: node.kubernetes.io/not-ready
      operator: Exists
      tolerationSeconds: 300
    - effect: NoExecute
      key: node.kubernetes.io/unreachable
      operator: Exists
      tolerationSeconds: 300
    volumes:
    - name: enterprise-gateway-sa-token-4c484
      secret:
        defaultMode: 420
        secretName: enterprise-gateway-sa-token-4c484
  status:
    conditions:
    - lastProbeTime: null
      lastTransitionTime: 2019-04-04T19:32:19Z
      status: "True"
      type: Initialized
    - lastProbeTime: null
      lastTransitionTime: 2019-04-04T19:32:20Z
      status: "True"
      type: Ready
    - lastProbeTime: null
      lastTransitionTime: null
      status: "True"
      type: ContainersReady
    - lastProbeTime: null
      lastTransitionTime: 2019-04-04T19:32:19Z
      status: "True"
      type: PodScheduled
    containerStatuses:
    - containerID: docker://6c7f58c9dd77ff4063537eef1116aa10990d709f6a7e357d65433195412a6f1b
      image: elyra/enterprise-gateway:dev
      imageID: docker-pullable://elyra/enterprise-gateway@sha256:5621319b2813b614b2f4f2b2ce2db238aff63c8bceae6bd06988227c9348ca14
      lastState: {}
      name: enterprise-gateway
      ready: true
      restartCount: 0
      state:
        running:
          startedAt: 2019-04-04T19:32:20Z
    hostIP: 172.19.34.186
    phase: Running
    podIP: 100.96.3.249
    qosClass: BestEffort
    startTime: 2019-04-04T19:32:19Z
kind: List
metadata:
  resourceVersion: ""
  selfLink: ""
```

### DIFF
--- a/etc/kubernetes/helm/Chart.yaml
+++ b/etc/kubernetes/helm/Chart.yaml
@@ -1,0 +1,3 @@
+name: enterprise-gateway
+version: 1.0.0
+
--- a/etc/kubernetes/helm/Chart.yaml
+++ b/etc/kubernetes/helm/Chart.yaml
@@ -1,3 +1,3 @@
 name: enterprise-gateway
-version: 1.0.0
+version: 0.8.0
 
--- a/etc/kubernetes/helm/templates/deployment.yaml
+++ b/etc/kubernetes/helm/templates/deployment.yaml
@@ -25,36 +25,36 @@ spec:
       # Created by this chart.
       serviceAccountName: enterprise-gateway-sa
       containers:
-        - name: enterprise-gateway
-          image: {{ .Values.image }}
-          imagePullPolicy: {{ .Values.imagePullPolicy }}
-          args: ["--gateway"]
-          env:
-          - name: EG_NAMESPACE
-            value: {{ .Release.Namespace }}
-          - name: EG_KERNEL_CLUSTER_ROLE
-            value: {{ .Values.kernel_cluster_role }}
-          - name: EG_SHARED_NAMESPACE
-            value: {{ if .Values.shared_namespace }}"True"{{ else }}"False"{{ end }}
-          - name: EG_MIRROR_WORKING_DIRS
-            value: {{ if .Values.mirror_working_dirs }}"True"{{ else }}"False"{{ end }}
-          - name: EG_CULL_IDLE_TIMEOUT
-            value: !!str {{ .Values.cull_idle_timeout }}
-          - name: EG_LOG_LEVEL
-            value: {{ .Values.log_level }}
-          - name: EG_KERNEL_LAUNCH_TIMEOUT
-            value: !!str {{ .Values.kernel_launch_timeout }}
-          - name: EG_KERNEL_WHITELIST
-            value: {{ toJson .Values.kernel_whitelist | squote }}
-          ports:
-          - containerPort: 8888
-{{ if .Values.nfs.enabled }}
-          volumeMounts:
-          - name: kernelspecs
-            mountPath: "/usr/local/share/jupyter/kernels"
-        volumes:
+      - name: enterprise-gateway
+        image: {{ .Values.image }}
+        imagePullPolicy: {{ .Values.imagePullPolicy }}
+        args: ["--gateway"]
+        env:
+        - name: EG_NAMESPACE
+          value: {{ .Release.Namespace }}
+        - name: EG_KERNEL_CLUSTER_ROLE
+          value: {{ .Values.kernel_cluster_role }}
+        - name: EG_SHARED_NAMESPACE
+          value: {{ if .Values.shared_namespace }}"True"{{ else }}"False"{{ end }}
+        - name: EG_MIRROR_WORKING_DIRS
+          value: {{ if .Values.mirror_working_dirs }}"True"{{ else }}"False"{{ end }}
+        - name: EG_CULL_IDLE_TIMEOUT
+          value: !!str {{ .Values.cull_idle_timeout }}
+        - name: EG_LOG_LEVEL
+          value: {{ .Values.log_level }}
+        - name: EG_KERNEL_LAUNCH_TIMEOUT
+          value: !!str {{ .Values.kernel_launch_timeout }}
+        - name: EG_KERNEL_WHITELIST
+          value: {{ toJson .Values.kernel_whitelist | squote }}
+        ports:
+        - containerPort: 8888
+{{- if .Values.nfs.enabled }}
+        volumeMounts:
         - name: kernelspecs
-          nfs:
-            server: {{ .Values.nfs.internal_server_ip_address }}
-            path: "/usr/local/share/jupyter/kernels"
-{{ end }}
+          mountPath: "/usr/local/share/jupyter/kernels"
+      volumes:
+      - name: kernelspecs
+        nfs:
+          server: {{ .Values.nfs.internal_server_ip_address }}
+          path: "/usr/local/share/jupyter/kernels"
+{{- end }}
--- a/etc/kubernetes/helm/templates/deployment.yaml
+++ b/etc/kubernetes/helm/templates/deployment.yaml
@@ -55,6 +55,6 @@ spec:
         volumes:
         - name: kernelspecs
           nfs:
-            server: {{ .Values.internal_server_ip_address }}
+            server: {{ .Values.nfs.internal_server_ip_address }}
             path: "/usr/local/share/jupyter/kernels"
 {{ end }}
--- a/etc/kubernetes/helm/templates/deployment.yaml
+++ b/etc/kubernetes/helm/templates/deployment.yaml
@@ -1,0 +1,60 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: enterprise-gateway
+  namespace: enterprise-gateway
+  labels:
+    gateway-selector: enterprise-gateway
+    app: enterprise-gateway
+    component: enterprise-gateway
+    chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+spec:
+  replicas: {{ .Values.replicas }}
+  selector:
+    matchLabels:
+      gateway-selector: enterprise-gateway
+  template:
+    metadata:
+      labels:
+        gateway-selector: enterprise-gateway
+        app: enterprise-gateway
+        component: enterprise-gateway
+    spec:
+      # Created by this chart.
+      serviceAccountName: enterprise-gateway-sa
+      containers:
+        - name: enterprise-gateway
+          image: {{ .Values.image }}
+          imagePullPolicy: {{ .Values.imagePullPolicy }}
+          args: ["--gateway"]
+          env:
+          - name: EG_NAMESPACE
+            value: {{ .Release.Namespace }}
+          - name: EG_KERNEL_CLUSTER_ROLE
+            value: {{ .Values.kernel_cluster_role }}
+          - name: EG_SHARED_NAMESPACE
+            value: {{ if .Values.shared_namespace }}"True"{{ else }}"False"{{ end }}
+          - name: EG_MIRROR_WORKING_DIRS
+            value: {{ if .Values.mirror_working_dirs }}"True"{{ else }}"False"{{ end }}
+          - name: EG_CULL_IDLE_TIMEOUT
+            value: !!str {{ .Values.cull_idle_timeout }}
+          - name: EG_LOG_LEVEL
+            value: {{ .Values.log_level }}
+          - name: EG_KERNEL_LAUNCH_TIMEOUT
+            value: !!str {{ .Values.kernel_launch_timeout }}
+          - name: EG_KERNEL_WHITELIST
+            value: {{ toJson .Values.kernel_whitelist | squote }}
+          ports:
+          - containerPort: 8888
+{{ if .Values.nfs.enabled }}
+          volumeMounts:
+          - name: kernelspecs
+            mountPath: "/usr/local/share/jupyter/kernels"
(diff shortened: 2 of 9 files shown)


kevin-bates:
@witten - this is fantastic!  I'll take this for a spin first, but on the surface, this looks really good and thorough - thank you!

etc/kubernetes/helm/Chart.yaml
```diff
@@ -0,0 +1,3 @@
+name: enterprise-gateway
+version: 1.0.0
```
kevin-bates: I'm thinking we would want the Chart version at 1.0 for the GA release, but perhaps have it prior to that until we GA.  Something like `0.8.0` or `0.9.0` perhaps?

etc/kubernetes/helm/values.yaml
```diff
@@ -0,0 +1,26 @@
+image: elyra/enterprise-gateway:dev
```
kevin-bates: The tag should be EG-version specific.  We have a [release script](https://github.com/jupyter/enterprise_gateway/blob/master/release.sh) that replaces `VERSION` with the appropriate value.  I'm also thinking that things like these charts and the k8s eg yaml file should be included in the distribution output somehow.

etc/kubernetes/helm/Chart.yaml
lresende: Container/Kubernetes support is only available with EG 2.x release, so I would recommend aligning the first release of helm with that to avoid confusion. But I am flexible.

witten: @kevin-bates I've set the version number to `0.8.0` as per your suggestion.

@lresende I may be misunderstanding, but given that this PR would merge into master, the Helm support wouldn't actually make its way into general availability until the next official release (or unless someone clones master directly).

witten:
> Also, in working with this, I found the k8s-master-ip thing to be virtually useless unless you're in a single-node env (for POCs, etc.). As a result, I'm thinking it would be really helpful if we could add a template (in another PR), that produces the equivalent of what our [ansible scripts produce for a traefik ingress](https://github.com/lresende/ansible-kubernetes-cluster/blob/3c2f5386c84771a2e788f3cf6e29ac26b6e58a86/roles/jupyter-enterprise-gateway/templates/enterprise-gateway-ingress.yaml.j2). Of course, we'd want to make the ingress 'provider' a variable, along with the path (which is likely a provider-specific variable). But I see that as another great thing to add from an example standpoint for others to glean from.

I agree about the need for ingress (versus the `k8s_master_public_ip` thing), which is why I filed #626! So I just want to make sure: Is what you're suggesting here aligned with / the same as that ticket? Or are you suggesting a different approach? Thanks.

etc/kubernetes/helm/values.yaml
witten: Good catch. Just pushed a change to that release script that sets the version in the Helm chart's `values.yaml` upon release. Testing done: While I wasn't able to run the whole release script (maybe a local virtualenv issue), I did confirm it got through the version substitution part, and successfully replaced the version in the Helm `values.yaml` as expected.

As for distributing the K8s YAML, I'm open to ideas on where to add them! It looks like the Makefile produces several build/release artifacts: Python wheels, Python source dists, Docker images, etc. I'm not sure it makes sense to put the K8s YAML into any of those, so maybe it should be a separate build artifact on its own? Or see below about the source tarball. Honestly, the fact that distributing K8s YAML is kind of a pain is part of the problem that Helm attempts to solve.

As for the Helm chart, the canonical way for distributing charts is to stand up a chart server that the Helm client can pull from. (Analogous to a container image server, but for Helm charts.) I think it can be as simple as a web server with a little bit of metadata in a standard location, or as full-featured as an [actual server](https://chartmuseum.com/). I thought about including that originally, but didn't want to gate this change on standing up something like that for this project.

Maybe the right place for a v0 is to include both the K8s YAML and the Helm chart in the source release tarball? If so, then the good news is the K8s YAML is already in there, and I imagine the Helm chart would be as well upon the next release. If you like, I can update the documentation to instruct users to use the release tarball for both.

kevin-bates: Yes - I had forgotten about the source tarball.  That's fine for now.

etc/kubernetes/helm/templates/deployment.yaml
```diff
@@ -0,0 +1,60 @@
...
+{{ if .Values.nfs.enabled }}
+          volumeMounts:
+          - name: kernelspecs
+            mountPath: "/usr/local/share/jupyter/kernels"
+        volumes:
+        - name: kernelspecs
+          nfs:
+            server: {{ .Values.internal_server_ip_address }}
+            path: "/usr/local/share/jupyter/kernels"
+{{ end }}
```
kevin-bates: I'm trying to enable the nfs portion and getting the following ...
```
helm upgrade enterprise-gateway etc/kubernetes/helm
UPGRADE FAILED
ROLLING BACK
Error: YAML parse error on enterprise-gateway/templates/deployment.yaml: error converting YAML to JSON: yaml: line 54: did not find expected '-' indicator
```
I can't figure out what's going on since this matches the previous yaml.  Any ideas?

kevin-bates:
Regarding the ingress, #626 completely slipped my mind!  Sorry about that.

etc/kubernetes/helm/templates/deployment.yaml
```diff
@@ -0,0 +1,60 @@
...
+          ports:
+          - containerPort: 8888
+{{ if .Values.nfs.enabled }}
+          volumeMounts:
+          - name: kernelspecs
+            mountPath: "/usr/local/share/jupyter/kernels"
+        volumes:
+        - name: kernelspecs
+          nfs:
+            server: {{ .Values.internal_server_ip_address }}
```
kevin-bates: Does this need to be `{{ .Values.nfs.internal_server_ip_address }}`?

etc/kubernetes/helm/values.yaml
witten: Cool. I've just added a mention in the docs that a source release can be used for either a K8s YAML or Helm install.

etc/kubernetes/helm/templates/deployment.yaml
witten: Yes, good catch! Fixed.

etc/kubernetes/helm/templates/deployment.yaml
kevin-bates: Just discovered `helm lint` and here's what it produces...
```
helm lint etc/kubernetes/helm
==> Linting etc/kubernetes/helm
[ERROR] Chart.yaml: directory name (helm) and chart name (enterprise-gateway) must be the same
[INFO] Chart.yaml: icon is recommended
[ERROR] templates/deployment.yaml: unable to parse YAML
	error converting YAML to JSON: yaml: line 54: did not find expected '-' indicator
```
I then renamed the helm directory to enrterprise-gateway (per the error) and it went away.  This parsing error only shows when nfs.enabled is true.

witten: It might be the missing `nfs.` prefix in `internal_server_ip_address` in the template that you just discovered.. A missing value may have tripped up the parser. Please give it another shot with the fix and let me know if it works! (I don't have an NFS server here, although if needed I can probably repro the issue even without one.)

witten:
> Regarding the ingress, #626 completely slipped my mind! Sorry about that.

No worries! I'll add your thoughts to that ticket.

etc/kubernetes/helm/templates/deployment.yaml
kevin-bates: I had already been running that way.  Since this can reproduce with `helm lint` I don't think NFS is required.  I'm happy to check it once we get past this parse issue.  I've even tried removing the conditional and still get the same thing.

witten: Strangely, I don't get a repro with `helm lint`, but I do get one with a deploy using `--set nfs.enabled=true`.. Looking into it.

kevin-bates: Found the issue.  The `volumes` block needs to be two spaces to the left.  So the indentation was throwing things off.

witten: Yes, just found that as well! Fix tested and pushed.

PR #633, opened 2019-04-08
Added docker-compose file

mattjtodd:
Provides a basic `docker-compose` file definition.

https://github.com/jupyter/enterprise_gateway/issues/632


etc/docker/docker-compose.yml
```diff
@@ -0,0 +1,32 @@
+version: "3.3"
+
+services:
+
+  enterprise-gateway:
+    image: elyra/enterprise-gateway:${VERSION:-2.0.0rc1}
```
kevin-bates: The "convention" we're adopting is to use a `:dev` tag in the various container yaml files, then add entries to `release.sh` that [replace dev tags with release-specific values](https://github.com/jupyter/enterprise_gateway/blob/master/release.sh#L192) at the time a release is produced.

I'm noticing this same issue exists in the docker-based shell script files, so if you want to address those at the same time, that would be great!  This might be moot should we decide that the compose scripts completely replace the shell scripts.

etc/docker/docker-compose.yml
```diff
@@ -0,0 +1,32 @@
...
+    user: root
+    volumes:
+      - "/var/run/docker.sock:/var/run/docker.sock"
```
kevin-bates: Would it be possible to add the `MOUNT_KERNELSPECS` entry but commented out?

etc/docker/docker-compose.yml
mattjtodd: That's great I'll fix up the branch to point to `dev` by default.

The shell scripts _should_ be able to be deprecated as the single compose file as it will work in both compose and swarm modes.

etc/docker/docker-compose.yml
```diff
@@ -3,7 +3,7 @@ version: "3.3"
 services:
 
   enterprise-gateway:
-    image: elyra/enterprise-gateway:${VERSION:-2.0.0rc1}
+    image: elyra/enterprise-gateway:${VERSION:-dev}
```
kevin-bates: Actually, this should be `elyra/enterprise-gateway:dev` and then release.sh should be updated to replace `:dev` with the release-specific tag.

mattjtodd:
I think it would be best if there is just a single `docker-compose` file which can be deployed using either compose or as a stack in swarm.  I can't see anything in the `.sh` scripts which would exclude this as an option, so they could be removed.

I'll move the comments into the compose file in the relevant positions.

For the getting started docs, I'll post a description of how to create the resources using both compose and swarm with links to the appropriate Docker documentation.

etc/docker/docker-compose.yml
mattjtodd: Yes, I've added this and the documentation.

release.sh
```diff
@@ -185,12 +185,15 @@ function update_version_to_release {
...
-    # Update Kubernetes deployment descriprot
+    # Update Kubernetes deployment descriptor
     sed -i .bak "s@elyra/enterprise-gateway:dev@elyra/enterprise-gateway:$RELEASE_VERSION@g" etc/kubernetes/enterprise-gateway.yaml
 
     # Update Kubernetes Helm values
     sed -i .bak "s@elyra/enterprise-gateway:dev@elyra/enterprise-gateway:$RELEASE_VERSION@g" etc/kubernetes/helm/values.yaml
 
+    # Update Docker compose version
+    sed -i .bak "s@elyra/enterprise-gateway:dev@elyra/enterprise-gateway:$RELEASE_VERSION@g" etc/docker/docker-compose.yml
+
```
kevin-bates: Please apply the reverse action to `update_version_to_development` below.  Then I think we're good to go.  Thanks.

PR #635, opened 2019-04-09
[K8S HELM DEPENDENCY][WIP] Mount kernelspecs via Helm without NFS

witten:
### Overview

This PR provides a way to mount kernelspecs into an Enterprise Gateway Kubernetes pod directly at [Helm](https://helm.sh/) deployment time if using Helm. This is an alternative to mounting kernelspecs into an Enterprise Gateway K8s pod via an NFS server. The motivation is twofold:

* Setting up, maintaining, and deploying kernelspecs to an NFS server is kind of a pain, especially within Kubernetes.
* It'd be super nice for usability if a Helm "release" (Helm's terminology for a deployment) included everything necessary to get Enterprise Gateway up and running.. including kernelspecs. No multiple-step deploys. This works today with the stock kernelspecs (just deploy EG and go), but it'd also be nice if this worked for custom kernelspecs.

### Usage

So the general approach is as follows. The user puts their kernelspecs in a local Helm values file like `kernelspecs.yaml` as follows:

```yaml
kernelspecs:
  spark_python_kubernetes:
    kernel.json: |
      {
        "language": "python",
        "display_name": "Spark - Python (Kubernetes Mode)",
        "metadata": {
          "process_proxy": {
            ...
          }
        }
      }
    run.sh: |
      #!/usr/bin/env bash

      if [ "${EG_IMPERSONATION_ENABLED}" = "True" ]; then
      ...
```

In other words, the files are pasted directly into the YAML. Then, to deploy:

```bash
$ helm upgrade --install --namespace enterprise-gateway enterprise-gateway --values kernelspecs.yaml etc/kubernetes/helm
```

Note the `--values kernelspecs.yaml` part. This passes the kernelspecs into the Helm chart, resulting in the given kernelspecs showing up in the Enterprise Gateway pod, very similar to how they'd show up if mounted from NFS.

### Changes

* Add a K8s ConfigMap containing the kernelspecs provided from the command-line values.
* Consume that ConfigMap within the K8s Deployment, mounting the expected `kernel.json` and `run.sh` (if given) as a volume in the Enterprise Gateway pod.

### Testing

See usage above. I tried:

 * Deploy the given `kernelspecs.yaml`. The files showed up as expected within the EG pod in the kernelspecs directory.
 * Deploy `kernelspecs.yaml` with the `run.sh` entry removed. Just the kernelspec showed up in the pod without any `run.sh`.
 * Deploy without any `--values kernelspecs.yaml` at all. The stock kernespecs showed up in the pod.

### Still to do

* Docs updates, including examples.
* More testing. Specifically, consuming one of these mounted kernels in an actual Jupyter notebook.

### Looking for

Thoughts/reactions? Would you entertain this sort of feature?

### DIFF
--- a/etc/kubernetes/helm/templates/configmap.yaml
+++ b/etc/kubernetes/helm/templates/configmap.yaml
@@ -1,0 +1,21 @@
+{{- if (len .Values.kernelspecs) }}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: enterprise-gateway-kernelspecs
+  labels:
+    name: enterprise-gateway-kernelspecs
+    app: enterprise-gateway
+    chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+data:
+  {{- range $kernel_name, $kernelspec := .Values.kernelspecs }}
+  {{ $kernel_name }}.kernel.json: |
+{{ index $kernelspec "kernel.json" | indent 4 }}
+  {{ $kernel_name }}.run.sh: |
+  {{- if hasKey $kernelspec "run.sh" }}
+{{ index $kernelspec "run.sh" | indent 4 }}
+  {{- end }}
+  {{- end }}
+{{- end }}
--- a/etc/kubernetes/helm/templates/deployment.yaml
+++ b/etc/kubernetes/helm/templates/deployment.yaml
@@ -57,4 +57,22 @@ spec:
         nfs:
           server: {{ .Values.nfs.internal_server_ip_address }}
           path: "/usr/local/share/jupyter/kernels"
+{{- else if (len .Values.kernelspecs) }}
+        volumeMounts:
+        - name: kernelspecs
+          mountPath: "/usr/local/share/jupyter/kernels"
+      volumes:
+      - name: kernelspecs
+        configMap:
+          name: enterprise-gateway-kernelspecs
+          {{- range $kernel_name, $kernelspec := .Values.kernelspecs }}
+          items:
+          - key: {{ $kernel_name }}.kernel.json
+            path: {{ $kernel_name }}/kernel.json
+          {{- if hasKey $kernelspec "run.sh" }}
+          - key: {{ $kernel_name }}.run.sh
+            path: {{ $kernel_name }}/bin/run.sh
+            mode: 0755
+          {{- end }}
+          {{- end }}
 {{- end }}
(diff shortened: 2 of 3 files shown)


witten:
Responding by numbered point:

1. Doh, I should've seen those other files included in the kernelspecs. I was just looking at a couple of kernels and assuming they were representative.
2. Okay.
3. Yes, the way that the ConfigMap is currently mounted makes the stock kernelspecs unavailable. Although if that's undesirable, it's possible to mount individual sub-directories instead.
4. Yes, YAML file inclusion is possible, although kind of a PITA with Helm. I have done it before for at least one chart.
5. See below about alternate PV ideas.
6. Yup, agreed that in theory this should "just work" assuming the appropriate files are present in the appropriate location!
7. With the approach in this PR, the mounted files within the pod _do_ update if the ConfigMap updates. So you can edit the ConfigMap either through command-line, Kubernetes Dashboard, or a Helm re-deploy, and within a few seconds it'll show up in the pod even if the pod doesn't restart. Not as convenient as hacking on files in your editor of choice on NFS, but perhaps not too bad.

However, I'm beginning to think that I probably shouldn't pursue this ConfigMap-based approach. Here's why: ConfigMaps are designed for deploying configuration, specifically. When I thought it was just a JSON file and perhaps a small shell script, I could convince myself that these kernelspecs are mostly just config. However, once you mentioned that Python files are also part of kernelspecs, that makes me think that trying to shove those into ConfigMaps too would be abusing the intent of ConfigMaps. Plus, there's just the practical challenge of trying to represent an arbitrary hierarchy of files within ConfigMaps, which are fairly non-hierarchical. (They're just keys/values.)

So I'm thinking that I should close this PR. But since I do still think it's valuable to provide non-NFS options, I have a couple of other ideas for how to get custom kernelspecs into Enterprise Gateway on Kubernetes. Some different options:

1. The user can build a custom Docker image that just contains kernelspecs and that's it. Then, the Enterprise Gateway Helm chart can provide a supported way to mount that Docker image's data directly into the Enterprise Gateway container at the expected kernelspec path location. So when you deploy Enterprise Gateway, both Docker images would deploy as part of the same Kubernetes pod. The thinking here is that a custom Docker image is a more appropriate (and practical) way to capture a custom filesystem hierarchy than a ConfigMap is.
2. Or, the user can build a custom Docker image derived from the stock Enterprise Gateway Docker image. In this derived image, replace some or all of the stock kernelspecs with custom kernelspecs. Sort of a fall-back option, and certainly less convenient.

Thoughts?

kevin-bates:
Thanks for the update.  I really appreciate you honoring the intention of configMaps and not try to force the issue - thanks!  This was a very useful exercise nonetheless.

Seems like a kernelspecs "sidecar" is a great way to address customizations.  Folks adopting that approach could either amend or replace the OOTB kernelspecs depending on how they mount and setup `JUPYTER_PATH`.

witten:
Okay, I'll give that approach a shot and see how it works out! Thanks for your feedback.

PR #637, opened 2019-04-09
Use a custom image as an init container to copy in kernelspecs.

witten:
### Overview

This PR provides a way to mount kernelspecs into an Enterprise Gateway Kubernetes pod from a custom kernelspecs Docker image. This is an alternative to mounting kernelspecs into an Enterprise Gateway K8s pod via an NFS server. The motivation is twofold:

* Setting up, maintaining, and deploying kernelspecs to an NFS server is kind of a pain, especially within Kubernetes.
* It'd be super nice for usability if a Helm "release" (Helm's terminology for a deployment) included everything necessary to get Enterprise Gateway up and running.. including kernelspecs. No multiple-step deploys. This works today with the stock kernelspecs (just deploy EG and go), but it'd also be nice if this worked for custom kernelspecs.

See https://github.com/jupyter/enterprise_gateway/pull/635 for more background.

### Usage

First, I built and published a custom Docker image with this Dockerfile:

```dockerfile
FROM alpine:3.9

COPY kernels /kernels
```

The `kernels/` sub-directory contained my custom kernel:

```bash
$ ls -R kernels/
kernels/:
spark_python_kubernetes/

kernels/spark_python_kubernetes:
bin/  kernel.json

kernels/spark_python_kubernetes/bin:
```

Then, with that custom image in place, I deployed Enterprise Gateway with Helm:

```bash
$ helm upgrade --atomic --install --namespace enterprise-gateway enterprise-gateway etc/kubernetes/helm/ --set kernelspecs_image=my-custom-jupyter-kernelspecs:example
```

Note that the `kernelspecs_image` value refers to the name of the kernelspecs image built above. (Image name is fake; the actual one is on a private registry.)

This resulted in the Enterprise Gateway pod coming up with its kernelspecs path containing the custom kernelspecs!

### Changes

1. Rather than a sidecar container, I went with a K8s Init Container. That seemed more appropriate, because you can't directly mount a container into another container in the same pod; you have to share data over a volume of some sort. In this case, I opted for an in-memory `emptyDir`. And the kernelspecs Init Container simply spins up, copies its data into the shared volume, and spins down. This copying is done in the Helm chart, so the end-user doesn't have to worry about it. All they need to do is abide by the convention of putting their kernelspecs into `/kernels` in their container.
2. Add appropriate values to the chart's `values.yaml`.
3. Documentation.

### Tesing

Did the deploy above. Observed that Enterprise Gateway spun up, and the kernelspecs appeared in the expected location in the running pod. Connected to Enterprise Gateway from a local Jupyter notebook. Confirmed that I could connect to the custom kernel and get notebook output.

Confirmed that I could still deploy NFS with the Helm chart without this new custom container image. Did not confirm anything about NFS actually functioning (no NFS server here).

Tried deploying both NFS and a custom container image together (which is not supported). Observed that Helm failed spectacularly on a template error, as expected. At some point it might be nicer to make it fail in a more friendly fashion.

### DIFF
--- a/docs/source/getting-started-kubernetes.md
+++ b/docs/source/getting-started-kubernetes.md
@@ -212,10 +212,15 @@ roleRef:
   apiGroup: rbac.authorization.k8s.io
 ```
 
-##### Kernelspec Modifications
-One of the more common areas of customization we see occur within the kernelspec files located in `/usr/local/share/jupyter/kernels`.  To accommodate the ability to customize the kernel definitions, the kernels directory can be mounted as an NFS volume into the Enterprise Gateway pod thereby making the kernelspecs available to all EG pods within the Kubernetes cluster (provided the NFS mounts exist on all applicable nodes).
+#### Kernelspec Modifications
 
-As an example, we have included the necessary entries for mounting an existing NFS mount point into the Enterprise Gateway pod. By default, these references are commented out as they require the system administrator configure the appropriate NFS mounts and server IP.
+One of the more common areas of customization we see occurs within the kernelspec files located in `/usr/local/share/jupyter/kernels`.  To accommodate the ability to customize the kernel definitions, you have two different options: NFS mounts, or custom container images.  The two options are mutually exclusive, because they mount kernelspecs into the same location in the Enterprise Gateway pod.
+
+##### Via NFS
+
+The kernels directory can be mounted as an NFS volume into the Enterprise Gateway pod, thereby making the kernelspecs available to all EG pods within the Kubernetes cluster (provided the NFS mounts exist on all applicable nodes).
+
+As an example, we have included the necessary entries for mounting an existing NFS mount point into the Enterprise Gateway pod.  By default, these references are commented out as they require the system administrator configure the appropriate NFS mounts and server IP.  If you are deploying Enterprise Gateway via the Helm chart (see Deploying Enterprise Gateway, below), you can enable NFS directly via Helm values.
 
 Here you can see how `enterprise-gateway.yaml` references use of the volume (via `volumeMounts`
 for the container specification and `volumes` in the pod specification):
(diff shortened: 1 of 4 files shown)


kevin-bates:
Very nice - what you said makes sense.  

1. It sounds like the contents of the init container are copied into the EG container.  This implies that users can both amend and replace the OOTB kernelspecs w/o having to adjust JUPYTER_PATH.  Is that correct?
2. Of course, to amend, one would also need adjust EG's whitelist variable to include the amended specs.
3. I would suggest testing both amended (entirely new directory and display name (in kernel.json)) and replaced (with updated display name in kernel.json) to ensure things work as expected. 
4. Ensure that run.sh is still executable after transfer.

Thank you!

witten:
1. Unfortunately, given the current approach, no.. This will replace the entirety of the kernelspecs container in the EG pod. That's because a whole volume is mounted in, rather than individual kernelspec directories. So it's either this, or the OOTB kernelspecs, but not both. Do you think this is a deal-breaker?
2. Yes. In any case, I can mention the whitelist variable in the docs.
3. See point 1 above. (There's no amending going on.)
4. Good call. It is! File permissions are maintained. As part of testing, I would actually try consuming this in an actual Jupyter notebook to make sure it runs.

kevin-bates:
No worries - I think we're still good to go.  

Amended functionality can be achieved by tweaking the mount point and adjusting JUPYTER_PATH.  It can also be achieved by loading our [tarball of k8s kernelspecs](https://github.com/jupyter/enterprise_gateway/releases/tag/v2.0.0-beta.1) into the init container - which is equivalent to the builtins, then extending the set of directories. 

Either case warrants adjustment to the whitelist - which is fine as well.

witten:
Got it, thanks. Still leaving it WIP though until I do some additional testing.

witten:
WIP off. Ready to merge!

Issue #638, opened 2019-04-12
Automatically pre-pull kernel images


witten:
Based on the comments below, this ticket has morphed into a feature to pre-pull kernel images to all K8s nodes, so as to sidestep Jupyter's kernel start timeout. This includes pre-pulling:

 * New kernel images, e.g. on a fresh deploy of Enterprise Gateway to a cluster or the addition of a new kernelspec to an existing deployment.
 * Updated kernel images, e.g. a given image name/tag doesn't change, but its underlying image has changed due to a new release.
 * New K8s nodes being added, e.g. due to auto-scaling.

Ideally this feature would dynamically source the set of images to pre-pull from the configured kernelspecs.

#### Original Description

I'm trying to make a custom Spark kernel as per [this comment](https://github.com/jupyter/enterprise_gateway/issues/625#issuecomment-479698455). I've set up a custom kernelspec as per [these docs](https://jupyter-enterprise-gateway.readthedocs.io/en/latest/docker.html#deploying-your-custom-kernel-image) and mounted into the Enterprise Gateway pod [via a custom container image](https://jupyter-enterprise-gateway.readthedocs.io/en/latest/getting-started-kubernetes.html#via-custom-container-image). That all appears to work fine.

However, when the kernel launches a driver pod, it does so with an `imagePullPolicy` of `IfNotPresent` (the default), apparently because one is [not otherwise configured](https://github.com/jupyter/enterprise_gateway/blob/master/etc/kernel-launchers/kubernetes/scripts/kernel-pod.yaml.j2#L55). I'd like to be able to set a different `imagePullPolicy`, e.g. `Always`, because now I'm in a position where I'd like to rev my driver image, but Enterprise Gateway refuses to pull updates and I have no way to override that behavior!

So this is a feature request to provide *some* mechanism for setting the `imagePullPolicy` for the kernel launcher, or by some other mechanism to allow my driver image to receive updates in the cluster.

Relevant section of kernelspec:

```json
  "metadata": {
    "process_proxy": {
      "class_name": "enterprise_gateway.services.processproxies.k8s.KubernetesProcessProxy",
      "config": {
        "image_name": ".../jupyter-python-spark-24-job:initial-import",
        "executor_image_name": ".../jupyter-python-spark-24-job:initial-import"
      }
    }
  },
```

Needless to say, the image tag is not changing with each revision to the image. (If it did, this wouldn't be an issue.)

And here's evidence of the resulting `imagePullPolicy`:

```bash
[...]

PR #641, opened 2019-04-17
Add daemon to pull kernel images to each node

kevin-bates:
This adds support for a _Kernel Image Puller_ image, meant to be run in a DaemonSet, that will periodically gather the set of configured kernel image names from Enterprise Gateway and pull the image that image to the node on which it is running.  Since it's configured as a DaemonSet, it will be running on each node (as well as started on newly added nodes) of the cluster.  This idea was discussed in #638.

The configurable items are all established via environment variables since that is the natural way to use parameters in container configurations.  These include:
-  **`KIP_INTERVAL`:** the number of seconds at which it waits before performing the operation again.  Default is every 5 minutes (`300`)
- **`KIP_NUM_PULLERS`:** the number of threads active at any given time operating from a shared queue.  The default number of puller threads is 2 but may vary.
- **`KIP_NUM_RETRIES`:** the number retries to perform if issues are encountered.  The exception to this are image not found exceptions. Default is 3.
- **`KIP_GATEWAY_HOST`:** the host (and path) that precedes the kernelspecs endpoint.

### DIFF
--- a/etc/docker/kernel-image-puller/kernel_image_puller.py
+++ b/etc/docker/kernel-image-puller/kernel_image_puller.py
@@ -1,0 +1,168 @@
+import os
+import logging
+import time
+import queue
+import requests
+from threading import Thread
+from docker.client import DockerClient
+from docker.errors import NotFound, APIError
+
+gateway_host = os.getenv("KIP_GATEWAY_HOST", "http://localhost:8888")
+num_pullers = int(os.getenv("KIP_NUM_PULLERS", "2"))
+num_retries = int(os.getenv("KIP_NUM_RETRIES", "3"))
+interval = int(os.getenv("KIP_INTERVAL", "300"))
+log_level = os.getenv("KIP_LOG_LEVEL", "INFO")
+
+POLICY_IF_NOT_PRESENT = "IfNotPresent"
+POLICY_ALYWAYS = "Always"
+policies = (POLICY_IF_NOT_PRESENT, POLICY_ALYWAYS)
+
+policy = os.getenv("KIP_PULL_POLICY", POLICY_IF_NOT_PRESENT)
+
+docker_client = DockerClient.from_env()
+
+logging.basicConfig(format='[%(levelname)1.1s %(asctime)s %(name)s.%(threadName)s] %(message)s')
+
+
+def get_kernelspecs():
+    """Fetches the set of kernelspecs from the gateway, returning a dict of configured kernel specs"""
+    end_point = '{}/api/kernelspecs'.format(gateway_host)
+    logger.info("Fetching kernelspecs from '{}' ...".format(end_point))
+    resp = requests.get(end_point)
+    if not resp.ok:
+        raise requests.exceptions.HTTPError('Gateway server response: {}'.format(resp.status_code))
+    return resp.json()
+
+
+def fetch_image_names():
+    """
+    Fetches the image names by hitting the /api/kernelspecs endpoing of the Gateway.
+
+    For process-proxy kernelspecs, the image names are contained in the config stanza - which
+    resides in the process-proxy stanza located in the metadata.
+    """
+
+    kspecs = None
+    try:
+        kspecs_response = get_kernelspecs()
+        kspecs = kspecs_response.get('kernelspecs')
+    except Exception as ex:
+        logger.error("Got exception attempting to retrieve kernelspecs - retrying. Exception was: {}".format(ex))
+    finally:
+        if kspecs is None:
+            return False
+
+    # Locate the configured images within the kernelspecs and add to set for duplicate management
+    images = set()
+    for key in kspecs.keys():
+        metadata = kspecs.get(key).get('spec').get('metadata')
+        if metadata is not None:
+            process_proxy = metadata.get('process_proxy')
+            if process_proxy is not None:
+                config = process_proxy.get('config')
+                if config is not None:
+                    image_name = config.get('image_name')
+                    if image_name is not None:
+                        images.add(image_name)
+                    executor_image_name = config.get('executor_image_name')
+                    if executor_image_name is not None:
+                        images.add(executor_image_name)
+
+    # Add the image names to the name queue
+    for image_name in images:
+        name_queue.put_nowait(image_name)
+
+    return True
+
+
+def pull_image(image_name):
+    """
+    Pulls the image.  If the policy is `IfNotPresent` the set of pulled image names is
+    checked and, if present, the method returns.  Otherwise, the pull attempt is made
+    and the set of pulled images is updated, when successful.
+
+    Since NotFound exceptions are tolerated, we trap for only that exception and let
+    the caller handle others.
+    """
+    if policy == POLICY_IF_NOT_PRESENT:
+        if image_name in pulled_images:
+            logger.info("Image '{}' already pulled and policy is '{}'.".format(image_name, policy))
+            return
+
+    logger.debug("Pulling image '{}'...".format(image_name))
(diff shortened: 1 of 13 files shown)


etc/docker/kernel-image-puller/kernel_image_puller.py
```diff
@@ -0,0 +1,124 @@
...
+
+logging.basicConfig(format='[%(levelname)1.1s %(asctime)s %(name)s.%(threadName)s] %(message)s')
+
+
+def get_kernelspecs():
+    """Fetches the set of kernelspecs from the gateway, returning a dict of configured kernel specs"""
+    end_point = '{}/api/kernelspecs'.format(gateway_host)
+    logger.info("Fetching kernelspecs from '{}' ...".format(end_point))
+    response = requests.get(end_point)
+    return response.json()
```
esevan: When server is unhealthy, this would raise `ValueError`. To clarify response error, I think it is better to raise clear error such as `HTTPError`
```
if not r.ok:
    raise requests.exceptions.HTTPError('Server reponse: {}'.format(r.status_code))
```

kevin-bates:
I thought of a couple other things in my sleep.

1. The behavior here is essentially an `Always` pull policy since the pull is attempted on each iteration.  Since the default pull policy is `IfNotPresent`, I think we should remember the images that have been pulled (simple set) and add a parameter `KIP_PULL_POLICY` that, if `IfNotPresent`, only pulls images it hasn't already pulled. 
2. We should still try to come up with a solution based on notifications rather than polling from each node since I'm a little concerned about scalability (100 nodes all retrieving kernelspecs every N seconds).  That said, I'm not sure how to go about detecting new images without some kind of minimal polling or radical change (like detecting new kernelspecs).  I suppose we could move the polling into EG (hooked into the tornado loop - similar to how idle watching works), that essentially does the gathering and sees what hasn't been gathered before, then triggers some kind of event for new entries.  Then the DaemonSet is merely monitoring that event, which then triggers its current behavior.  However, I think this approach would be limited to only an `IfNotPresent` policy.  Is that good enough though?  Yeah, `Always` pods will incur slow downs, but only when the actual image was changed.  OTOH, admins could adopt behaviors like "hold off using a given kernel until the cluster has had a chance to pull".
3. Truncate the pull timing values to 3 decimal places (ms).

esevan:
> I suppose we could move the polling into EG

Hmm, this is just my opinion -> I think it makes sense that "Detecting kernelspec change" in EG (event), "Detecting image change" (periodic) and "Detecting new node" (event) in KIP. I'm still not sure "image change" event can be detected unless watching it periodically.

> "hold off using a given kernel until the cluster has had a chance to pull"

Isn't this good to avoid "timeout" issue? Looks like "rolling upgrade" and "zero-downtime" upgrade.

witten:
Enterprise Gateway could publish its kernelspec list to a ConfigMap whenever it changes, and then the image puller could `kubectl get configmap --watch thatconfigmap`, thereby hopefully sidestepping the scalability concern.. Let K8s do the notifications for you! That of course only helps with the "Detecting kernelspec change" case.

kevin-bates:
Right, this would move the polling to one spot instead of N.  This would force us to a 'IfNotPresent' model - which, IMO, is sufficient since tag names should vary in the real world.

There's some heavy lifting here in that we'd need to override the KernelSpecManager, have it perform k8s ops (only when configured for k8s), then modify KIP to watch the map.  Since there's a chance this area could change depending on the jupyter server project, I'm inclined to leave things simple for now and see how it goes.   Once the server direction is set or we find scalability is a concern in this area, we can check back on this.

Does that sound ok to others?

witten:
Seems like a reasonable place to start!

kevin-bates:
Updated to include helm presence.  I decided not to parameterize the service name since that would imply the same happen for nearly all other objects.  Also, I have not parameterized the complete set of KIP parameters - only those I suspect might be adjusted more frequently.

Makefiles, documentation and release scripts have also been updated.

What is NOT updated at this time is the docker-compose file since I'm not sure how this works.  I suspect it's a matter of defining a _global_ service.  I'm hoping @mattjtodd lend his docker expertise for this.

I'm going to remove WIP.  If we find more is needed or feel like including Docker in this merge, we can do that.

Your continued reviews are appreciated!

kevin-bates:
Yes. I'll add a flag for helm tomorrow. I guess the daemonset will need to be commented out in the eg yaml file.

kevin-bates:
@esevan - I updated with the default option being `true`.  This means the eg yaml can leave the daemonset configured since that's the default.

kevin-bates:
@esevan - thank you for the review and comments.

While I definitely think there's room for improvement here, I believe we should treat those as future enhancements and go ahead and merge this PR (assuming others are okay with it).

I've been thinking about these issues quite a bit - mostly in how can KIP see a kernelspec before the users - and have thought of a couple approaches.  Both of which would require adding at a minimum a kernelspec handler and, possibly, a kernelspec manager class.

1. Hidden Kernelspecs.  The idea here would be to introduce a field in the `metadata:` stanza like `hidden: True`.  If hidden, the handler would filter the specs it should return - removing those marked as hidden.  We would then need to add a query parameter of some sort, like, `/api/kernelspecs?include_hidden=True` that KIP (and perhaps devOps) would use.  Once the _hidden_ kernelspec has been validated - both in its content and operation but also in KIP's pulling of the image - the field would be removed or set to `false` - making the kernelspec available to end users.
2. Authorized Users.  We currently support the notion of `authorized_users` (and `unauthorized_users`) lists at a per kernelspec basis, but these lists are only enforced at kernel startup using the value of KERNEL_USERNAME in the json body.  As a result, users that are not permitted to use certain kernelspecs will see those kernelspecs as options - only to fail on attempted use.  It would be nice to include this enforcement when fetching the kernelspecs so users only see what they can use.  Solving this introduces a number of things.  
    1. All clients would need to include something akin to a query parameter like `?kernel_username=alice`.  We'd tackle the Notebook/Lab cases via NB2KG.  Failure to add this parameter means that they could not see any kernelspecs with a non-zero users list.  If it didn't exist in the kernelspecs, the spec would still be returned.
    2. KIP would need to use some kind of meta-user or token that indicates its special/privileged OR its username would need to be added to all kernelspecs (which seems like a non-starter) - so preferrably something like the former.  
    3. These fields are currently buried in `metadata.process_proxy.config` and should probably get moved to `metadata` since they really aren't process-proxy specific.  (They existed before we moved process_proxy to metadata, so they needed a 'container' at the time.) They should be moved irrespective of any other decisions - IMO.

  Even if we wanted to address authorized_users for kernelspecs, we'd need to resolve how KIP could access specs regardless.

So, I guess I'm leaning toward Hidden Kernelspecs although both would be nice (I think we need a way to say a certain request (e.g., from KIP) is privileged anyhow.  ideas welcome!)

The other area I've been looking into is dynamic configuration in general.  Via a debugger I found that if I load the configuration after startup (and after making a change), I see the updated value reflected in the member (traitlet) variable.  This implies that you could have a PeriodicCallback that simply refreshes the configuration (although this is another case where 'watching' is better).  So things like remote-hosts, authorized-users, or whitelists could (in theory) be updated without having to restart EG.  

Sorry for the ramble.  I've been thinking about this stuff literally for the past couple days and its good to talk about it.  Great therapy. :smile:

esevan:
@kevin-bates So deep diving into the issue! Thanks for sharing your idea. Moving your comment to the issue will be very useful to share with others!

kevin-bates:
I added a couple related issues last week.  Are others okay to have this merged?

PR #645, opened 2019-04-22
Refactor KernelSessionManager to accommodate various persistence methods

### LINKED ISSUE #644: Refactor KernelSessionManager to allow for different forms of persistence
The current `KernelSessionManager` does a decent job of isolating it's persistence methods, but we should convert those methods into abstract methods and create a subclass `FileKernelSessionManager` to hold the file-based persistence implementations.

This will more easily enable others to bring their own forms of persistence (namely NoSQL or KV stores).

kevin-bates:
This change does some minor refactoring of `KernelSessionManager` such
that it is now a base class with two abstract methods.  A new subclass,
`FileKernelSessionManager`, has been created to provide an example subclass
which writes to a file.  This will be the default implementation.

Other forms of persistence are encouraged, namely NoSQL or KV providers.

To use a different kernel session manager add the following to the EG
command line (or configuration file equivalent):

`--EnterpriseGatewayApp.kernel_session_manager_class=<MyKernelSessionManager>`

Note there may be some additional refactoring required once we determine
the needs for other kinds of persistence.

Fixes #644

### DIFF
--- a/enterprise_gateway/enterprisegatewayapp.py
+++ b/enterprise_gateway/enterprisegatewayapp.py
@@ -20,7 +20,7 @@ from kernel_gateway.gatewayapp import KernelGatewayApp
 from kernel_gateway.services.sessions.sessionmanager import SessionManager
 
 from ._version import __version__
-from .services.sessions.kernelsessionmanager import KernelSessionManager
+from .services.sessions.kernelsessionmanager import KernelSessionManager, FileKernelSessionManager
 from .services.kernels.remotemanager import RemoteMappingKernelManager
 
 
@@ -163,7 +163,7 @@ class EnterpriseGatewayApp(KernelGatewayApp):
         default_value=KernelSpecManager,
         config=True,
         help="""
-        The kernel spec manager class to use. Should be a subclass
+        The kernel spec manager class to use. Must be a subclass
         of `jupyter_client.kernelspec.KernelSpecManager`.
         """
     )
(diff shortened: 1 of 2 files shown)


kevin-bates:
I'm going to close this PR **without merge**.  This is because I think we should instead use a persistence-class from ContentServices to save the kernel session.  As a result, there will not be a need to sub-class the KernelSessionManager - although that's still allowed, but instead the form of ContentsManager will be specified.

This approach still results in allowing different forms of persistence methods, but because that should be the only thing that really varies, its a much better use of existing implementations.  FileContentsManager will, of course, be the default, but PostgreSQL and S3 implementations already exist, among others.

I will submit a PR that uses ContentsManager shortly.

kevin-bates:
After looking into the ContentsManager and the various implementations, I think it might be best to re-open this PR for the following reasons:

1. Since session persistence only needs a small subset of `ContentsManager` functionality (file and directory operations), we shouldn't require users to implement a Contents Manager just for the few methods they need.
2. I found that the `root_dir` trait is a bit problematic relative to the [`FileContentsManager`](https://github.com/jupyter/notebook/blob/master/notebook/services/contents/filemanager.py#L70) implementation in that it a) requires that directory already exist and b) would conflict from a configuration standpoint if users wanted to use `FileContentsManager` for notebook handling as well.
3. I found it necessary to hack the setting of `root_dir` to work around the issues in 2, but that would only be necessary for `FileContentsManager`.  This led to having an `isinstance` check, which isn't cool.
4. Having an architecture that requires explicit `KernelSessionManager` sub-classes allows the best of both worlds.  Then `KernelSessionManager` subclasses can add their own "setup" and choose if they want to use a `ContentsManager` implementation for persistence themselves.  This allows for a cleaner separation and additional flexibility.  I believe some might find that rolling their own, specific, persistence implementations is more performant, etc.

kevin-bates:
I'd like to ensure some ContentsManager-type behavior/structure that would make things easy should a user wish to use a ContentsManager implementation, so changing to WIP to look into that.

PR #651, opened 2019-05-01
Add ingress resource to helm chart

akchinSTC:
My stab at #626. 
- Ingress template is parameterized with a boolean to enable/disable. 
- Default is set to assume user is using nginx as ingress controller.

Working on including Istio as well.

### DIFF
--- a/etc/kubernetes/helm/templates/ingress.yaml
+++ b/etc/kubernetes/helm/templates/ingress.yaml
@@ -1,0 +1,24 @@
+{{ if .Values.ingress.enabled }}
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  namespace: {{ .Values.ingress.namespace }}
+  name: enterprise-gateway-ingress
+{{- with .Values.ingress.annotations }}
+  annotations:
+{{ toYaml . | indent 4 }}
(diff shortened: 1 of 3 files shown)


etc/kubernetes/helm/templates/ingress.yaml
```diff
@@ -0,0 +1,24 @@
+{{ if .Values.ingress.enabled }}
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  namespace: {{ .Values.ingress.namespace }}
```
kevin-bates: Let's default this to the same namespace configured on the helm command line.  This is represented by `{{ .Release.Namespace }}`

etc/kubernetes/helm/values.yaml
```diff
@@ -23,4 +23,16 @@ nfs:
   enabled: false
   internal_server_ip_address:
 
+ingress:
+  enabled: false
+  namespace: enterprise-gateway
```
kevin-bates: Since we're using `{{ .Release.Namespace }}` for the namespace, this entry can be removed.

akchinSTC:
Good to merge..will open another PR for Istio later

PR #662, opened 2019-05-09
Update helm chart with best practices

akchinSTC:
Working on #657 according to helm best practices..so far
- updated variables to camel case
- included descriptions for each parameter in values.yaml
- logically nested some of the values for better understanding ( even though helm suggests flat values as better.. I can change if needed)
-  included jupyter icon in chart as suggested by helm lint
- updated kubernetes documentation


etc/kubernetes/helm/enterprise-gateway/values.yaml
```diff
@@ -0,0 +1,74 @@
...
+# Whether to mirror working directories.
+mirrorWorkingDirs: false
+# Master public IP on which to expose EG.
+k8sMasterPublicIP:
+
+kernel:
+  # Kernel cluster role created by this chart.
+  clusterRole: kernel-controller
+  # Will start kernels in the same namespace as EG if True. Otherwise kernel will have its own
+  sharedNamespace: false
```
kevin-bates: This sounds a bit ambiguous to me since users that "bring their own namespace" could be considered shared as well.  However about something like `share_gateway_namespace`?  If so, we should probably update the use of the `shared_namespace` variable in k8s.py.

akchinSTC:
unwip'ed, helm dry run looks ok. think we can merge this. Thanks!

PR #673, opened 2019-05-17
Fix container termination signal handling in EG container

### LINKED ISSUE #672: Kernels get orphaned when EG pod is stopped
I'm opening this for the issue we started with @kevin-bates by the way in #671.

The issue is: the shutdown of the EG pod orhpans kernel pods in 100% cases. The kernel pods (and namespaces) still live, after the EG pod is stopped. If a new EG pod is started, it's not aware of the kernel pods.

The expected behaviour is that k8s sends `SIGTERM` to the EG pod that then reacts and iterates over the kernel pods to shut them down OR kernel session is persisted so the newly-created EG pod knows about the old kernels.

@kevin-bates's quotes from the other thread:

> Yes, I reproduced the orphaned kernel issue. We typically don't shutdown EG - at least until all kernels have been shutdown themselves. To "shutdown" EG in k8s, what are you doing? Are you stopping the service? deleting the deployment? When I reproduced the issue, I deleted the service.

> Ok. I've tried spending a bit more time with this tonight. Actually, deleting the service is not sufficient since I can just restart the service and the kernel is still usable. However, when I delete the EG pod, another is started, but it doesn't know about the running kernel(s) - which makes sense. I think we could address this by enabling kernel session persistence, but first, we should understand why the appropriate shutdown protocol is not evident in the pod. According to the docs, [`SIGTERM` should be sent to the EG during the 30 second grace period](https://kubernetes.io/docs/concepts/workloads/pods/pod/#termination-of-pods) when deleting the pod and [EG's built-in signal handler should cycle through and attempt to shutdown the kernels](https://github.com/jupyter/enterprise_gateway/blob/master/enterprise_gateway/enterprisegatewayapp.py#L293).

kevin-bates:
When pods are deleted, kubernetes will send a SIGTERM to the running process
to allow the process to gracefully terminate.  For Enterprise Gateway, this
should include the shutdown of each running kernel, including the appropriate
deletion of the kernel pod and, possibly, its namespace (if EG created one).
Because the ENTRYPOINT was a shell script, the signal was not propagating to
the python process. This change restores the already existing `tini` program
as the ENTRYPOINT (inherited from minimal-notebook) and sets the command
to `start-enterprise-gateway.sh` - which then execs the python script.  Once
in place, EG pod termination leads to the termination of running kernels as
well - just as is the case on bare metal installations.

Fixes #672

Special thanks to @rolweber and @michzimny for their suggestions to a solution.

Those wishing to try this out can find an image in docker hub at [`kbates/enterprise-gateway:signal`](https://cloud.docker.com/repository/docker/kbates/enterprise-gateway/tags).


michzimny:
I have deployed and tested the new image in a real environment. It works now as expected. I think this PR can be merged.

kevin-bates:
Thanks for the confirmation Michal - merging now.

PR #677, opened 2019-06-02
Docs enhancements

echarles:
A few proposals to make the docs more readable and fix minor issues (links...):

+  fix broken download link and use latest release
+ add missing yml extension to docker-compose file
+ dedicated section on orechtestrators in toc
+ user kern-... as file name to describe kernels running on resource managers / orchestrators

Issue #346, opened 2018-05-15
Can we add kernelspecs for kernels with spark standalone mode also. 


agsachin:
Can we add standalone specs using both LocalProcessProxy(default proxy) and DistributedProcessProxy.
https://github.com/jupyter-incubator/enterprise_gateway/tree/master/etc/kernelspecs	
spark_R_standalone
spark_python_standalone
spark_scala_standalone

kevin-bates:
Note that this is now documented in our docs: https://jupyter-enterprise-gateway.readthedocs.io/en/latest/kernel-spark-standalone.html

Since kernelspecs, especially those that are Spark-based, require customizations to each environment, and to reduce the support matrix involved in maintaining physical copies that differ slightly from the YARN client deliverables, I'm closing this issue.

When addressing issue #654 or #655, we can be sure to add a standalone option, but for now I think the documentation should prove sufficient.

PR #680, opened 2019-06-04
Docs Enhancement (part 2)

echarles:
As follow-up of #677 I bring here more enhancements. There is a bit of TOC change... but information has not been removed - The reader path is smoother and more logical.

rename kern-* files to kernel-*
split yarn-client-spark-standalone in two
mv scala/python/r description to yarn-cluster and yarn-client + fix some bad path
mv getting-started-kernel in getting-started (makes more sense to be there directly)
rephrase the getting-started to make it more generic in terms of resource manager (don't use YARN but resource manager)


docs/source/kernel-yarn-client-mode.md
```diff
@@ -0,0 +1,133 @@
...
+    "{kernel_id}",
+    "--RemoteProcessProxy.response-address",
+    "{response_address}"
+  ]
+}
+```
+
+After making any necessary adjustments such as updating SPARK_HOME or other environment specific configuration, you now should have a new Kernel available which will use Jupyter Enterprise Gateway to execute your notebook cell contents.
+
+### Scala Kernel (Apache Toree kernel)
```
kevin-bates: The installation instructions appear duplicated in Yarn cluster and client but don't exist in standalone. 

I think we should probably have a section dedicated to kernel installation - at least for the non-containerized stuff.  Its important to point out that the kernels be installed on all nodes of the cluster with the exception of Apache Toree in Yarn cluster mode since spark-submit will transfer the files at startup.

For distributed modes (Yarn client and Spark Standalone and vanilla across a set of hosts) all kernels **and their corresponding kernelspecs** must reside on each node of the cluster because we essentially use a remote shell to invoke the kernel - so the kernelspec hierarchies (i.e., paths) must be identical.

echarles:
Hi Kevin,
1. I had those warning from the very first build before any changes and also though using`.md` instead of `.html` but didn't want to go that way for now (.html was maybe needed for PDF links... or whatever). Maybe you see them because you updated your sphinx env...
2. I have pushed a commit which takes into account most of your comments.
3. For spark standalone, there is no dedicated spec but it uses yarn_client ones. I have tried to reflect that in the doc but it sounds to me like a separated spec would be welcome to avoid name confusion.
4.  Now the last non-resolved conversation here above about what you say: `I think we should probably have a section dedicated to kernel installation` -> that general section is now in `getting-started-kernels#Installing_Kernels` - I have moved the specific paragraphs (scala/python/r) to yarn_cluster and yarn_client because the initial text was talking about that. My understanding is that we will have after to also add more specifics for scala/python/R in the k8s... pages. Does it make sense?

echarles:
I will be off soon and will push something tomorrow. I get all you points except the 4th where you talk about `container customization, `standalone`. I think I am missing something and it may be worth to let you explain yet a bit more how you see the structure (this will help also to better understand your logic of kernels...). Maybe a simple TOC for the kernels/resource manager parts with a few words on each leaf would help?

echarles:
Maybe this list can help. I have assigned the kernelspecs to their "resource manager" (the idea is to have one page per resource manager + 1 page for local):

Local: R_docker, python_distributed, python_docker, scala_docker, python_tf_docker, python_tf_gpu_docker

> Comment: funny to name distributed for something supposed to be `local` - TensorFlow is about distributed computing, does it make sense in `local`?

YARN client: spark_R_yarn_client, spark_python_yarn_client, spark_scala_yarn_client

YARN cluster: dask_python_yarn_remote, spark_scala_yarn_cluster, spark_R_yarn_cluster, spark_python_yarn_cluster

K8S: R_kubernetes, python_kubernetes, scala_kubernetes, spark_R_kubernetes, spark_python_kubernetes, python_tf_gpu_kubernetes, python_tf_kubernetes, spark_scala_kubernetes,

Conductor: spark_R_conductor_cluster, spark_python_conductor_cluster, 
 spark_scala_conductor_cluster

kevin-bates:
This seems really good, although I think we should drop local and add a docker section. The real intent for docker is that they be used in docker swarm.  python_distributed should be an example of a DistributedProcessProxy that is "vanilla" (i.e., non spark, yarn, etc.)

Actually, one could argue that Yarn client should be 'Distributed' and we can add python_distributed there.  They all use the DistributedProcessProxy (and ssh).

echarles:
I now understand your not-yet process comments and have tried to formalise them in the doc. Could you review the last commit:
- I have not created a separated page yarn/spark-standalone but have duplicated the setup on each of those pages.
- I have created and referenced #681 for spark-standalone kernelspec
- I have clearly stated the 2 cases (docker / not-docker) in the `Installing Kernels` section.
- I have added a page `kernel-vanilla` -> I have put questions there (list of nodes...) - Can you bring light on these? Thx.

docs/source/kernel-spark-standalone.md
```diff
@@ -0,0 +1,142 @@
...
+
+* EG_REMOTE_HOSTS must be set to a comma-separated set of FQDN hosts indicating the hosts available for running kernels. (This can be specified via the command line as well: `--EnterpriseGatewayApp.remote_hosts`)
+
+```
+EG_REMOTE_HOSTS=elyra-node-1.fyre.ibm.com,elyra-node-2.fyre.ibm.com,elyra-node-3.fyre.ibm.com,elyra-node-4.fyre.ibm.com,elyra-node-5.fyre.ibm.com
+```
+
+### Configuring Kernels for Spark Standalone
+
+We will use some `yarn_client` specs for Spark Standalone. Follow [#681](https://github.com/jupyter/enterprise_gateway/issues/681) that will ensure Spark Standalone gets its own kernelspec.
```
kevin-bates: #681 is a duplicate of #346.  I'd prefer we not reference an open issue in the docs but instead say something something like, _Although Enterprise Gateway does not currently provide sample kernelspecs for Spark standalone, here are the steps necessary to convert a `yarn_client` kernelspec to standalone._

(We view the kernelspecs we do provide as samples since, in nearly all cases, tweaks and customizations for the target environment will be required.)

docs/source/kernel-spark-standalone.md
```diff
@@ -0,0 +1,142 @@
...
+  },
+  "argv": [
+    "/usr/local/share/jupyter/kernels/spark_python_spark_standalone/bin/run.sh",
+     "--RemoteProcessProxy.kernel-id",
+    "{kernel_id}",
+    "--RemoteProcessProxy.response-address",
+    "{response_address}"
+  ]
+}
+```
```
kevin-bates: Looks good  - thank you!

docs/source/kernel-vanilla.md
```diff
@@ -0,0 +1,19 @@
...
+
+``` Bash
+wget https://github.com/jupyter/enterprise_gateway/releases/download/v2.0.0rc1/jupyter_enterprise_gateway_kernelspecs-2.0.0rc1.tar.gz
+KERNELS_FOLDER=/usr/local/share/jupyter/kernels
+tar -zxvf jupyter_enterprise_gateway_kernelspecs-2.0.0.dev2.tar.gz --strip 1 --directory $KERNELS_FOLDER/python_distributed/ python_distributed/
+```
+
+The `python_distributed` kernelspec uses `DistributedProcessProxy` which is responsible for the launch and management of kernels distributed across and explicitly defined set of hosts using ssh. Hosts are determined via a round-robin algorithm (that we should make pluggable someday).
+
+**QUESTIONS: how to set the list of hosts - Also, SSH should be available passwordless, right? Which user should be used?***
```
kevin-bates: The set of remote hosts are derived from two places.  
1. The configuration option: `EnterpriseGatewayApp.remote_hosts` who's default value comes from the env variable `EG_REMOTE_HOSTS` - which, itself, defaults to 'localhost'.
2. The config option can be [overridden on a per-kernel basis](https://jupyter-enterprise-gateway.readthedocs.io/en/v2.0.0rc1/config-options.html#per-kernel-configuration-overrides) if the `process_proxy` stanza contains a `config` stanza where there's a `remote_hosts` entry.  If present, this value will be used instead.

Unfortunately, it doesn't look like the SSH requirement is very well stated.  We touch on this when describing [`DistributedProcessProxy`](https://jupyter-enterprise-gateway.readthedocs.io/en/v2.0.0rc1/system-architecture.html#distributedprocessproxy) and in the [troubleshooting section](https://jupyter-enterprise-gateway.readthedocs.io/en/v2.0.0rc1/troubleshooting.html), but nothing of any substance.

cc: @lresende 

We recommend running EG as the _gateway user_ which is touched on in the security section - but mostly relative to impersonation topics.

docs/source/kernel-kubernetes.md
```diff
@@ -4,6 +4,17 @@ This page describes the approach taken for integrating Enterprise Gateway into a
 
 In this solution, Enterprise Gateway is, itself, provisioned as a Kubernetes _deployment_ and exposed as a Kubernetes _service_.  In this way, Enterprise Gateway can leverage load balancing and high availability functionality provided by Kubernetes (although HA cannot be fully realized until EG supports persistent sessions).
 
+Following kernelspecs are available on Kubernetes:
```
kevin-bates: I think we should emphasize that these are _sample_ kernelspecs provided OOTB.  So perhaps _The following sample kernelspecs are currently available on Kubernetes:_

docs/source/kernel-vanilla.md
```diff
@@ -0,0 +1,19 @@
+## Vanilla
+
+This page describes the approach taken for integrating Enterprise Gateway into a vanilla set of hosts, without cluster resource managers.
+
+Following kernelspec is available on Vanilla mode:
```
kevin-bates: _The following sample kernelspec is currently available_ ...

docs/source/kernel-yarn-client-mode.md
```diff
@@ -0,0 +1,140 @@
+## YARN Client Mode
+
+Jupyter Enterprise Gateway extends Jupyter Kernel Gateway which means that by installing kernels in Enterprise Gateway and using the vanila kernelspecs created during installation you will have your kernels running in client mode with drivers running on the same host as Enterprise Gateway. 
+
+Having said that, even if you are not leveraging the full distributed capabilities of Jupyter Enterprise Gateway, client mode can still help mitigate resource starvation by enabling a pseudo-distributed mode, where kernels are started in different nodes of the cluster utilizing a round-robin algorithm. In this case, you can still experience bottlenecks on a given node that receives requests to start "large" kernels, but otherwise, you will be better off compared to when all kernels are started on a single node or as local processes, which is the default for vanilla Jupyter Notebook.
+
+Following kernelspecs are available on YARN client:
```
kevin-bates: _The following sample kernelspecs are currently available_ ...

docs/source/kernel-yarn-cluster-mode.md
```diff
@@ -0,0 +1,134 @@
+## YARN Cluster Mode
+
+To leverage the full distributed capabilities of Jupyter Enterprise Gateway, there is a need to provide additional configuration options in a cluster deployment.
+
+Following kernelspecs are available on YARN cluster:
```
kevin-bates: _The following sample kernelspecs are currently available_ ...

docs/source/kernel-docker.md
```diff
@@ -6,7 +6,16 @@ In this solution, Enterprise Gateway is, itself, provisioned as a Docker Swarm _
 
 The base Enterprise Gateway image is [elyra/enterprise-gateway](https://hub.docker.com/r/elyra/enterprise-gateway/) and can be found in the Enterprise Gateway dockerhub organization [elyra](https://hub.docker.com/r/elyra/), along with other images.  See [Runtime Images](docker.html#runtime-images) for image details.
 
+Following kernelspecs are available on Docker:
```
kevin-bates: _The following sample kernelspecs are currently available_ ...

docs/source/kernel-conductor.md
```diff
@@ -1,3 +1,9 @@
 ## IBM Spectrum Conductor
 
 This information will be added shortly.  The configuration is similar to that of [YARN Cluster mode](kernel-yarn-cluster-mode.html) with the `ConductorClusterProcessProxy` used in place of `YARNClusterProcessProxy`.
+
+Following kernelspecs are available on Conductor:
```
kevin-bates: _The following sample kernelspecs are currently available_ ...

docs/source/kernel-vanilla.md
```diff
@@ -0,0 +1,19 @@
+## Vanilla
```
kevin-bates: Although we use the term `vanilla` to describe a plain environment/kernel, I'm not sure that concept should be officially published by that name.  I'm thinking `Distributed` might be more appropriate, but I'd love to hear ideas.  

Btw, Yarn client mode is exactly this as well, we just happen to use spark-submit form different nodes in the cluster, that's all.

docs/source/getting-started.md
```diff
@@ -76,23 +76,39 @@ conda uninstall jupyter_enterprise_gateway
...
+
+We have two cases:
 
-* [Enabling YARN Cluster Mode support](kernel-yarn-cluster-mode.html)
-* [Enabling YARN Client Mode or Spark Standalone support](kernel-yarn-client-mode.html)
-* [Enabling Kubernetes Support](kernel-kubernetes.html)
-* [Enabling Docker Swarm Support](kernel-docker.html)
-* [Enabling IBM Spectrum Conductor Support](kernel-conductor.html)
+1. The kernel is run in (Docker) container: in that case, the image should ensure the availability of the kernel libraries and kernelspec. There is no additional action to take on the node.
+2. The kernel is not run in a (Docker) container, e.g. Vanilla across a set of hosts, YARN (cluster mode, client mode) or Spark Standalone: All kernels (libraries...) and their corresponding kernelspecs must reside on each nodes. This requirement is applicable for vanilla using a set of distributed hosts and for all reource managers. The kernelspec hierarchies (i.e., paths) must be available and identical on all nodes. This is not applicable to Apache Toree in Yarn cluster mode since spark-submit will transfer the files at startup.
```
kevin-bates: The kernelspecs are only required on all nodes if using the DistributedProcessProxy - which can apply to YARN Client mode, Standalone, and _Vanilla_.  (See comment at start of Vanilla section)

docs/source/getting-started.md
```diff
@@ -76,23 +76,39 @@ conda uninstall jupyter_enterprise_gateway
...
+**Important requirements regarding the nodes**
+
+We have two cases:
 
-* [Enabling YARN Cluster Mode support](kernel-yarn-cluster-mode.html)
-* [Enabling YARN Client Mode or Spark Standalone support](kernel-yarn-client-mode.html)
-* [Enabling Kubernetes Support](kernel-kubernetes.html)
-* [Enabling Docker Swarm Support](kernel-docker.html)
-* [Enabling IBM Spectrum Conductor Support](kernel-conductor.html)
+1. The kernel is run in (Docker) container: in that case, the image should ensure the availability of the kernel libraries and kernelspec. There is no additional action to take on the node.
```
kevin-bates: in a (Docker) container.

The kernelspec is not necessary here, only the launcher.  We talk about this in container customization.

echarles:
Just pushed a commit which takes into account all your comments and proposals:

My questions for the next iteration:

+ You say for containers: `The kernelspec is not necessary here, only the launcher. We talk about this in container customization.` : The `launcher` concept is not really defined, isn't it? Is this just the script without the json.file?
+ Vanilla is out for sure... I have used `Distributed` as I don't find any other one.
+ ... and for SSH the reader needs to wait for the `system-architecture` to understand that DistributedProcessProxy uses SSH and that he should ensure passwordless SSH. I wonder if the system-architecture should move up in the TOC after the getting started and before the kernels. I have added in the `getting-started` page for the distributed case: `SSH passwordless is needed between the EG node and the other nodes.`
+ BTW what is `SSH Tunneling` explained in security page.

kevin-bates:
>The launcher concept is not really defined, isn't it? Is this just the script without the json.file?

We talk about the launchers in the section regarding [container customization](https://jupyter-enterprise-gateway.readthedocs.io/en/latest/docker.html#kernel-launcher) which then links for more details.  We provide a zip file with this information that needs to be added to a custom image (provided its not extending an existing image.

>Vanilla is out for sure... I have used Distributed as I don't find any better name.

:+1:

>... and for SSH the reader needs to wait for the system-architecture to understand that DistributedProcessProxy uses SSH and that he should ensure passwordless SSH. I wonder if the system-architecture should move up in the TOC after the getting started and before the kernels. I have added in the getting-started page for the distributed case: SSH passwordless is needed between the EG node and the other nodes.

That sounds like a good approach to me.

>BTW what is SSH Tunneling explained in security page.

Since kernels are remote from the "server" messages are sent over the wire and potentially exposed.  The tunneling enablement allows users in more secure environments to enable the tunneling of the 5 kernel ports so that traffic between the server and kernel is protected.

echarles:
Thx Kevin for your didactic explanations. I have moved up the `system-architecture`. Yes, the Kernel launchers are well described.

It looks good to me now, just one question remaining: in `getting-started`, I don't get completely that sentence `1. The kernel is run in (Docker) container: in that case, the image should ensure the availability of the kernel libraries and kernelspec. The kernelspec is not necessary here, only the launcher. We talk about this in container customization.`. My understanding is that the docker container is launched via SSH (simple docker) or by Swarm itself (docker swarm) with some parameters defined by the kernelspec on the EG node. Then the docker container is launched and contains what it is supposed to contain (python/R, kernelspec and kernel launcher...). Why should the remote node have a kernel launcher on the node? Or do you mean that the container itself should only contain the launcher, not the spec?

docs/source/getting-started.md
```diff
@@ -76,23 +76,40 @@ conda uninstall jupyter_enterprise_gateway
...
+**Important requirements regarding the nodes**
 
-* [Enabling YARN Cluster Mode support](kernel-yarn-cluster-mode.html)
-* [Enabling YARN Client Mode or Spark Standalone support](kernel-yarn-client-mode.html)
-* [Enabling Kubernetes Support](kernel-kubernetes.html)
-* [Enabling Docker Swarm Support](kernel-docker.html)
-* [Enabling IBM Spectrum Conductor Support](kernel-conductor.html)
+We have two cases:
+
+1. The kernel is run in (Docker) container: in that case, the image should ensure the availability of the kernel libraries and kernelspec. The kernelspec is not necessary here, only the launcher. We talk about this in container customization.
```
kevin-bates: Perhaps we could say:
_1. The kernel is run in a (Docker) container: in that case, the image should ensure the availability of the kernel libraries and its corresponding launcher.  We talk about the kernel launchers in [container customization](docker.html#bringing-your-own-kernel-image)._  (Not sure if that's the correct format for an anchored link off the top of my head.)

kevin-bates:
I've added a comment relative to that particular section that might be helpful.

The launch of containerized kernels (at least via EG) is two-fold.  First, there's the argv section in the kernelspec that is processed by EG.  In these cases, the command that is invoked is a python script using the container's api (kubernetes, docker, or docker swarm) that is responsible for converting any necessary "parameters" to environment variables, etc. that are used during the actual container creation.  

The command that is then run in the container is THEN the actual kernel launcher script. This launcher is responsible for taking the response address (which is now an env variable) and returning the kernel's connection information back on that response address to EG.  The kernel launcher does additional things - but primarily listens for interrupt and shutdown requests, which it then passes along to the actual (embedded) kernel.

So container envs have two launches - one to launch the container itself, the other to launch the kernel (within the container).

echarles:
Pushed your last inputs (in `getting-started` with subheading to make it readable). We are nearly there...

We have `*Case 2 - The kernel is run via DistributedProcessProxy* The kernelspecs are only required on all nodes if using the DistributedProcessProxy - which apply to YARN Client mode, Standalone, and 
Distributed modes.` What about YarnClusterProcessProxy? Shouldn't also the kernelspecs be deployed on all hosts? We should anyway cover YarnClusterProcessProxy in (1) A third case (2) rename Case 2 to "non docker"

docs/source/getting-started.md
```diff
@@ -90,20 +90,35 @@ Please follow the links below to learn specific details about how to enable/conf
...
+
+The command that is then run in the container is THEN the actual kernel launcher script. This launcher is responsible for taking the response address (which is now an env variable) and returning the kernel's connection information back on that response address to EG. The kernel launcher does additional things - but primarily listens for interrupt and shutdown requests, which it then passes along to the actual (embedded) kernel.
+
+So container envs have two launches - one to launch the container itself, the other to launch the kernel (within the container).
+
+*Case 2 - The kernel is run via DistributedProcessProxy*
+
+The kernelspecs are only required on all nodes if using the DistributedProcessProxy - which apply to YARN Client mode, Standalone, and 
+Distributed modes. All kernels (libraries...) and their corresponding kernelspecs must reside on each nodes. The kernelspec hierarchies (i.e., paths) must be available and identical on all nodes. This is not applicable to Apache Toree in Yarn cluster mode since spark-submit will transfer the files at startup. SSH passwordless is needed between the EG node and the other nodes.
```
kevin-bates: hmm.  I think we could use a third case here since theres some convolution going on.  How about something like this...

*Case 2 - The kernel is run via DistributedProcessProxy*

The kernelspecs are required on all nodes if using the `DistributedProcessProxy` - which apply to YARN Client mode, Standalone, and Distributed modes. All kernels (libraries...) and their corresponding kernelspecs must reside on each node. The kernelspec hierarchies (i.e., paths) must be available and identical on all nodes.  SSH passwordless is needed between the EG node and the other nodes.

*Case 3 - The kernel is run via YarnClusterProcessProxy or ConductionClusterProcessProxy*

With _cluster_ process proxies, distribution of kernelspecs to all nodes besides the EG node is not required. However, the IPython and R Kernel kernel must be installed on each node. Note that because the Apache Toree kernel, and its supporting libraries, will be transferred to the target node via spark-submit, installation of Apache Toree (the scala kernel) is not required except on the Enterprise Gateway node itself.


Based on this, I'm thinking Case 1 may want to read something like...

*Case 1 - The kernel is run in via a container-based process proxy (Kubernetes, Docker or DockerSwarm)*

Not sure the parenthesized details are necessary but I think the use of 'process proxy' is more consistent across the 3.

echarles:
@kevin-bates Latest commit covers your latest comments (3rd case + the rephrasing you proposed). I went through all the pages and I feel good with them.

Oh... One more iteration? I have added a `Local Mode` page intended to document how to run the kernels locally. One could argue this is out of scope, but I think Local mode deserves a place even for an enterprise solution. Big Data tools (Hadoop, Spark...) have a documented and supported local mode.

BTW I can understand the reasoning behind the name `enterprise gateway` (you know: scalable, secured and so on...) but it often strikes my mind as this should not be only for the `enterprises` (academics, small groups... are also potential users). I agree `distributed gateway` is not good either as the gateway is not distributed, but well the kernels... anyway... what's in a name

docs/source/getting-started.md
```diff
@@ -90,38 +92,46 @@ Please follow the links below to learn specific details about how to enable/conf
 
 In each of the resource manager sections, we set the `KERNELS_FOLDER` to `/usr/local/share/jupyter/kernels` since that's one of the default locations searched by the Jupyter framework.  Co-locating kernelspecs hierarchies in the same parent folder is recommended, although not required.
 
-#### Important requirements regarding the nodes
+Depending on the resource manager, we details in the related section the implemented kernel languages (python, scala, R...). The following kernels have been tested with the Jupyter Enterprise Gateway:
```
kevin-bates: details -> detail

docs/source/kernel-local.md
```diff
@@ -0,0 +1,28 @@
...
+
+If you just want to try EG in a local setup, you can use the following kernelspec:
+
+```json
+{
+  "display_name": "Python 3 Local",
+  "language": "python", 
+  "metadata": {
+    "process_proxy": {
+      "class_name": "enterprise_gateway.services.processproxies.processproxy.LocalProcessProxy"
```
kevin-bates: Just FYI (but might be worth mentioning), if Enterprise Gateway encounters a kernelspec without the `process_proxy` stanza, it will treat that kernelspec as if it contained `LocalProcessProxy`.  There are some basic "features" that local kernels can leverage from EG, but, for the most part, this is purely for backwards compatibility.

I agree that it makes sense to discuss this nonetheless - thank you!

echarles:
yep, I got the background and future looks like I love it.

Last bit to complete this PR on my side: can you give me a snippet for the launcher for local mode. The following can not stay like this :)

with the following `launch_ipykernel.py`.

```python
# ???
pass
```

kevin-bates:
The launchers are purely for remote kernels (RemoteProcessProxy subclasses).  LocalProcessProxy just takes the native kernel.json file - no need to wrap in a launcher.

Side note: You _can_ run a local kernel in "distributed mode" by setting `remote_hosts` to the local host.  Why would you do that?  

One reason is that it decreases the window in which a port conflict can occur since the 5 kernel ports are created by the launcher (within the same process and therefore closer to the actual invocation of the kernel) rather than by the server prior to the launch of the kernel process.  

The second reason is that auto-restarted kernels - when an issue occurs - say due to a port conflict - will create a new set of ports rather than try to re-use the same set that produced the failure in the first place.  In this case, should anyone want to do this, you'd want to use the [per-kernel configuration](https://jupyter-enterprise-gateway.readthedocs.io/en/latest/config-options.html#per-kernel-configuration-overrides) approach and set `remote_hosts` in the `config` stanza of the `process_proxy` stanza.

echarles:
That's it. I have literally copied/pasted many @kevin-bates great comments. But good question drive great answers :) LOL

Hope this will make other guys better understand EG like I now better understand it.

Happy to get any other reviews. On my side, I have reached my goal with this PR. Thx again.

docs/source/kernel-spark-standalone.md
```diff
@@ -0,0 +1,142 @@
...
+
+``` Bash
+wget https://github.com/jupyter/enterprise_gateway/releases/download/v2.0.0rc1/jupyter_enterprise_gateway_kernelspecs-2.0.0rc1.tar.gz
+SCALA_KERNEL_DIR="$(jupyter kernelspec list | grep -w "python3" | awk '{print $2}')"
+KERNELS_FOLDER="$(dirname "${SCALA_KERNEL_DIR}")"
+tar -zxvf enterprise_gateway_kernelspecs.tar.gz --strip 1 --directory $KERNELS_FOLDER/spark_python_yarn_client/ spark_python_yarn_client/
+mv $KERNELS_FOLDER/spark_python_yarn_client $KERNELS_FOLDER/spark_python_spark_standalone
+```
+
+After that, you should have a kernel.json that looks similar to the one below:
```
kevin-bates: I think we need to mention the edits that need to occur to convert a yarn-client kernel.json file to standalone prior to this statement.

The edits are to update the `display_name` (however they like, your text is great) and to update the `--master` option in the `SPARK_OPTS` to point to the spark master node rather than indicate `--deploy-mode client`.

I think we would probably drop off the `spark.yarn.submit.waitAppCompletion` option as well, although I'm not certain about that.  (@lresende - is `--conf spark.yarn.submit.waitAppCompletion` still necessary in standalone mode?)

After THAT, they should see a kernel.json ... :smiley:

docs/source/kernel-local.md
```diff
@@ -0,0 +1,31 @@
+## Local Mode
+
+The Local deployment is just for local development and is not meant to be run in real life operations.
```
kevin-bates: Let's replace 'is just for' with 'can be useful for' and 'real life operations' with 'production environments as subjects the gateway server to resource exhaustion'.

_The Local deployment can be useful for local development and is not meant to be run in production environments as it subjects the gateway server to resource exhaustion._

docs/source/kernel-local.md
```diff
@@ -0,0 +1,31 @@
...
+  ]
+}
+```
+
+`process_proxy` is optional (if Enterprise Gateway encounters a kernelspec without the `process_proxy` stanza, it will treat that kernelspec as if it contained `LocalProcessProxy`).
+
+Side note: You can run a Local kernel in [Distributed mode](./kernel-distributed.html) by setting `remote_hosts` to the localhost. Why would you do that?
+
+1. One reason is that it decreases the window in which a port conflict can occur since the 5 kernel ports are created by the launcher (within the same process and therefore closer to the actual invocation of the kernel) rather than by the server prior to the launch of the kernel process.
+2. The second reason is that auto-restarted kernels - when an issue occurs - say due to a port conflict - will create a new set of ports rather than try to re-use the same set that produced the failure in the first place. In this case, you'd want to use the [per-kernel configuration](./config-options.html#per-kernel-configuration-overrides) approach and set `remote_hosts` in the config stanza of the `process_proxy` stanza.
```
kevin-bates: can we append: _(so as to not prevent Standalone or YARN Client kernels from leveraging the full cluster)._

echarles: not sure to understand... where to append `(so as to not prevent Standalone or YARN Client kernels from leveraging the full cluster).` - Do you mean Spark Standalone?

docs/source/kernel-spark-standalone.md
echarles: I have removed spark.yarn.submit.waitAppCompletion (it is about yarn, not spark cluster), I can add it back if @lresende says otherwise.

docs/source/kernel-local.md
kevin-bates: At the end of the last sentence in bullet 2.  I want readers to understand why they'd want to update the kernel.json config stanza as opposed to `EG_REMOTE_HOSTS` (global config).

echarles:
2 more fixes pushed.

echarles:
@lresende I think we are missing your review to merge this. Thx, Eric

lresende:
Don't need to wait for me, I trust that @kevin-bates has done a great review here.

kevin-bates:
@echarles - could you please rebase against EG/master to resolve the conflict?  I'll squash/merge after that.  Thanks.

echarles:
@kevin-bates done.

kevin-bates:
Thank you for this great contribution!

PR #688, opened 2019-06-13
Fix scoping issues relative to python kernels on Spark

### LINKED ISSUE #687: variable scope problems in function definition
Hi there,

I'm using enterprise gateway on our kubernetes cluster (following [this](https://blog.jupyter.org/on-demand-notebooks-with-jupyterhub-jupyter-enterprise-gateway-and-kubernetes-e8e423695cbf) instructions).

Using a python notebook (in my case the kernel is `elyra/kernel-spark-py:dev`) I found the problem, that I cannot use variables inside a function definition, which are defined outside. 

Example:
``` 
a = 123

def fun():
    print(a)
    
fun()
```
ends in:
```
---------------------------------------------------------------------------
NameError                                 Traceback (most recent call last)
/usr/local/bin/kernel-launchers/python/scripts/launch_ipykernel.py in <module>
     4     print(a)
     5
----> 6 fun()

/usr/local/bin/kernel-launchers/python/scripts/launch_ipykernel.py in fun()
     2
     3 def fun():
----> 4     print(a)
     5
     6 fun()

NameError: name 'a' is not defined
```

Same with import...

Any ideas?

kevin-bates:
The python kernel launcher was not initializing the namespace used
in the embed_kernel properly.  This change reverts an earlier change
to use the spark context variables from the global scope, instead
moving them to the local namespace.

Thanks to @jcrist for significant insight and guidance!

Fixes #687

### DIFF
--- a/etc/kernel-launchers/python/scripts/launch_ipykernel.py
+++ b/etc/kernel-launchers/python/scripts/launch_ipykernel.py
@@ -28,7 +28,7 @@ logger = logging.getLogger('launch_ipykernel')
 logger.setLevel(log_level)
 
 
-def initialize_namespace(cluster_type='spark', spark_ns=None):
+def initialize_namespace(namespace, cluster_type='spark'):
     """Initialize the kernel namespace.
 
     Parameters
@@ -45,7 +45,7 @@ def initialize_namespace(cluster_type='spark', spark_ns=None):
                         "Spark context creation will not occur.")
             return {}
 
-        def initialize_spark_session(ns):
+        def initialize_spark_session():
             """Initialize Spark session and replace global variable
             placeholders with real Spark session object references."""
             spark = SparkSession.builder.getOrCreate()
@@ -57,36 +57,34 @@ def initialize_namespace(cluster_type='spark', spark_ns=None):
             # Stop the spark session on exit
             atexit.register(lambda: spark.stop())
 
-            ns.update({'spark': spark,
+            namespace.update({'spark': spark,
                               'sc': spark.sparkContext,
                               'sql': spark.sql,
                               'sqlContext': spark._wrapped,
                               'sqlCtx': spark._wrapped})
 
-        init_thread = Thread(target=initialize_spark_session, args=(spark_ns,))
-        spark = WaitingForSparkSessionToBeInitialized('spark', init_thread)
-        sc = WaitingForSparkSessionToBeInitialized('sc', init_thread)
-        sqlContext = WaitingForSparkSessionToBeInitialized('sqlContext', init_thread)
+        init_thread = Thread(target=initialize_spark_session)
+        spark = WaitingForSparkSessionToBeInitialized('spark', init_thread, namespace)
+        sc = WaitingForSparkSessionToBeInitialized('sc', init_thread, namespace)
+        sqlContext = WaitingForSparkSessionToBeInitialized('sqlContext', init_thread, namespace)
 
         def sql(query):
             """Placeholder function. When called will wait for Spark session to be
             initialized and call ``spark.sql(query)``"""
             return spark.sql(query)
 
-        spark_ns.update({'spark': spark,
-                     'sc': sc,
-                     'sql': sql,
-                     'sqlContext': sqlContext,
-                     'sqlCtx': sqlContext})
+        namespace.update({'spark': spark,
+                 'sc': sc,
+                 'sql': sql,
+                 'sqlContext': sqlContext,
+                 'sqlCtx': sqlContext})
 
         init_thread.start()
-
-        return spark_ns
 
     elif cluster_type == 'dask':
         import dask_yarn
         cluster = dask_yarn.YarnCluster.from_current()
-        return {'cluster': cluster}
+        namespace.update({'cluster': cluster})
     elif cluster_type != 'none':
         raise RuntimeError("Unknown cluster_type: %r" % cluster_type)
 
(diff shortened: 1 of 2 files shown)


etc/kernel-launchers/python/scripts/launch_ipykernel.py
```diff
@@ -357,14 +356,19 @@ def gateway_listener(sock, parent_pid):
...
     # Initialize the kernel namespace for the given cluster type
     if cluster_type == 'spark' and spark_init_mode == 'none':
         cluster_type = 'none'
-    namespace = initialize_namespace(cluster_type=cluster_type)
+
+    spark_ns = None
+    if cluster_type == 'spark':
+        spark_ns = locals()
+
+    namespace = initialize_namespace(cluster_type=cluster_type, spark_ns=spark_ns)
```
jcrist: This isn't quite correct. You properly mutating the locals for spark, but you need to for dask as well (this isn't spark specific, it's just a namespace issue). You're also not clearing the namespace, so it's still polluted by everything in this initialization script.

I recommend the following:

- Change `initialize_namespace` to the following signature:

```python
def initialize_namespace(namespace, cluster_type="spark"):
    """Mutate the namespace in place to initialize it for the given cluster"
    if cluster_type == "spark":
       namespace.update(...)
    elif cluster_type == "dask":
        namespace.update(...)

- Add the following function:

```python
def start_ipython(namespace, cluster_type="spark", **kwargs):
    from IPython import embed_kernel

    # create an initial list of variables to clear
    # we do this without deleting to preserve the locals so that
    # initialize_namespace isn't affected by this mutation
    to_delete = [k for k in namespace if not k.startswith('__')]
    
    # initialize the namespace with the proper variables
    initialize_namespace(namespace, cluster_type=cluster_type)

    # delete the extraneous variables
    for k in to_delete:
        del namespace[k]

    # Start the kernel
    embed_kernel(local_ns=namespace, **kwargs)
```

- Replace the code here with the following:

```python
start_ipython(locals(), cluster_type=cluster_type, connection_file=connection_file, ip=ip)
```

You may need to update the lazy loading thread code to use local imports/references as the mutation of the external locals dict in `initialize_namespace` may break things. But for the dask side this should work at least.

kevin-bates: Thanks @jcrist.  So the for loop to remove extraneous variables ends up removing the various spark variables that, I thought, we wanted.  Are you sure we only want things in the namespace that start with `__`?

jcrist: Are you saying that the returned namespace has no `sc` in it, or that the proxy wrappers as written no longer work?

The code above assumes that `initialize_namespace` only adds variables to the existing namespace, and doesn't shadow any (that may already be in the `to_delete` list). If this isn't true, perhaps `initialize_namespace` should also return a list of variables added that should be kept? We do want to delete everything that doesn't start with `__`, as the user should be presented with a (mostly) empty namespace. `__things__` are part of python's internals and shouldn't be removed.

The `WaitingForSparkSessionToBeInitialized` will need to be rewritten to forward calls to take the passed in namespace and get the wrapped variable from in there rather than using the call to `globals()` in `__getattr__`.

kevin-bates: Sorry, I misread your updated code and forgot that the `sc` variables are getting added after the set of non-__ symbols have been built.  I'll give the above a shot.

Still barely following this discussion but have learned that python is way too lenient of a language - its kind of joke IMHO.

Thank you for your help!

jcrist: > but have learned that python is way too lenient of a language - its kind of joke IMHO.

To be fair, the use case here is quite uncommon - you're trying to mutate the locals of one frame directly from a different thread. I encourage not doing this (5-10 seconds isn't much, especially if the user is already planning on using spark anyway), but if you need to you're going to have to get into the runtime which complicates things. Messing with any language's runtime is bound to expose demons.

kevin-bates: @jcrist - please see the latest incantation.  I found that `WaitingForSparkSessionToBeInitialized` and `atexit` where getting removed from the namespace, but they're needed when the various spark variables are dereferenced.  Still unsure about `atexit`, but until it was kept, dereference of the spark variables would still blow out the stack.

As these changes are right now, everything works in yarn/spark.  I can't try Dask, but I suspect it will be fine - although those two symbols will be present.

Thanks again for all your help on this!

kevin-bates:
@jcrist - could you please give this another review? Not sure if you want to do anything about the `protected_vars` for the dask case.  I suppose that array could be defined globally, then the dask code could remove them from the namespace dict when it adds 'cluster'.  Perhaps it should be renamed to something like `spark_context_functions` to fit the context better.

jcrist:
What you have now seems fine. If everything works on your end for spark, then this should also work for dask.

If you wanted to go further and fully clear the user's namespace of enterprise-gateway internals, you could do the following:

- Move the `atexit` import to inside `initialize_namespace`, removing it from the user's namespace.
- Define `WaitingForSparkSessionToBeInitialized` inside `initialize_namespace`. From looking at the code as written, I don't think things will fail if you *do* delete this name from the namespace, all instances are already created when `initialize_namespace` returns, so there should be no name error.

If you feel up for trying it, moving `atexit` to a local import in `initialize_namespace`, and removing `protected_vars` would be an easy thing to test the above. Otherwise things are fine as is, the number of extraneous names is low.

kevin-bates:
@jrcist - great points Jim!  I moved the class def and `atexit` import into the spark block of `initialize_namespace` and the need for the `protected_vars` stuff went away!  Thank you.

A final looksy would be greatly appreciated!

jcrist:
I can't be 100% sure, but I think things would have worked leaving `WaitingForSparkSessionToBeInitialized` at the top level, and just moving `atexit` to a local import. This looks fine as is though.

kevin-bates:
@jcrist - you're right.  Restoring `WaitingForSparkSessionToBeInitialized` to top-level still works.  Since that feels more correct (I find it odd to define a class in a method's if-block), I'd rather keep it at the top.

If you could mark your review as approved, that would be great. 

Thank you for all your help on this!

PR #695, opened 2019-06-24
Add helm chart tarball to distribution

### LINKED ISSUE #656: Investigate how to integrate release process to handle Helm charts
When releasing EG, what artifacts needs to be produced (and made available into dist/ and later on into git release page)to point helm install to it.

kevin-bates:
Added more (hopefully useful) metadata to the Chart.yaml file that
includes making the version of the chart match that of the EG release.
Modified the release.sh to update the version information in the chart.
Added `helm-chart` target to Makefile and tied to `dist` target.
Updated docs to reference helm chart tarball on release page.

Fixes #656

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@
     clean-kernel-py clean-kernel-spark-py clean-kernel-r clean-kernel-spark-r clean-kernel-scala clean-kernel-tf-py \
     clean-kernel-tf-gpu-py clean-kernel-image-puller push-images push-enterprise-gateway-demo push-nb2kg push-demo-base \
     push-kernel-images push-enterprise-gateway push-kernel-py push-kernel-spark-py push-kernel-r push-kernel-spark-r \
-    push-kernel-scala push-kernel-tf-py push-kernel-tf-gpu-py push-kernel-image-puller publish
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -88,7 +91,13 @@ sdist:
 	$(SA) $(ENV) && python setup.py sdist $(POST_SDIST) \
 		&& rm -rf *.egg-info
 
-dist: lint bdist sdist kernelspecs ## Make source, binary and kernelspecs distribution to dist folder
+helm-chart:
```
esevan: Isn't this command available to run alone? If it is, adding description will be helpful to print it out as a result of `make help`

Makefile
```diff
@@ -88,7 +91,13 @@ sdist:
...
+	make $(HELM_CHART)
+
+$(HELM_CHART): $(HELM_CHART_FILES)
+	(mkdir -p dist; cd etc/kubernetes/helm; tar -cvzf ../../../$(HELM_CHART) enterprise-gateway)
+
+dist: lint bdist sdist kernelspecs helm-chart ## Make source, binary and kernelspecs distribution to dist folder
```
esevan: I think description also be modified to
```
## Make source, binary, kernelspecs distribution, and helm chart to dist folder
```

kevin-bates:
Thanks @esevan - the updates have been pushed.

PR #698, opened 2019-06-27
Zmq port setting

IMAM9AIS:
-- Adding option to set ws_ping_interval variable for notebook server usages.

### DIFF
--- a/docs/source/config-options.md
+++ b/docs/source/config-options.md
@@ -249,11 +249,11 @@ EnterpriseGatewayApp options
     Default: False
     Is YARN Kerberos/SPNEGO Security enabled (True/False).
     (EG_YARN_ENDPOINT_SECURITY_ENABLED env var)
---EnterpriseGatewayApp.zmq_port_ping_interval=<Int>
-    Default: 30000
+--EnterpriseGatewayApp.ws_ping_interval=<Int>
+    Default: 30
     Specifies the value of ws_ping_interval that is being used for websocket
-    ping pong mechanism in ZMQ Port Handler from notebook server.
-    (NOTEBOOK_ZMQ_PORT_PING_INTERVAL env var) 
+    ping pong mechanism in ZMQ Port Handler from notebook server in seconds.
+    (EG_WS_PING_INTERVAL_SECS env var) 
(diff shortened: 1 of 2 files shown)


docs/source/config-options.md
```diff
@@ -249,6 +249,11 @@ EnterpriseGatewayApp options
     Default: False
     Is YARN Kerberos/SPNEGO Security enabled (True/False).
     (EG_YARN_ENDPOINT_SECURITY_ENABLED env var)
+--EnterpriseGatewayApp.zmq_port_ping_interval=<Int>
+    Default: 30000
+    Specifies the value of ws_ping_interval that is being used for websocket
+    ping pong mechanism in ZMQ Port Handler from notebook server.
+    (NOTEBOOK_ZMQ_PORT_PING_INTERVAL env var)
```
kevin-bates: I'd prefer we follow the naming (and units) convention that @esevan introduced in https://github.com/jupyter/nb2kg/pull/29.  Let's call the option `ws_ping_interval` (with the help string indicating seconds) and the env variable `EG_WS_PING_INTERVAL_SECS` (defaulting to 30), then expand to `ms` on usage.

I don't think we need to mention ZMQ since that corresponds more closely to the (multiple) ZMQ ports - we're really pinging the (single) WS port.

enterprise_gateway/enterprisegatewayapp.py
```diff
@@ -170,6 +170,17 @@ def port_range_default(self):
...
+    zmq_port_ping_interval_default_value = 30000
+    zmq_port_ping_interval = Integer(zmq_port_ping_interval_default_value, config=True,
+                                     help="""Specifies the ping interval that should be used by zmq port associated with
+                                    spawned kernels.Set this variable to 0 to disable ping mechanism.
+                                    (NOTEBOOK_ZMQ_PORT_PING_INTERVAL env var)""")
+
+    @default('zmq_port_ping_interval')
+    def zmq_port_ping_interval_default(self):
+        return int(os.getenv(self.zmq_port_ping_interval_env, self.zmq_port_ping_interval_default_value))
+
```
kevin-bates: See previous regarding naming and units.

PR #699, opened 2019-07-03
Kernel Startup using Yarn with resource and request check

IMAM9AIS:
Raising an Initial PR for this issue:- 
https://github.com/jupyter/enterprise_gateway/issues/664
There is a seperate PR to be raised on yarn-api-client repo. We can have a discussion on this PR before moving on another one.

### DIFF
--- a/enterprise_gateway/services/processproxies/yarn.py
+++ b/enterprise_gateway/services/processproxies/yarn.py
@@ -159,13 +159,15 @@ class YarnClusterProcessProxy(RemoteProcessProxy):
         self.candidate_queue = self.resource_mgr.cluster_scheduler_queue(candidate_queue_name)
 
         if self.candidate_queue is None:
-            self.log.debug("Queue: {} does not exist in the cluster".format(candidate_queue_name))
+            self.log.warning("Queue: {} not found in cluster."
+                             "Availability check will not be performed".format(candidate_queue_name))
             return
 
         self.candidate_partition = self.resource_mgr.cluster_queue_partition(self.candidate_queue, node_label)
 
         if self.candidate_partition is None:
-            self.log.debug("Parition: {} not found in {} queue".format(node_label, candidate_queue_name))
+            self.log.debug("Partition: {} not found in {} queue."
+                           "Availability check will not be performed".format(node_label, candidate_queue_name))
             return
 
         self.log.debug("Checking endpoint: {} if partition: {} "
--- a/enterprise_gateway/services/processproxies/yarn.py
+++ b/enterprise_gateway/services/processproxies/yarn.py
@@ -79,12 +79,15 @@ class YarnClusterProcessProxy(RemoteProcessProxy):
             self.log.debug("{class_name} shutdown wait time adjusted to {wait_time} seconds.".
                            format(class_name=type(self).__name__, wait_time=kernel_manager.shutdown_wait_time))
 
+        # If yarn resource check is enabled and it isn't available immediately,
+        # 20% of kernel_launch_timeout is used to wait
+        # and retry at fixed interval before pronouncing as not feasible to launch.
         self.yarn_resource_check_wait_time = 0.20 * self.kernel_launch_timeout
 
     def launch_process(self, kernel_cmd, **kwargs):
         # checks to see if the queue resource is available
         # if not kernel startup is not tried
-        self.confirm_yarn_queue_avaibility(**kwargs)
+        self.confirm_yarn_queue_availability(**kwargs)
         """Launches the specified process within a YARN cluster environment."""
         super(YarnClusterProcessProxy, self).launch_process(kernel_cmd, **kwargs)
 
@@ -108,49 +111,67 @@ class YarnClusterProcessProxy(RemoteProcessProxy):
 
         """
 
-    def confirm_yarn_queue_avaibility(self, **kwargs):
+    def confirm_yarn_queue_availability(self, **kwargs):
         """
         confirms if the yarn queue has capacity to handle the resource requests that
         will be sent to it.
-        if not, proper error messages are sent back for user experience
-        :param env_dict:
+
+        First check ensures the driver and executor memory request falls within
+        the container size of yarn configuration. This check requires executor and
+        driver memory to be available in the env.
+
+        Second,Current version of check, takes into consideration node label partitioning
+        on given queues. Provided the queue name and node label this checks if
+        the given partition has capacity available for kernel startup.
+
+        All Checks are optional. If we have KERNEL_EXECUTOR_MEMORY and KERNEL_DRIVER_MEMORY
+        specified, first check is performed.
+
+        If we have KERNEL_QUEUE and KERNEL_NODE_LABEL specified, second check is performed.
+
+        Proper error messages are sent back for user experience
+        :param kwargs:
         :return:
         """
+        env_dict = kwargs.get('env', {})
 
-        executer_memory = int(kwargs['env']['KERNEL_EXECUTOR_MEMORY'])
-        driver_memory = int(kwargs['env']['KERNEL_DRIVER_MEMORY'])
+        executor_memory = int(env_dict.get('KERNEL_EXECUTOR_MEMORY', 0))
+        driver_memory = int(env_dict.get('KERNEL_DRIVER_MEMORY', 0))
 
-        candidate_queue_name = str(kwargs['env']['KERNEL_QUEUE'])
+        if executor_memory*driver_memory > 0:
+            container_memory = self.resource_mgr.cluster_node_container_memory()
+            if max(executor_memory, driver_memory) > container_memory:
+                self.log_and_raise(http_status_code=500,
+                                   reason="Container Memory not sufficient for a executor/driver allocation")
 
-        node_label = str(kwargs['env']['KERNEL_NODE_LABEL'])
+        candidate_queue_name = (env_dict.get('KERNEL_QUEUE', None))
+        node_label = env_dict.get('KERNEL_NODE_LABEL', None)
 
-        self.container_memory = self.resource_mgr.cluster_node_container_memory()
-        self.candidate_queue = self.resource_mgr.cluster_scheduler_queue(candidate_queue_name)
-        self.candidate_partition = self.resource_mgr.cluster_queue_partition(self.candidate_queue, node_label)
-
-        if max(executer_memory, driver_memory) > self.container_memory:
-            self.log_and_raise(http_status_code=500,
-                               reason="Container Memory not sufficient for a executor/driver allocation")
+        if candidate_queue_name is None or node_label is None:
+            return
 
         # else the resources may or may not be available now. it may be possible that if we wait then the resources
         # become available. start  a timeout process
 
         self.start_time = RemoteProcessProxy.get_current_time()
+        self.candidate_queue = self.resource_mgr.cluster_scheduler_queue(candidate_queue_name)
+        self.candidate_partition = self.resource_mgr.cluster_queue_partition(self.candidate_queue, node_label)
+
         yarn_available = False
 
-        self.log.info("Yarn Endpoint to check for resources : " + self.yarn_endpoint)
-        self.log.info("Retrying for " + str(self.yarn_resource_check_wait_time) + " ms if resources not found ")
+        self.log.debug("Yarn Endpoint to check for resources : " + self.yarn_endpoint)
+        self.log.debug("Retrying for " + str(self.yarn_resource_check_wait_time) + " ms if resources not found ")
         while not yarn_available:
             self.handle_yarn_queue_timeout()
+            self.log.debug("Checking if " + self.candidate_partition + " has used capacity < 95% ")
             yarn_available = self.resource_mgr.cluster_scheduler_queue_availability(self.candidate_partition)
+            self.log.debug("Result: " + str(yarn_available))
+            # each time we sleep we substract equal amount of time for the kernel launch timeout
+            self.kernel_launch_timeout -= poll_interval
 
     def handle_yarn_queue_timeout(self):
 
-        # each time we sleep we substract equal amount of time for the kernel launch timeout
-        self.log.debug("Kernel Launch Timeout - {}".format(self.kernel_launch_timeout))
         time.sleep(poll_interval)
-        self.kernel_launch_timeout -= poll_interval
-
         time_interval = RemoteProcessProxy.get_time_diff(self.start_time, RemoteProcessProxy.get_current_time())
 
         if time_interval > self.yarn_resource_check_wait_time:
--- a/enterprise_gateway/services/processproxies/yarn.py
+++ b/enterprise_gateway/services/processproxies/yarn.py
@@ -102,14 +102,12 @@ class YarnClusterProcessProxy(RemoteProcessProxy):
 
         return self
 
-    """Submitting jobs to yarn queue and then checking till the jobs are in running state 
-           will lead to orphan jobs being created in some scenarios.
-           We take kernel_launch_timeout time and divide this into two parts.
-           if the queue is unavailable we take max 20% of the time to poll the queue periodically
-           and if the queue becomes available the rest of timeout is met in 80% of the remmaining 
-           time.  
-
-        """
+    """Submitting jobs to yarn queue and then checking till the jobs are in running state
+        will lead to orphan jobs being created in some scenarios.
+        We take kernel_launch_timeout time and divide this into two parts.
+        if the queue is unavailable we take max 20% of the time to poll the queue periodically
+        and if the queue becomes available the rest of timeout is met in 80% of the remmaining
+        time."""
 
     def confirm_yarn_queue_availability(self, **kwargs):
         """
@@ -159,11 +157,20 @@ class YarnClusterProcessProxy(RemoteProcessProxy):
 
         self.start_time = RemoteProcessProxy.get_current_time()
         self.candidate_queue = self.resource_mgr.cluster_scheduler_queue(candidate_queue_name)
+
+        if self.candidate_queue is None:
+            self.log.debug("Queue: {} does not exist in the cluster".format(candidate_queue_name))
+            return
+
         self.candidate_partition = self.resource_mgr.cluster_queue_partition(self.candidate_queue, node_label)
 
-        self.log.debug("Checking endpoint: {} if partition: {} has used capacity <= {}%".format(self.yarn_endpoint,
-                                                                                                self.candidate_partition,
-                                                                                                    partition_availability_threshold))
+        if self.candidate_partition is None:
+            self.log.debug("Parition: {} not found in {} queue".format(node_label, candidate_queue_name))
+            return
+
+        self.log.debug("Checking endpoint: {} if partition: {} "
+                       "has used capacity <= {}%".format(self.yarn_endpoint,
+                                                         self.candidate_partition, partition_availability_threshold))
 
         yarn_available = self.resource_mgr.cluster_scheduler_queue_availability(self.candidate_partition,
                                                                                 partition_availability_threshold)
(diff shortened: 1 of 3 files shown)


enterprise_gateway/services/processproxies/yarn.py
```diff
@@ -79,7 +79,12 @@ def __init__(self, kernel_manager, proxy_config):
             self.log.debug("{class_name} shutdown wait time adjusted to {wait_time} seconds.".
                            format(class_name=type(self).__name__, wait_time=kernel_manager.shutdown_wait_time))
 
+        self.yarn_resource_check_wait_time = 0.20 * self.kernel_launch_timeout
```
kevin-bates: Let's add a comment stating we're going to take up 20% of the kernel launch timeout to check on resource availability.

enterprise_gateway/services/processproxies/yarn.py
```diff
@@ -79,7 +79,12 @@ def __init__(self, kernel_manager, proxy_config):
...
+
     def launch_process(self, kernel_cmd, **kwargs):
+        # checks to see if the queue resource is available
+        # if not kernel startup is not tried
+        self.confirm_yarn_queue_avaibility(**kwargs)
```
kevin-bates: Typo in method name.

enterprise_gateway/services/processproxies/yarn.py
```diff
@@ -94,6 +99,65 @@ def launch_process(self, kernel_cmd, **kwargs):
...
+        :param env_dict:
+        :return:
+        """
+
+        executer_memory = int(kwargs['env']['KERNEL_EXECUTOR_MEMORY'])
+        driver_memory = int(kwargs['env']['KERNEL_DRIVER_MEMORY'])
+
+        candidate_queue_name = str(kwargs['env']['KERNEL_QUEUE'])
+
+        node_label = str(kwargs['env']['KERNEL_NODE_LABEL'])
```
kevin-bates: We shouldn't require these values, nor do I think we can reliably come up with default values for these.  We should pick one of these that is required to enable this functionality.  I'm thinking `KERNEL_QUEUE`.  If that is provided, then the others must also be present for this logic to proceed.  If `KERNEL_QUEUE` is not provided, this logic is silently skipped.  However, if `KERNEL_QUEUE` is provided, but __any__ of the others are not, let's log a warning message indicating the missing others, but NOT raise an exception - just exit the method.

Probably better to use `**kwargs.get('env')` to avoid `KeyError` exceptions.

enterprise_gateway/services/processproxies/yarn.py
```diff
@@ -94,6 +99,65 @@ def launch_process(self, kernel_cmd, **kwargs):
...
+        self.candidate_partition = self.resource_mgr.cluster_queue_partition(self.candidate_queue, node_label)
+
+        if max(executer_memory, driver_memory) > self.container_memory:
+            self.log_and_raise(http_status_code=500,
+                               reason="Container Memory not sufficient for a executor/driver allocation")
+
+        # else the resources may or may not be available now. it may be possible that if we wait then the resources
+        # become available. start  a timeout process
+
+        self.start_time = RemoteProcessProxy.get_current_time()
```
kevin-bates: We should start the timer prior to the yarn-api calls.

enterprise_gateway/services/processproxies/yarn.py
```diff
@@ -94,6 +99,65 @@ def launch_process(self, kernel_cmd, **kwargs):
...
+        yarn_available = False
+
+        self.log.info("Yarn Endpoint to check for resources : " + self.yarn_endpoint)
+        self.log.info("Retrying for " + str(self.yarn_resource_check_wait_time) + " ms if resources not found ")
```
kevin-bates: Can we change these to DEBUG messages.

enterprise_gateway/services/processproxies/yarn.py
```diff
@@ -94,6 +99,65 @@ def launch_process(self, kernel_cmd, **kwargs):
...
+        while not yarn_available:
+            self.handle_yarn_queue_timeout()
+            yarn_available = self.resource_mgr.cluster_scheduler_queue_availability(self.candidate_partition)
+
+    def handle_yarn_queue_timeout(self):
+
+        # each time we sleep we substract equal amount of time for the kernel launch timeout
+        self.log.debug("Kernel Launch Timeout - {}".format(self.kernel_launch_timeout))
+        time.sleep(poll_interval)
+        self.kernel_launch_timeout -= poll_interval
```
kevin-bates: Rather than update on each iteration, can we decrement the `self.kernel_launch_timeout` whatever the total duration was back in the calling method?  We either get an exception or succeed, so it can be the last statement in that method.

enterprise_gateway/services/processproxies/yarn.py
```diff
@@ -94,6 +99,65 @@ def launch_process(self, kernel_cmd, **kwargs):
...
+        self.log.info("Yarn Endpoint to check for resources : " + self.yarn_endpoint)
+        self.log.info("Retrying for " + str(self.yarn_resource_check_wait_time) + " ms if resources not found ")
+        while not yarn_available:
+            self.handle_yarn_queue_timeout()
+            yarn_available = self.resource_mgr.cluster_scheduler_queue_availability(self.candidate_partition)
+
+    def handle_yarn_queue_timeout(self):
+
+        # each time we sleep we substract equal amount of time for the kernel launch timeout
+        self.log.debug("Kernel Launch Timeout - {}".format(self.kernel_launch_timeout))
```
kevin-bates: This debug isn't very useful.  I'd rather we add a debug statement following the call to `cluster_scheduler_queue_availability` that includes some of the details we're checking for, along with the result.

enterprise_gateway/services/processproxies/yarn.py
IMAM9AIS: Hi Kevin,

It's good we are discussing on this part of the code which needs to be generic and is most important.

To give you a background, we are internally using a capacity scheduler and the queue that we specify is actually split into two partitions internally. Using the value of node_label helps us identify which queue partition we should be looking at to query it's absolute used capacity.

Now if we are to generalize this, should we just look at the queue name and query it's used capacity on parent level ? Or do you suggest some other route ?

IMAM9AIS: Also from what i understand, in scenarios were we are using node labels to schedule the jobs, 
mentioning a particular node label only schedule jobs on nodes having that label, while if you don't mention any node label while submitting jobs, it will scheduled in the default partition.

In that scenario, assuming we have a queue name specified with/without node labels, we can look in the default/actual partition respectively inside that queue. 

But will this logic be generic enough ? I don't think people use a node partitioning to submit jobs in every scenario or do they ?

IMAM9AIS: Actually on further thoughts, there are various cases we will need to cover within capacity scheduling itself,

**Case 1 - Where user specifies queue name & node label**
This means the resources need to be requested exclusively on these node labels so we will have to query this node label for it's used capacity which is already implemented above.

**Case 2 - Where user specifies queue name without a node label**
According to yarn concepts, this will allow yarn to schedule job on nodes that are default as well as on nodes that are non-exclusive on which queue has access. This would require us to actually query all the partitions that are associated with this queue, then query this partition to understand if it is non-exclusive and then add the results up to determine a yes or no. (Work required)

**Case 3 - No node label concept is used.**
Have not dived into this, but this should be simple, we can just look into the used capacity of the queue itself.

**Case 4 -  It's not a capacity scheduler (some other scheduler )**
Open question

kevin-bates: Thanks for the information.  I am unfamiliar with YARN queueing.  Adding @lresende.

However, I think if we let the existence of `KERNEL_QUEUE_NAME` drive this, we might be okay (for these cases).  If `KERNEL_QUEUE_NAME` does not exist, I don't think we have any choice but to skip this code and proceed as we do today (which I assume is __Case 4__ for now).

How is __Case 3__ different from __Case 2__? In Case 3 you say _look into the used capacity of the queue itself_ implying that `KERNEL_QUEUE_NAME` was specified (unless there's some default queue that is implied by this).

If we require BOTH queue name and node label to enable this logic, then we address cases 1 and 4 (correct?).

I'm not sure how important Case 2 and 3 are.  Hoping others more familiar with YARN can chime in here.

IMAM9AIS: @kevin-bates I think this page can help us understand scheduling in YARN in a better way.

https://developer.ibm.com/hadoop/2017/03/10/yarn-node-labels/

Since we were using node labels, the PR actually tries to derive the remaining queue capacities given we have exclusive partitions(refer link). This covers Case 1. Case 2 is something in which the no nodel label is specified but the system still has node label partitioning.

Case 3 - which i have not researched as of now, is something where we have not set up yarn to have exlusive/non-exclusive partitions. It might be possible in this scenario, the solution is similar to Case 2 but we will have to dive a little.

kevin-bates: @IMAM9AIS - thank you for the link.

I don't think we should get caught up in how to cover all the cases at this point.  Let's identify either the `KERNEL_QUEUE_NAME` or `KERNEL_NODE_LABEL` (or both) as the "drivers" that enable this logic and get your primary case covered. If you can knock out others - great.  Other users that might have different use cases can adjust this method accordingly.  (Please add a link to the article in a nearby comment.)

We just need to ensure that not setting these values results in today's (default) behavior.

IMAM9AIS: @kevin-bates Agreed.

From what we understand, we will have to use both (queue name and node label as drivers ) for this logic. I will be fixing all comments and committing these changes soon (along with default behaviour logic) .

enterprise_gateway/services/processproxies/yarn.py
IMAM9AIS: Added

enterprise_gateway/services/processproxies/yarn.py
IMAM9AIS: Changed

enterprise_gateway/services/processproxies/yarn.py
IMAM9AIS: I have added the changes. Please have a look @kevin-bates .

enterprise_gateway/services/processproxies/yarn.py
```diff
@@ -108,49 +111,67 @@ def launch_process(self, kernel_cmd, **kwargs):
 
         """
 
-    def confirm_yarn_queue_avaibility(self, **kwargs):
+    def confirm_yarn_queue_availability(self, **kwargs):
         """
         confirms if the yarn queue has capacity to handle the resource requests that
```
kevin-bates: Please capitalize 'confirms' -> 'Confirms'.

I think we should also add a statement near the beginning that this algorithm is subject to change as use cases become more defined.  That then would allow the removal of "Current version of check," from the second bullet below.

enterprise_gateway/services/processproxies/yarn.py
```diff
@@ -108,49 +111,67 @@ def launch_process(self, kernel_cmd, **kwargs):
...
-        self.log.info("Retrying for " + str(self.yarn_resource_check_wait_time) + " ms if resources not found ")
+        self.log.debug("Yarn Endpoint to check for resources : " + self.yarn_endpoint)
+        self.log.debug("Retrying for " + str(self.yarn_resource_check_wait_time) + " ms if resources not found ")
         while not yarn_available:
             self.handle_yarn_queue_timeout()
+            self.log.debug("Checking if " + self.candidate_partition + " has used capacity < 95% ")
             yarn_available = self.resource_mgr.cluster_scheduler_queue_availability(self.candidate_partition)
+            self.log.debug("Result: " + str(yarn_available))
+            # each time we sleep we substract equal amount of time for the kernel launch timeout
+            self.kernel_launch_timeout -= poll_interval
```
kevin-bates: Let's remove any debug statements from within the loop as these will be posted every half second and nothing is changing.  We can move the first statement before the loop and the fact that we didn't throw an exception implies the result was successful.  (Btw, you might want to keep this in mind when looking into the yarn-api-client implementations since some will be called frequently in loops like this.)

Where does the 'capacity < 95%' come from?  That sounds like a detail of checking the availability and I'm not sure we should be assuming the 95% threshold.  Perhaps we can make that more generic?

My previous comment regarding decrementing `self.kernel_launch_timeout` was to do something like this...
```python
while not yarn_available:
            self.handle_yarn_queue_timeout()
            yarn_available = self.resource_mgr.cluster_scheduler_queue_availability(self.candidate_partition)

check_availability_total_time = RemoteProcessProxy.get_time_diff(self.start_time, RemoteProcessProxy.get_current_time())
self.kernel_launch_timeout -= check_availability_total_time
```
If you'd rather decrement on each iteration, then please move it back into the timeout handler.  I think it would be good to isolate the use of `poll_interval` to as few places as possible.

IMAM9AIS: > Where does the 'capacity < 95%' come from? That sounds like a detail of checking the availability and I'm not sure we should be assuming the 95% threshold. Perhaps we can make that more generic?

Sure.
A thing to note here is, right now, we have not implemented the logic of checking the memory requests during kernel launch with the amount of memory available in yarn queue. 
Rather we are checking if the queue is more than 95% full. If yes, we don't allow the kernel to be launched. This logic appeared more generic to us, given that we can actually tune this threshold according to yarn configurations and admin requirements to allow for kernel launch. 
And yes, we should be making this check more generic enough. 
I Will be committing files with relevant changes.

IMAM9AIS: Uploaded.

enterprise_gateway/services/processproxies/yarn.py
```diff
@@ -94,6 +102,89 @@ def launch_process(self, kernel_cmd, **kwargs):
...
+
+        if executor_memory * driver_memory > 0:
+            container_memory = self.resource_mgr.cluster_node_container_memory()
+            if max(executor_memory, driver_memory) > container_memory:
+                self.log_and_raise(http_status_code=500,
+                                   reason="Container Memory not sufficient for a executor/driver allocation")
+
+        candidate_queue_name = (env_dict.get('KERNEL_QUEUE', None))
+        node_label = env_dict.get('KERNEL_NODE_LABEL', None)
+        partition_availability_threshold = float(env_dict.get('PARTITION_THRESHOLD',95.0))
```
kevin-bates: Please prefix partition_threshold env name with `YARN_`

enterprise_gateway/services/processproxies/yarn.py
```diff
@@ -94,6 +102,89 @@ def launch_process(self, kernel_cmd, **kwargs):
...
+        self.start_time = RemoteProcessProxy.get_current_time()
+        self.candidate_queue = self.resource_mgr.cluster_scheduler_queue(candidate_queue_name)
+        self.candidate_partition = self.resource_mgr.cluster_queue_partition(self.candidate_queue, node_label)
+
+        yarn_available = False
+        retry_attempt = 0
+
+        self.log.debug("Yarn Endpoint to check for resources : " + self.yarn_endpoint)
+        self.log.debug("Checking if " + self.candidate_partition + " has used capacity <= " + str(partition_availability_threshold) + "%" )
+        self.log.debug("Retrying for " + str(self.yarn_resource_check_wait_time) + " ms if resources not found ")
```
kevin-bates: Can we please combine the yarn_endpoint statement into the 'Checking' statement?
`self.log.debug("Checking {endpoint} if {partition} has used capacity...`

Also, please use the string formatting used elsewhere using `format()` rather sting concatenation.

enterprise_gateway/services/processproxies/yarn.py
```diff
@@ -94,6 +102,89 @@ def launch_process(self, kernel_cmd, **kwargs):
...
+        while not yarn_available:
+            self.handle_yarn_queue_timeout()
+            yarn_available = self.resource_mgr.cluster_scheduler_queue_availability(self.candidate_partition,partition_availability_threshold)
```
kevin-bates: We should either copy this statement outside the loop so that we don't incur unnecessary sleep calls or switch the two and make the handle-timeout method call conditional.  I prefer the former.  It would be good to produce the 'Retrying for' debug statement ONLY if yarn_available is initially false after the first call.

enterprise_gateway/services/processproxies/yarn.py
```diff
@@ -94,6 +102,89 @@ def launch_process(self, kernel_cmd, **kwargs):
...
+            retry_attempt+=1
+
+        #subtracting the total amount of time spent for polling for queue availability
+        self.kernel_launch_timeout -= (poll_interval*retry_attempt)
```
kevin-bates: Please update this statement to:
`self.kernel_launch_timeout -= RemoteProcessProxy.get_time_diff(self.start_time, RemoteProcessProxy.get_current_time())`

and remove the use of `retry_attempt`

I will be updating `RemoteProcessProxy.get_time_diff()` to default the second parameter to `get_current_time()` so the statement will then become `self.kernel_launch_timeout -= RemoteProcessProxy.get_time_diff(self.start_time)`

enterprise_gateway/services/processproxies/yarn.py
```diff
@@ -94,6 +102,89 @@ def launch_process(self, kernel_cmd, **kwargs):
...
+
+    def handle_yarn_queue_timeout(self):
+
+        time.sleep(poll_interval)
+        time_interval = RemoteProcessProxy.get_time_diff(self.start_time, RemoteProcessProxy.get_current_time())
+
+        if time_interval > self.yarn_resource_check_wait_time:
+            error_http_code = 500
+            reason = "Yarn Compute Resource is unavailable"
```
kevin-bates: Let's add `yarn_resource_check_wait_time` to the reason (`Yarn compute resource is unavailable after {} seconds`). This will give users the correct impression that that timeout value can be configured.

IMAM9AIS:
@kevin-bates https://github.com/jupyter/enterprise_gateway/pull/699/commits/2458d158e0bfc892238eb59f8ad4f64786cf745d

kevin-bates:
I should have checked the build issue.  We'll also need to get rid of lint errors.  Please make sure you can run `make lint` prior to updating the PR next time.

enterprise_gateway/services/processproxies/yarn.py
```diff
@@ -159,11 +157,20 @@ def confirm_yarn_queue_availability(self, **kwargs):
 
         self.start_time = RemoteProcessProxy.get_current_time()
         self.candidate_queue = self.resource_mgr.cluster_scheduler_queue(candidate_queue_name)
+
+        if self.candidate_queue is None:
+            self.log.debug("Queue: {} does not exist in the cluster".format(candidate_queue_name))
```
kevin-bates: Since this is requested functionality, by virtue of a non-null `KERNEL_QUEUE` value, I think we should log this as a warning and add " - availability check will not be performed."

enterprise_gateway/services/processproxies/yarn.py
```diff
@@ -159,11 +157,20 @@ def confirm_yarn_queue_availability(self, **kwargs):
...
+            return
+
         self.candidate_partition = self.resource_mgr.cluster_queue_partition(self.candidate_queue, node_label)
 
-        self.log.debug("Checking endpoint: {} if partition: {} has used capacity <= {}%".format(self.yarn_endpoint,
-                                                                                                self.candidate_partition,
-                                                                                                    partition_availability_threshold))
+        if self.candidate_partition is None:
+            self.log.debug("Parition: {} not found in {} queue".format(node_label, candidate_queue_name))
```
kevin-bates: See previous comment - log as warning, add suffix statement.

kevin-bates: Need to fix typo in 'Partition' as well.

IMAM9AIS: Commited the changes.

enterprise_gateway/services/processproxies/yarn.py
IMAM9AIS: Commited the changes.

IMAM9AIS:
@kevin-bates  please have a look.

kevin-bates:
The tests can be somewhat flakey.  I restarted that job and suspect things will be fine.

Let's restore the dependency cap and we're good.  Thanks.

IMAM9AIS:
@kevin-bates Please have a look.

IMAM9AIS:
@kevin-bates done.

PR #709, opened 2019-07-16
termination of EG in SIGTERM

### LINKED ISSUE #701: EG is not terminating correctly via SIGTERM
I want to add functionality of handling "SIGKILL" signal in jupyter enterprise gateway. I am thinking of cleaning up and then killing the current process in "SIGKILL" signal. 

I wanted to ask whether JEG creates any child processes in any scenarios so I need to kill them as well.

### LINKED ISSUE #710: Backporting SIGTERM issue in 1.x version
Hi Kevin,
As we discussed before, I have created a pull request for backporting "EG not terrminating in SIGTERM" in 1.x version.

Previous Issue - https://github.com/jupyter/enterprise_gateway/issues/701

Can you please look into this?

Thanks
Vikas

vikasgarg1996:
In Current code, EG is not terminating gateway server process when handling SIGTERM signal. It is killing any running kernels correctly but the main process is not being killed.

In this fix, sending SIGTERM will kill any running kernels and will shutdown the gateway server as well.

(Issue resolution backporting in version 1.x)

Fixes #710 
Closes #710

### DIFF
--- a/enterprise_gateway/enterprisegatewayapp.py
+++ b/enterprise_gateway/enterprisegatewayapp.py
@@ -299,6 +299,6 @@ class EnterpriseGatewayApp(KernelGatewayApp):
 
     def _signal_stop(self, sig, frame):
         self.log.info("Received signal to terminate Enterprise Gateway.")
-        self.io_loop.stop()
+        self.io_loop.add_callback_from_signal(self.io_loop.stop)
 
 launch_instance = EnterpriseGatewayApp.launch_instance


kevin-bates:
Thanks for the backport.  It looks like we have a build issue in our 1.x line that's going to require changing from oracle to open jdk (which we've already done in 2.x).  I'm going to go ahead and merge this change since its unrelated.

PR #713, opened 2019-07-30
Converted connection_info key to string for json dump after loading a…

### LINKED ISSUE #711: Persistent Kernel loading failing in v1.x
Hi Kevin,

I was testing Persistent Kernels in JEG code of v1.x and python version 3.7
There is an issue I am facing after startup of persistent kernel was successful.

So when the file (kernel.json) which stores the information about running kernels gets loaded, JEG uses `json.load(fp)` where `fp` is file pointer to the file. In the loaded JSON, for every kernel session, `kernel_session["connection_info"]["key"]` will be a string variable. 

While starting the session we change this variable to bytes by using `str_to_bytes` from `ipython_genutils.py3compat`. So startup of this kernel gets successful. 

In the end, we are trying to commit sessions by saving them again to the same file using `json.dump`. But the issue comes because `kernel_session["connection_info"]["key"]` is still a bytes variable and `json.dump` fails. 

In v2.0.0rc2, it is handled using a static method `pre_save_transformation`. 
Can we do something similar here?
Screenshot 2019-07-19 at 7 15 58 PM
Screenshot 2019-07-19 at 7 16 15 PM
Screenshot 2019-07-19 at 7 16 33 PM
Screenshot 2019-07-19 at 7 16 42 PM

## Environment

- Enterprise Gateway Version 1.x
- Platform: YARN
- Python 3.7

vikasgarg1996:
In Current Code of 1.x, Persistent kernels have a small bug, where it is changing kernel's connection_info key to ***bytes*** for loading the kernel but it's not converting back it to ***String*** for json dumping and tht's why it is failing.
I have added corresponding code for that. Please look into this -

Closing issues -
Closes #711

### DIFF
--- a/enterprise_gateway/services/kernels/remotemanager.py
+++ b/enterprise_gateway/services/kernels/remotemanager.py
@@ -52,6 +52,10 @@ class RemoteMappingKernelManager(SeedingMappingKernelManager):
         # Load connection info into member vars - no need to write out connection file
         km.load_connection_info(connection_info)
 
+        info_key = connection_info.get('key')
+        if info_key:
+            connection_info['key'] = bytes_to_str(info_key)
+
         km._launch_args = launch_args
 
         # Construct a process-proxy
--- a/enterprise_gateway/services/kernels/remotemanager.py
+++ b/enterprise_gateway/services/kernels/remotemanager.py
@@ -52,6 +52,7 @@ class RemoteMappingKernelManager(SeedingMappingKernelManager):
         # Load connection info into member vars - no need to write out connection file
         km.load_connection_info(connection_info)
 
+        # Converting `connection_info['key']` back to string so that json operations work properly
         info_key = connection_info.get('key')
         if info_key:
             connection_info['key'] = bytes_to_str(info_key)


enterprise_gateway/services/kernels/remotemanager.py
```diff
@@ -52,6 +52,10 @@ def start_kernel_from_session(self, kernel_id, kernel_name, connection_info, pro
         # Load connection info into member vars - no need to write out connection file
         km.load_connection_info(connection_info)
 
+        info_key = connection_info.get('key')
+        if info_key:
+            connection_info['key'] = bytes_to_str(info_key)
+
```
kevin-bates: This looks fine. I think we should add a comment as to why this is necessary since `load_connection_info()` is doing the exact opposite operation (converting the key to bytes). As a result, someone looking at this will find it odd. I think something like `Convert key back to string so json operations are properly behaved` may be sufficient.

kevin-bates: @esevan - do you mind giving this a look as well?  Thanks.

kevin-bates:
Please ignore the build issues - we're aware of what's going on and working on it.

enterprise_gateway/services/kernels/remotemanager.py
esevan: I'm the one confused by missing comment :confused:. Looks good to me except for the comment thing :+1:

vikasgarg1996:
Added the corresponding comment. Please check.

PR #734, opened 2019-09-13
Removed dependency on Kernel Gateway

### LINKED ISSUE #733: Remove dependency on Kernel Gateway
In order to move forward with some roadmap items and open issues that require alterations to kernel and kernelspec handlers, it makes sense for Enterprise Gateway to remove its dependency on Kernel Gateway.  At quick glance, this would essentially amount to copying the websocket, base, mixin handlers and some config traits from KG to EG - in addition to some doc updates.

kevin-bates:
This PR consists of the following changes to remove Enterprise Gateway's dependency on Kernel Gateway:

- [x] Copy necessary handlers from Kernel Gateway to EG
    - [x] Promote WebSocketPersonality code since NotebookHTTPPersonality will not be supported and we only have one personality (whew! :smile:)
- [x] Update all dependency files (setup, requirements, etc)
- [x] Rename `KG_` prefixed variables (env and code) to `EG_`.  For env variables we fallback to checking the "old" `KG_` form when the `EG_` form is not set - allowing existing configs to continue working.
- [x] Address lint issues that are not required in KG, but are in EG
- [x] Update/remove tests accordingly
- [x] Update references in docs (and README) - replace _"built on Kernel Gateway"_ with things like "_instead of Kernel Gateway"_.  Add explanation of when you'd chose one over the other.
- [x] Extend `/api` endpoint to return Gateway version info as well (only returns underlying notebook version)

Fixes #733


docs/source/getting-started.md
```diff
@@ -32,11 +32,25 @@ Jupyter Enterprise Gateway exposes the following features and functionality:
...
+1. You have a large computer cluster consisting of limited resources (GPUs, large memory, etc) and users require those resources from notebooks
+2. You have large numbers of users requiring access to a shared compute cluster
+3. You require some amount of High Availability/Disaster Recovery such that another Gateway server can be spun up to service existing (and remote) kernels
+
+#### Use Kernel Gateway if...
+1. You have a small user pool where the resources of the Gateway server can be shared amongst those users (no remote kernel support)
+2. You wish to configured the [notebook-http mode](https://jupyter-kernel-gateway.readthedocs.io/en/latest/http-mode.html) functionality where a specific Notebook provides HTTP endpoints
+
+Note that Enterprise Gateway also supports local kernels by default.  However, HA/DR functionality won't be affective unless kernels run remotely.
+
```
esevan: Sooooo clear! Thanks! :tada:

enterprise_gateway/enterprisegatewayapp.py
```diff
@@ -39,10 +66,195 @@ class EnterpriseGatewayApp(KernelGatewayApp):
...
+
+    list_kernels_env = 'EG_LIST_KERNELS'
+    list_kernels = Bool(config=True,
+                        help="""Permits listing of the running kernels using API endpoints /api/kernels
+                        and /api/sessions. (EG_LIST_KERNELS env var) Note: Jupyter Notebook
+                        allows this by default but Jupyter Enterprise Gateway does not.""")
+
+    @default('list_kernels')
+    def list_kernels_default(self):
+        return os.getenv(self.list_kernels_env, os.getenv('KG_LIST_KERNELS', 'False')) == 'True'
```
esevan: How about
```suggestion
        return os.getenv(self.list_kernels_env, os.getenv('KG_LIST_KERNELS', 'False')).lower() == 'true'
```

kevin-bates: Good catch! Suggestion pushed.

echarles:
@kevin-bates I understand this PR removes the ability for EG to support `notebook-http-mode` which is a big feature loose for me. I would love to use EG for both WebSocket and HTTP modes.

Any reason to have taken that decision? Would it make sense that I look to bring back that feature with a PR?

lresende:
@echarles We would prefer to have the `notebook-http-mode` as its own library either from JKG or a new one. Also, the JKG personalities are kind of mutually exclusive, and won't allow for distributed kernel management, which is sort of why we decided to go in this direction.

echarles:
@lresende Thx for the answer. A HTTP Personality which runs on a distributed cluster will be at some point needed. If EG does not offer it, I will look at other options for that. JKG is not a valid option to me as it can not be distributed.

kevin-bates:
@echarles - so you were using http-personality via EG to host a "remote web server"?  Which form of remote kernel were you using?

Sorry, but the http-personality is just not in the design center for EG as it requires the complete server just to host a kernel.

echarles:
@kevin-bates I am not using EG for http-personality for now but it sounded logical to me, in my quest of a scalable jupyter stack, to consider it.

Saying in another way, if I want to serve the users' kernels runtime via EG websocket-mod, all Websocket communication would go through the single EG Websocket server (which is also BTW a single HTTP Tornado Server). This means single server limitation which can be overcomed eg by running that single server on a platform like K8S.

 

The same logic applies exactly for http-mode. Say if you want to expose that cell to a user (eg javascript REST client), your endpoint server should be able to access your spark distributed cluster.

```python
# GET /api/filter
res = spark.load(...)
...
```

With this change, this is no more possible.

kevin-bates:
Yes, but in http-mode only ONE kernel can be served by the gateway server - so "scale" doesn't really apply.  If you're interested in k8s kernels, then just run JKG in k8s with its single local kernel.  Remote kernels really don't come into play when there can only be one.

Likewise for a Spark cluster.  JKG can launch a client-mode kernel (rather than cluster-mode) since there will only be one kernel - having it run in client mode isn't that big a deal - it still can access the spark distributed cluster.

echarles:
oh Should have done more tests to realize that only ONE kernel would be shared across all kenerl-gateway clients.

This leads to the need to support session based as discussed in https://github.com/jupyter/jupyter_server/issues/122

kevin-bates:
Hmm - I'm not sure I follow. In http-mode, the gateway process user is the notebook user in the sense that it is the thing actually issuing the kernel cell execution requests.  The actual user (from the client) is merely hitting different web endpoints, each of which correspond to a different cell to execute.  I suspect most notebooks used in this way would be completely stateless between cells so as to not side-affect other endpoint results.

echarles:
yeah maybe I expect too much from http-mode (it has been a long time I have used it). Let's leave it like that now and maybe I will try out something when I have a more clear idea about how it works with kernel gateway and the requirements I would like to have.

Issue #562, opened 2019-01-25
High Availability - Session persistence


Gsbreddy:
Hi, I was working on setting up EG in yarn cluster mode. It is working now but I have a question regarding the kernel session persistence in case of HA. In case of bare metal machines, to achieve HA, Lets say I am running 3 EG's on three different machines with a vip enabled. Then I will point the vip address to my hub to launch notebook kernels on those 3 machines. 

Through this setup I can ensure my EG's are highly available but my concern here is that I can't ensure the kernel persistency. According to the project road map, I can see work is being done on Session persistence. Any updates on this or how I can achieve HA on bare metal machines would really be helpful. Thanks!

kevin-bates:
Hi @Gsbreddy - thanks for the question/issue.  You must also have session affinity enabled - correct?

We have not started to address this road map item.  Presently, we have the ability to persist kernel session state to a file and load that file at EG startup and attempt to reconnect to the persisted kernel sessions.  So this would be able to address more of a _failover_ strategy, although it, itself, has some limitations:
1. It requires the file reside on some shared storage (NFS, etc.).
2. It is an Active/Passive strategy.  So if there are multiple active EG instances persisting sessions to the shared file and one node fails, there's no way of preventing the newly started node from trying to connect to kernels managed by the other successfully running nodes.
As a result, the current implementation is purely a POC to show that kernel connections can be re-established.

What I believe is needed is the following:
1. Swap out the file persistence methods with a NoSQL DB solution.  @lamOrigin007  sounds to be looking into a mongoDB approach based on his comments on [this PR](https://github.com/jupyter/enterprise_gateway/pull/530).
2. If we want active/active, then we'll need to indicate the _servicing EG instance_ in the persisted data and maintain some _active server_ state so that newly started EG instances can distinguish between currently maintained kernel sessions and those orphaned due to the death of their EG instance.
3. We should probably add logic to the kernel web handlers such that if a request comes in for an unknown kernel ID, we attempt to re-establish connections to that kernel using persisted state (if any).  However, since the web handlers are currently inherited from Kernel Gateway, we'd want to move the web handlers into Enterprise Gateway (and should discuss merging the projects).

I believe these issues are true for containerized environments as well.

lresende:
Please make sure there is a good abstraction for the type of persistence, and if choosing MongoDB, also make sure the license is compatible. Looking at it today I saw:
LICENSE

  MongoDB is free and open-source. Versions released prior to October 16,
[...]

PR #737, opened 2019-09-25
Updating FileSessionKernelManager to load&save kernel sessions to mul…

### LINKED ISSUE #735: Changing saving of kernels from single to multiple files.
This is a sub issue which updates FileSessionKernelManager to load and save kernel sessions to multiple files as discussed in https://github.com/jupyter/enterprise_gateway/issues/562

Gsbreddy:
…tiple files(#735)

This fixes #735 which updates FileSessionKernelManager to load and save kernel sessions to multiple files as discussed in #562.

### DIFF
--- a/enterprise_gateway/services/kernels/handlers.py
+++ b/enterprise_gateway/services/kernels/handlers.py
@@ -1,12 +1,13 @@
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
 """Tornado handlers for kernel CRUD and communication."""
-
+import json
 import os
 
 import tornado
 import notebook.services.kernels.handlers as notebook_handlers
-from tornado import gen
+from jupyter_client.jsonutil import date_default
+from tornado import gen, web
 from functools import partial
 from ...mixins import TokenAuthorizationMixin, CORSMixin, JSONErrorsMixin
 
@@ -102,9 +103,17 @@ class KernelHandler(TokenAuthorizationMixin,
     """Extends the notebook kernel handler with token auth, CORS, and
     JSON errors.
     """
+
     def options(self, **kwargs):
         """Method for properly handling CORS pre-flight"""
         self.finish()
+
+    @web.authenticated
+    def get(self, kernel_id):
+        km = self.kernel_manager
+        km._check_kernel_id(kernel_id)
+        model = km.kernel_model(kernel_id)
+        self.finish(json.dumps(model, default=date_default))
 
 
 default_handlers = []
--- a/enterprise_gateway/services/kernels/handlers.py
+++ b/enterprise_gateway/services/kernels/handlers.py
@@ -111,7 +111,7 @@ class KernelHandler(TokenAuthorizationMixin,
     @web.authenticated
     def get(self, kernel_id):
         km = self.kernel_manager
-        km._check_kernel_id(kernel_id)
+        km.check_kernel_id(kernel_id)
         model = km.kernel_model(kernel_id)
         self.finish(json.dumps(model, default=date_default))
 
--- a/enterprise_gateway/services/kernels/remotemanager.py
+++ b/enterprise_gateway/services/kernels/remotemanager.py
@@ -7,7 +7,7 @@ import signal
 import re
 import uuid
 
-from tornado import gen
+from tornado import gen, web
 from ipython_genutils.py3compat import unicode_type
 from ipython_genutils.importstring import import_item
 from notebook.services.kernels.kernelmanager import MappingKernelManager
@@ -46,8 +46,25 @@ def get_process_proxy_config(kernelspec):
 
 class RemoteMappingKernelManager(MappingKernelManager):
     """Extends the MappingKernelManager with support for managing remote kernels via the process-proxy. """
+
     def _kernel_manager_class_default(self):
         return 'enterprise_gateway.services.kernels.remotemanager.RemoteKernelManager'
+
+    def _check_kernel_id(self, kernel_id):
+        """Check that a kernel_id exists and raise 404 if not."""
+        if kernel_id not in self:
+            if not self._refresh_kernel(kernel_id):
+                self.parent.kernel_session_manager.delete_session(kernel_id)
+                raise web.HTTPError(404, u'Kernel does not exist: %s' % kernel_id)
+
+    def _refresh_kernel(self, kernel_id):
+        self.parent.kernel_session_manager.load_session(kernel_id)
+
+        if kernel_id in self.parent.kernel_session_manager._sessions:
+            return self.parent.kernel_session_manager._start_session(
+                self.parent.kernel_session_manager._sessions.get(kernel_id))
+
+        return False
 
     @gen.coroutine
     def start_kernel(self, *args, **kwargs):
@@ -141,7 +158,7 @@ class RemoteMappingKernelManager(MappingKernelManager):
         self.start_watching_activity(kernel_id)
         self.add_restart_callback(kernel_id,
                                   lambda: self._handle_kernel_died(kernel_id),
-                                  'dead',)
+                                  'dead', )
         # Only initialize culling if available.  Warning message will be issued in gatewayapp at startup.
         func = getattr(self, 'initialize_culler', None)
         if func:
@@ -166,7 +183,7 @@ class RemoteMappingKernelManager(MappingKernelManager):
             The uuid string to associate with the new kernel
         """
         env = kwargs.get('env')
-        if env and env.get('KERNEL_ID'):   # If there's a KERNEL_ID in the env, check it out
+        if env and env.get('KERNEL_ID'):  # If there's a KERNEL_ID in the env, check it out
             # convert string back to UUID - validating string in the process.
             str_kernel_id = env.get('KERNEL_ID')
             try:
@@ -233,7 +250,7 @@ class RemoteKernelManager(IOLoopKernelManager):
         """
         env = kwargs.get('env', {})
         self.user_overrides.update({key: value for key, value in env.items()
-                                   if key.startswith('KERNEL_') or
+                                    if key.startswith('KERNEL_') or
                                     key in self.parent.parent.env_process_whitelist or
                                     key in self.parent.parent.env_whitelist})
 
@@ -352,8 +369,8 @@ class RemoteKernelManager(IOLoopKernelManager):
                             else:  # Python 3
                                 self.sigint_value = sig_value.value
                             self.log.debug(
-                                "Converted EG_ALTERNATE_SIGINT '{}' to value '{}' to use as interrupt signal.".
-                                format(alt_sigint, self.sigint_value))
+                                "Converted EG_ALTERNATE_SIGINT '{}' to value '{}' to use as interrupt signal.".format(
+                                    alt_sigint, self.sigint_value))
                         except AttributeError:
                             self.log.warning("Error received when attempting to convert EG_ALTERNATE_SIGINT of "
                                              "'{}' to a value. Check kernelspec entry for kernel '{}' - using "
--- a/enterprise_gateway/services/kernels/remotemanager.py
+++ b/enterprise_gateway/services/kernels/remotemanager.py
@@ -50,7 +50,7 @@ class RemoteMappingKernelManager(MappingKernelManager):
     def _kernel_manager_class_default(self):
         return 'enterprise_gateway.services.kernels.remotemanager.RemoteKernelManager'
 
-    def _check_kernel_id(self, kernel_id):
+    def check_kernel_id(self, kernel_id):
         """Check that a kernel_id exists and raise 404 if not."""
         if kernel_id not in self:
             if not self._refresh_kernel(kernel_id):
@@ -59,12 +59,7 @@ class RemoteMappingKernelManager(MappingKernelManager):
 
     def _refresh_kernel(self, kernel_id):
         self.parent.kernel_session_manager.load_session(kernel_id)
-
-        if kernel_id in self.parent.kernel_session_manager._sessions:
-            return self.parent.kernel_session_manager._start_session(
-                self.parent.kernel_session_manager._sessions.get(kernel_id))
-
-        return False
+        return self.parent.kernel_session_manager.start_session(kernel_id)
 
     @gen.coroutine
     def start_kernel(self, *args, **kwargs):
--- a/enterprise_gateway/services/sessions/kernelsessionmanager.py
+++ b/enterprise_gateway/services/sessions/kernelsessionmanager.py
@@ -18,8 +18,6 @@ kernels_lock = threading.Lock()
 # These will be located under the `persistence_root` and exist
 # to make integration with ContentsManager implementations easier.
 KERNEL_SESSIONS_DIR_NAME = "kernel_sessions"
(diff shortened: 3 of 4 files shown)


enterprise_gateway/services/sessions/kernelsessionmanager.py
```diff
@@ -279,22 +277,48 @@ def persistence_root_default(self):
...
+            temp_sessions = self._save_sessions(kernel_session_file_path)
+        else:
+            kernel_session_files = [json_files for json_files in os.listdir(self._get_sessions_loc()) if
+                                    json_files.endswith('.json')]
+
+            temp_sessions = self._save_sessions(kernel_session_files)
+
+        self._sessions = KernelSessionManager.post_load_transformation(temp_sessions)
+
+    def _save_sessions(self, kernel_session_files):
```
kevin-bates: Not sure this method is necessary.  I think `post_load_transformation()` and `pre_save_transformation()` need to be updated to work on single sessions and then the assignment (on line 310 above) would be an update to `self._sessions` for a specific kernel_id and its loaded (and transformed) session.

enterprise_gateway/services/sessions/kernelsessionmanager.py
```diff
@@ -279,22 +277,48 @@ def persistence_root_default(self):
...
+            os.rmdir(self._get_sessions_loc())
+            for kernel_id, session in self._sessions.items():
+                if kernel_id is not None:
+                    json_file_name = "".join(['kernel_', kernel_id, '.json'])
+                    kernel_session_file_path = os.path.join(self._get_sessions_loc(), json_file_name)
+                    temp_session = dict()
+                    temp_session[kernel_id] = session
+                    with open(kernel_session_file_path, 'w') as fp:
+                        json.dump(KernelSessionManager.pre_save_transformation(temp_session), fp)
+                        fp.close()
```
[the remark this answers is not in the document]
Gsbreddy: I didn't want to change base abstract method. If that's ok, I would be happy to make this change in base class.

enterprise_gateway/services/sessions/kernelsessionmanager.py
```diff
@@ -279,22 +277,48 @@ def persistence_root_default(self):
...
 
     def save_sessions(self):
         if self.enable_persistence:
-            with open(self.kernel_session_file, 'w') as fp:
-                json.dump(KernelSessionManager.pre_save_transformation(self._sessions), fp)
-                fp.close()
+            os.rmdir(self._get_sessions_loc())
+            for kernel_id, session in self._sessions.items():
+                if kernel_id is not None:
+                    json_file_name = "".join(['kernel_', kernel_id, '.json'])
```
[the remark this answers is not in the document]
Gsbreddy: Yes, will do this change from kernel-<id>.json to just <id>.json

enterprise_gateway/services/sessions/kernelsessionmanager.py
Gsbreddy: Again I didn't want to change any base methods, so I built subclass accordingly. Let me make these changes and update the PR.

enterprise_gateway/services/sessions/kernelsessionmanager.py
```diff
@@ -279,22 +277,48 @@ def persistence_root_default(self):
...
 
     def load_sessions(self):
         if self.enable_persistence:
-            if os.path.exists(self.kernel_session_file):
-                self.log.debug("Loading saved sessions from {}".format(self.kernel_session_file))
-                with open(self.kernel_session_file) as fp:
-                    self._sessions = KernelSessionManager.post_load_transformation(json.load(fp))
+            self._load_sessions()
+
+    def _load_sessions(self, kernel_id=None):
```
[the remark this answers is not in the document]
Gsbreddy: You suggest we should have two more abstract methods load_session(id) save_session(id). Please correct me if I am wrong.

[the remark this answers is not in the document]
kevin-bates: Yes - every subclass will need to implement the singleton versions that load/save individual files/records/etc.

If we also add 'get_all' methods that return enumerations of the entries, then the full implementations for the plural methods (load_sessions, save_sessions) could be implemented in the abstract base class since there would no longer be "storage-specific" knowledge need for those methods.  But we can leave that for another exercise.

kevin-bates:
@Gsbreddy - is this ready for review?  I see you've made some recent changes.  If it's not ready, please prefix the PR's title with `[WIP]` and, once that prefix is removed, I'll be happy to review things.

At quick glance, I think we will want the ability to start a single session as well (given a kernel_id).  Also, the removal code in `save_sessions()` doesn't make sense to me.  In fact, I'm not sure we even need `save_sessions` since all sessions will be saved as a given kernel starts.  Unlike `load_sessions()` where the startup of a new server may want to load all persisted sessions.  This would apply to Active/Passive configurations.

This is coming along nicely - thank you!

Gsbreddy:
@kevin-bates To answer your question, `load_session(kernel_id)` will have the ability to start a single session. In next PR, I will handle the same in `remotemanager.py`. I get that the removal code in `save_sessions()` doesn't make sense, it's just the name actually. Let me rename it to something more meaningful like `commit_sessions`

kevin-bates:
Not really following your last statement.  Methods that ONLY delete persisted sessions shouldn't be named "save_" or "commit_".  That said, I'll wait for `[WIP]` to be removed before reviewing.  Thanks again.

Gsbreddy:
Sure. Thanks.

Gsbreddy:
Please retrigger travis.

kevin-bates:
I restarted the build and it still failed rather quickly.  The issue is due to a linting problem:
```
enterprise_gateway/services/kernels/remotemanager.py:373:37: E131 continuation line unaligned for hanging indent
Makefile:52: recipe for target 'lint' failed
make: *** [lint] Error 1
```
I recommend you run `make lint` prior to pushing changes.

Gsbreddy:
Please review the changes @kevin-bates

kevin-bates:
Hi @Gsbreddy - I just wanted you to know that I'm not ignoring this.  I'm on vacation until 2020, but I hope to sit down and take a good look at this in the next few days.  This will be a great addition and I'm excited to take a look!

enterprise_gateway/services/sessions/kernelsessionmanager.py
```diff
@@ -174,39 +171,40 @@ def _delete_sessions(self, kernel_ids):
...
-
-            self.save_sessions()  # persist changes
+                kernel_session = self._sessions.get(kernel_id, None)
+                if kernel_session is not None:
+                    username = kernel_session['username']
+                    if username in self._sessionsByUser and kernel_id in self._sessionsByUser[username]:
+                        self._sessionsByUser[username].remove(kernel_id)
+                    self._sessions.pop(kernel_id, None)
+
+            self.delete_sessions()
```
kevin-bates: Rather than store the kernel_ids in `self.sessions_to_remove`, I think we should just define the method to take a list of kernel_ids - just like `_delete_sessions` and avoid maintaining the `self.sessions_to_remove` member since `delete_sessions()` is only called from `_delete_sessions(kernel_ids)`.  I'm concerned that a member variable will be difficult to manage between calls since its essentially shared state.

enterprise_gateway/services/sessions/kernelsessionmanager.py
```diff
@@ -216,12 +214,26 @@ def load_sessions(self):
...
-        Saves the sessions dictionary to persistent store.  Caller is responsible for synchronizing call.
+        Load and initialize _sessions member from persistent storage for a single kernel.  This method is called from
+        refresh_sessions().
+        """
+        raise NotImplementedError("KernelSessionManager.load_sessions() requires an implementation!")
+
+    # abstractmethod
+    def delete_sessions(self):
+        """
+        Persists the changes to the persistent storage.  Caller is responsible for synchronizing call.
```
kevin-bates: Fix the description to be relative to deletion.  Add a kernel_ids list parameter - depending on previous comment discussion.

Please update the `NotImplementedError` description below as well.  Thanks.

enterprise_gateway/services/sessions/kernelsessionmanager.py
```diff
@@ -279,21 +291,47 @@ def persistence_root_default(self):
...
-            self.log.info("Kernel session persistence location: {}".format(self.kernel_session_file))
+            self.log.info("Kernel session persistence location: {}".format(self._get_sessions_loc()))
 
-    def save_sessions(self):
+    def delete_sessions(self):
         if self.enable_persistence:
-            with open(self.kernel_session_file, 'w') as fp:
-                json.dump(KernelSessionManager.pre_save_transformation(self._sessions), fp)
-                fp.close()
+            for kernel_id in self.sessions_to_remove:
```
kevin-bates: Per previous comment, I'd prefer to provide the list of sessions (kernel_ids) to remove rather than rely on the member variable.

enterprise_gateway/services/sessions/kernelsessionmanager.py
```diff
@@ -279,21 +291,47 @@ def persistence_root_default(self):
...
+
+    def load_session(self, kernel_id):
+        if kernel_id is not None:
+            kernel_file_name = "".join([kernel_id, '.json'])
+            kernel_session_file_path = os.path.join(self._get_sessions_loc(), kernel_file_name)
+            if os.path.exists(kernel_session_file_path):
+                self.log.debug("Loading saved session(s) from {}".format(kernel_session_file_path))
+                with open(kernel_session_file_path) as fp:
+                    self._sessions.update(KernelSessionManager.post_load_transformation(json.load(fp)))
                     fp.close()
```
kevin-bates: The previous 6 lines are duplciated in `load_sessions`.  Let's create a private method, `_load_session_from_file(file_name)` that is essentially this...
```python
        def _load_session_from_file(self, file_name):
            kernel_session_file_path = os.path.join(self._get_sessions_loc(), file_name)
            if os.path.exists(kernel_session_file_path):
                self.log.debug("Loading saved session(s) from {}".format(kernel_session_file_path))
                with open(kernel_session_file_path) as fp:
                    self._sessions.update(KernelSessionManager.post_load_transformation(json.load(fp)))
                    fp.close()
```
and replace the duplicated lines.

enterprise_gateway/services/kernels/handlers.py
```diff
@@ -102,10 +103,18 @@ class KernelHandler(TokenAuthorizationMixin,
...
     """
+
     def options(self, **kwargs):
         """Method for properly handling CORS pre-flight"""
         self.finish()
 
+    @web.authenticated
+    def get(self, kernel_id):
+        km = self.kernel_manager
+        km._check_kernel_id(kernel_id)
```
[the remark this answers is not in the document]
Gsbreddy: Done.

enterprise_gateway/services/kernels/remotemanager.py
```diff
@@ -46,9 +46,26 @@ def get_process_proxy_config(kernelspec):
 
 class RemoteMappingKernelManager(MappingKernelManager):
     """Extends the MappingKernelManager with support for managing remote kernels via the process-proxy. """
+
     def _kernel_manager_class_default(self):
         return 'enterprise_gateway.services.kernels.remotemanager.RemoteKernelManager'
 
+    def _check_kernel_id(self, kernel_id):
```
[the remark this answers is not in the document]
Gsbreddy: Done.

enterprise_gateway/services/sessions/kernelsessionmanager.py
Gsbreddy: Done.

enterprise_gateway/services/kernels/remotemanager.py
```diff
@@ -46,9 +46,26 @@ def get_process_proxy_config(kernelspec):
...
+            if not self._refresh_kernel(kernel_id):
+                self.parent.kernel_session_manager.delete_session(kernel_id)
+                raise web.HTTPError(404, u'Kernel does not exist: %s' % kernel_id)
+
+    def _refresh_kernel(self, kernel_id):
+        self.parent.kernel_session_manager.load_session(kernel_id)
+
+        if kernel_id in self.parent.kernel_session_manager._sessions:
+            return self.parent.kernel_session_manager._start_session(
+                self.parent.kernel_session_manager._sessions.get(kernel_id))
```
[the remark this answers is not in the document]
Gsbreddy: done.

enterprise_gateway/services/sessions/kernelsessionmanager.py
Gsbreddy: Done.

enterprise_gateway/services/sessions/kernelsessionmanager.py
Gsbreddy: Done.

enterprise_gateway/services/sessions/kernelsessionmanager.py
Gsbreddy: Done.

enterprise_gateway/services/sessions/kernelsessionmanager.py
```diff
@@ -66,6 +63,7 @@ def __init__(self, kernel_manager, **kwargs):
         self.kernel_manager = kernel_manager
         self._sessions = dict()
         self._sessionsByUser = dict()
+        self.sessions_to_remove = []
```
kevin-bates: Now that `delete_sessions` takes a list of kernel_ids, we don't need a member variable for this.

Gsbreddy: Done.

Gsbreddy:
I agree, we have to think this through and come up with a solution..

kevin-bates:
I was thinking a few hours after my last response that using sticky sessions will go a long way here.  That way, the lifecycle of the kernel would stay pinned to a single instance.  Should that instance go down, then your changes would kick in on the other instance - and subsequent requests on that kernel would stay pinned to that server.  Since there's still only one instance ever operating against a given kernel, we should still be okay.  Of course, if the original instance didn't entirely go down (say due to a web-level network issue), then we could be facing some chaotic behaviors once the kernel is terminated.

kevin-bates:
Shoot, there's one thing I wanted to followup on and I mentioned it above in [this comment](https://github.com/jupyter/enterprise_gateway/pull/737#pullrequestreview-333466040).  In thinking more about this, I don't think we want this code at all.  Even users deploying an active/passive operation would still benefit from your change and leaving that code in place will certainly mess things up unless all EG instances started when no kernels are hosted by other instances.  So I think we should probably remove that call as part of this PR.

Gsbreddy:
I will remove the code `start_sessions()` from kernelsessionmanager.py and update the PR.

Gsbreddy:
> Shoot, there's one thing I wanted to followup on and I mentioned it above in [this comment](https://github.com/jupyter/enterprise_gateway/pull/737#pullrequestreview-333466040). In thinking more about this, I don't think we want this code at all. Even users deploying an active/passive operation would still benefit from your change and leaving that code in place will certainly mess things up unless all EG instances started when no kernels are hosted by other instances. So I think we should probably remove that call as part of this PR.

Done.

kevin-bates:
Sorry for the delayed response as I've been out on holiday (Happy New Year!).  Thank you for this awesome contribution!

Issue #749, opened 2019-11-14
/dev/shm shared memory expansion 


JorisJamers:
We are trying to expand our shared memory `/dev/shm` in the kernel pods. In Kubernetes this can be done with the following mount : 

```
spec:
  volumes:
  - name: dshm
    emptyDir:
      medium: Memory
  containers:
  - image: gcr.io/project/image
    volumeMounts:
      - mountPath: /dev/shm
        name: dshm
``` 

When adding this to the jupyter enterprise gateway we also get an expansion on the shared memory inside of that pod. Now, I was wondering if it is possible to get this working in the Kernel pods as well. 

I have tried to add it in the `kernel-pod.yaml.j2` file hardcoded, but this does not work. I'm not 100% sure how that `launch_kubernetes.py` script works so I was wondering if anyone has ever done this?

kevin-bates:
Thanks for the issue.  I have a couple questions about your attempt.
1. Is the kernel you're trying to expand a spark-based kernel or a _vanilla_ kernel (doesn't launch via `spark-submit`)?  I'm assuming the latter since you reference `launch_kubernetes.py` - but need to clarify since pod creation is not under our control for the spark-based kernels ([at least not until Spark 3.0 last I heard](#559)).
2. Are you updating the `kernel-pod.yaml.j2` files on the deployment itself?  I.e., you've exposed the kernelspecs folder hierarchy out of the EG pod and have modified the appropriate kernel-pod yaml files in each of the kernelspec scripts directories?
3. Are you seeing any kind of issue when attempting this?  Have you confirmed the yaml corresponding to the launched kernel pod to see if the changes (or any others) have taken affect?  (You might try to confirm other, more simple, changes like adding a bogus environment variable, etc.)

I see no reason why this wouldn't work especially since you see this succeed relative to the EG pod.

JorisJamers:
Hi Kevin, it's a vanilla kernel indeed. I updated the `kernel-pod.yaml.j2` file on the node of the kubernetes cluster, whilest the folder being mounted in the EG. 

I updated the files as shown in the docs, meaning : 

```
{
  "language": "python",
  "display_name": "rapids",
  "metadata": {
    "process_proxy": {
      "class_name": "enterprise_gateway.services.processproxies.k8s.KubernetesProcessProxy",
      "config": {
        "image_name": "tafl-docker.artifactrepo.jnj.com/dgx2rapids/kernel:35"
      }
    }
  },
  "env": {
  },
[...]

PR #750, opened 2019-11-18
Restructure the kernel-pod.yaml.j2 template

JorisJamers:
As talked about in issue #749 there was a need to document the conditional and unconditional volumeMounts and volumes. 

This should be documented?

### DIFF
--- a/docs/source/kernel-kubernetes.md
+++ b/docs/source/kernel-kubernetes.md
@@ -395,7 +395,7 @@ There are a number of items worth noting:
 1. The parameters to the launcher that is built into the image are communicated via environment variables as noted in the `env:` section above.
 
 ### Unconditional Volume Mounts
-Unconditional volume mounts can be added in the `kernel-pod.yaml.j2` template. An example for these unconditional can be the docker shared memory. For some I/O jobs the pod will need more than the default `64mb` of shared memory on the `/dev/shm` path.
+Unconditional volume mounts can be added in the `kernel-pod.yaml.j2` template. An example of these unconditional volume mounts can be found when extending docker shared memory. For some I/O jobs the pod will need more than the default `64mb` of shared memory on the `/dev/shm` path.
 
 ```
 volumeMounts:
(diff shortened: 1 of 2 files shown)


kevin-bates:
@JorisJamers - thank you for the contribution.  Yes, we should probably document this, probably as a subsection at the end of this (parent) section: https://jupyter-enterprise-gateway.readthedocs.io/en/latest/kernel-kubernetes.html#kubernetes-kernel-instances

Seems like a `Mounting volumes` section would be very helpful where we show both "unconditional" mount adjustments, along with those that might vary per user.  You could probably use the examples from @esevan's PR in that portion.  

If you're not comfortable including doc changes, we can create an issue with the appropriate information instead.

JorisJamers:
Started adding some documentation in the appropriate section, feel free to add / change / move the documentation at your will @kevin-bates, used the `/dev/shm` example because I'm most familiar with that use case.

docs/source/kernel-kubernetes.md
```diff
@@ -382,18 +382,62 @@ spec:
...
 1. Kernel pods have restart policies of `Never`.  This is because the Jupyter framework already has built-in logic for auto-restarting failed kernels and any other restart policy would likely interfere with the built-in behaviors.
 1. The parameters to the launcher that is built into the image are communicated via environment variables as noted in the `env:` section above.
 
+### Unconditional Volume Mounts
+Unconditional volume mounts can be added in the `kernel-pod.yaml.j2` template. An example for these unconditional can be the docker shared memory. For some I/O jobs the pod will need more than the default `64mb` of shared memory on the `/dev/shm` path.
```
kevin-bates: I think we're going to want some examples for how to use the the conditional logic.  This is great for the unconditional discussion.  We can add the conditional discussion in a follow-on PR if you prefer.

How about updating the second sentence to some like this... _An example of these unconditional volume mounts can be found when extending docker shared memory._

JorisJamers:
Sentence change happened! Someone might add in to the conditional volume mounts because I'm not too familiar with that concept.

Issue #762, opened 2019-12-19
Query for Spark application state failed


amangarg96:
## Description
The EG instance becomes unresponsive, returning HTTP 500: Internal Server Error to the client. In the EG logs, the EG is repeatedly querying for the state of the Spark Application (Every 3-4 second), which is failing.

I have not found a way to reproduce the issue. Restarting the EG instance works, but this is a frequently occurring issue (Once every 3-4 days). 

The YARN API it's using to query the state of the application (http://<rm-url>:<rm-port>/ws/v1/cluster/apps/application_1573339086213_974645/state), is working and returning the state of the spark application. 

Going through the stack trace, I observed that `resource_manager` object of _YarnClusterProcessProxy_ becomes `None`, because of which the query is failing.

## Screenshots / Logs
Stack trace of EG:
```
[W 2019-12-18 12:33:27.020 EnterpriseGatewayApp] Query for application 'application_1573339086213_974645' state failed with exception: 'HTTPConnectionPool(host=<rm-url>, port=<rm-port>): Max retries exceeded with url: /ws/v1/cluster/apps/application_1573339086213_974645/state (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7f77d0408cf8>: Failed to establish a new connection: [Errno 11] Resource temporarily unavailable'))'.  Continuing...
[E 191218 12:33:27 ioloop:909] Exception in callback <bound method KernelRestarter.poll of <jupyter_client.ioloop.restarter.IOLoopKernelRestarter object at 0x7f782b480550>>
    Traceback (most recent call last):
      File "/usr/share/mlp-jeg/lib/python3.7/site-packages/tornado/ioloop.py", line 907, in _run
        return self.callback()
      File "/usr/share/mlp-jeg/lib/python3.7/site-packages/jupyter_client/restarter.py", line 93, in poll
        if not self.kernel_manager.is_alive():
      File "/usr/share/mlp-jeg/lib/python3.7/site-packages/jupyter_client/manager.py", line 453, in is_alive
        if self.kernel.poll() is None:
      File "/usr/share/mlp-jeg/lib/python3.7/site-packages/enterprise_gateway/services/processproxies/yarn.py", line 152, in poll
        state = self.query_app_state_by_id(self.application_id)
      File "/usr/share/mlp-jeg/lib/python3.7/site-packages/enterprise_gateway/services/processproxies/yarn.py", line 378, in query_app_state_by_id
        return response.data['state']
[...]

PR #764, opened 2020-01-03
Fix YARN API error handling

kevin-bates:
This addresses issues relative to interactions with the yarn-api-client.
It introduces `last_known_state` so that cases where the YARN resource
manager has access issues during EG's polling interval (every 3 seconds)
the last known state will be used.  Previously, EG would essentially
crash when referencing a null response (which has also been addressed).
With these changes, EG may be able to continue if the kernel is still
communicating, even though YARN's RM is not.

This PR is a prerequisite to troubleshooting #762.

### DIFF
--- a/enterprise_gateway/services/processproxies/yarn.py
+++ b/enterprise_gateway/services/processproxies/yarn.py
@@ -15,7 +15,7 @@ from yarn_api_client.resource_manager import ResourceManager
 from .processproxy import RemoteProcessProxy
 
 # Default logging level of yarn-api and underlying connectionpool produce too much noise - raise to warning only.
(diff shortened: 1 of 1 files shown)


kevin-bates:
@amangarg96 - I know you're working on the 1.x branch, but I'm hoping you might be able to apply these changes to your configuration to see if it sheds any light on #762.  I've taken these for a spin on my 2.x env and things appear to be working as expected.

kevin-bates:
@amangarg96 - here's a version relative to the 1.x branch.  If you could please take this for a spin, it would be greatly appreciated.  Please note that I have not deployed this version.  I've only updated according to this PR.

[yarn.py.txt](https://github.com/jupyter/enterprise_gateway/files/4020709/yarn.py.txt)

PR #775, opened 2020-01-21
[WIP] update enterprise-gateway.yaml & kernel-pod.yaml.j2 with k8s limits

### LINKED ISSUE #770: Limits and Resource Quotas
Since updating the conditional volume mounts I have encountered a new issue. While having resource quotas and limits in any given namespace where the Enterprise Gateway is deployed. The EG and the Kernels are not spawning because the following specs are missing in the templates : 

kernel-pod.yaml.j2
```
    name: "{{ kernel_pod_name }}"
    resources:
      limits:
        cpu: "1"
        memory: "1Gi"
        nvidia.com/gpu: "0"
    {% if kernel_working_dir is defined %}
    workingDir: "{{ kernel_working_dir }}"
    {% endif %}
    volumeMounts:
```

Enterprise Gateway Deployment YAML snippet 
```
image: elyra/enterprise-gateway:dev
        resources:
          limits:
            cpu: "1"
            memory: "1Gi"
            nvidia.com/gpu: "0"
```

Image puller daemonset YAML snippet 
```
image: elyra/kernel-image-puller:dev
        resources:
          limits:
            cpu: "1"
            memory: "1Gi"
            nvidia.com/gpu: "0"
``` 

Basically in all the templates there is no definition for limits, this blocks the pods from spawning in a namespace where these limits and quotas are defined. 

I think this is something that can be catched with an Admission Controller, but might be a good idea to implement it in the templates? 

I'm not sure how to pass those variables, so in my own setup I hardcoded the limits in the templates. 

I browsed some issues but I don't think this is mentioned somewhere already? Correct me if I'm wrong @kevin-bates

JorisJamers:
Documentation has to be added, updated the resource limits in the following files : 

- enterprise-gateway.yaml 
- kernel-pod.yaml.j2 

Closes #770

### DIFF
--- a/etc/kernel-launchers/kubernetes/scripts/kernel-pod.yaml.j2
+++ b/etc/kernel-launchers/kubernetes/scripts/kernel-pod.yaml.j2
@@ -54,11 +54,11 @@ spec:
       value: "{{ kernel_namespace }}"
     image: "{{ kernel_image }}"
     name: "{{ kernel_pod_name }}"
-    resources:
-      limits:
-        cpu: "1"
-        memory: "2Gi"
-        nvidia.com/gpu: "0"
+    # resources:
+    #   limits:
+    #     cpu: "1"
+    #     memory: "2Gi"
+    #     nvidia.com/gpu: "0"
     {% if kernel_working_dir is defined %}
     workingDir: "{{ kernel_working_dir }}"
     {% endif %}
--- a/etc/kernel-launchers/kubernetes/scripts/kernel-pod.yaml.j2
+++ b/etc/kernel-launchers/kubernetes/scripts/kernel-pod.yaml.j2
@@ -54,6 +54,11 @@ spec:
       value: "{{ kernel_namespace }}"
     image: "{{ kernel_image }}"
     name: "{{ kernel_pod_name }}"
+    resources:
+      limits:
+        cpu: "1"
+        memory: "2Gi"
+        nvidia.com/gpu: "0"
     {% if kernel_working_dir is defined %}
     workingDir: "{{ kernel_working_dir }}"
     {% endif %}
--- a/etc/kubernetes/enterprise-gateway.yaml
+++ b/etc/kubernetes/enterprise-gateway.yaml
@@ -142,6 +142,11 @@ spec:
 
         # Ensure the following VERSION tag is updated to the version of Enterprise Gateway you wish to run
         image: elyra/enterprise-gateway:dev
+        resources:
+          limits:
+            cpu: "1"
+            memory: "2Gi"
+            nvidia.com/gpu: "0"
         # Use IfNotPresent policy so that dev-based systems don't automatically
         # update. This provides more control.  Since formal tags will be release-specific
         # this policy should be sufficient for them as well.
@@ -167,17 +172,22 @@ metadata:
 spec:
   selector:
     matchLabels:
-      name: kernel-image-puller 
+      name: kernel-image-puller
   template:
     metadata:
       labels:
-        name: kernel-image-puller 
+        name: kernel-image-puller
         app: enterprise-gateway
         component: kernel-image-puller
     spec:
       containers:
-      - name: kernel-image-puller 
+      - name: kernel-image-puller
         image: elyra/kernel-image-puller:dev
+        resources:
+          limits:
+            cpu: "1"
+            memory: "2Gi"
+            nvidia.com/gpu: "0"
         env:
           - name: KIP_GATEWAY_HOST
             value: "http://enterprise-gateway.enterprise-gateway:8888"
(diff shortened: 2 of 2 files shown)


JorisJamers:
It's also possible to ask for more resources of course. But when you ask too much resources with a very limited namespace the kernel pods will not spawn. Something we have to take into account.

etc/kernel-launchers/kubernetes/scripts/kernel-pod.yaml.j2
```diff
@@ -54,6 +54,11 @@ spec:
       value: "{{ kernel_namespace }}"
     image: "{{ kernel_image }}"
     name: "{{ kernel_pod_name }}"
+    # resources:
+    #   limits:
+    #     cpu: "1"
+    #     memory: "2Gi"
+    #     nvidia.com/gpu: "0"
```
kevin-bates: Are the `requests` stanzas similar?  If so, I think it would make sense to add a `requests` stanza, still commented out, but rather than using fixed values, use the following parameters, `{{ kernel_cpus }}`, `{{ kernel_memory }}`, and `{{ kernel_gpus }}`.  This would tie in nicely for any parameterization efforts moving forward.

etc/kubernetes/enterprise-gateway.yaml
```diff
@@ -142,6 +142,11 @@ spec:
 
         # Ensure the following VERSION tag is updated to the version of Enterprise Gateway you wish to run
         image: elyra/enterprise-gateway:dev
+        # resources:
+        #   limits:
+        #     cpu: "1"
+        #     memory: "2Gi"
+        #     nvidia.com/gpu: "0"
```
kevin-bates: The EG and KIP yaml settings have corresponding helm charts.  For those, it probably makes sense to parameterize the limits when helm is used.  The means of doing this is fairly straightforward if you'd like give this a try.

The EG definition is in the deployment.yaml file, while KIP is defined in daemonset.yaml and their respective "variables/values" would be listed in values.yaml.

etc/kernel-launchers/kubernetes/scripts/kernel-pod.yaml.j2
JorisJamers: Yep `requests` are similar, and it would be better to not use fixed values indeed.

kevin-bates:
No worries at all.  How about prefixing your PR title with "[WIP] " to indicate its a work in progress and we can continue the review when you're ready.

wiput1999:
Hi, I'm looking for the same feature likes this PR. For K8s, Can we define with prefix `KERNEL_` for resource limit too?

wiput1999:
My opinion is adding `KERNEL_CPU_LIMIT` and `KERNEL_MEM_LIMIT` and adding condition into .j2 file

kevin-bates:
Hi @wiput1999 - thank you for your input.  Yes, all templated parameters named like `{{ kernel_xxx }}` correspond to an env in the kernel start request body of `KERNEL_XXX`.  I don't think these envs need to be further qualified with `K8S` indicators since these kinds of limits are fairly common across resource managers.

Given the 10 months that @JorisJamers has last updated this PR the world has gone through some chaotic times, I think it would be fair to check in with Joris to see where things stand relative to this PR. I'm inclined to wait a few days and, if no response has been made in that time, go ahead and entertain a more complete PR.  Is that okay with you?  In the meantime, your time wouldn't go to waste working on similar changes that, should Joris come back to this, could be used in a collaborative manner.

JorisJamers:
@kevin-bates , @wiput1999 times have been chaotic the past few months indeed. I think the lack of documentation and templating is / was the reason we never used this PR. 

Since my focus has changed a bit on my own project we are using this "hacky / hard coded" version for our own environment. I'm still having those "few busy days at work" focusing on different topics. 

@wiput1999 feel free to continue on the work we have started, because I'm still not too familiar with .j2 templating. :)

wiput1999:
@kevin-bates as I see in your comment you suggest to add `{{ kernel_cpus }}`, `{{ kernel_memory }}`, and `{{ kernel_gpus }}` right? I'll wait a few days if @JorisJamers maybe busy to continue working on this. I'll open new PR with completed work. Thank you

wiput1999:
@kevin-bates another suggestion for me about resources quota. I think `{{ kernel_cpus }}`, `{{ kernel_memory }}`, and `{{ kernel_gpus }}` use for pod resource request and `{{ kernel_cpus_limit }}`, `{{ kernel_memory_limit }}`, and `{{ kernel_gpus_limit }}` for pod resource limit. How about your opinion?

kevin-bates:
@JorisJamers - thank you for your quick response.  Your time spent on this so far is really appreciated.

@wiput1999, given Joris' response, I think it's probably best we close this PR and I ask that you open a new PR associated with your own fork/branch.

Regarding your last comment, yes, I think having entries for both requests and limits is the correct approach.  Thank you.

@wiput1999 - when opening your new PR, please add a reference to this PR so as to keep track of its history.

wiput1999:
@JorisJamers I'll continue on this. Thank you for the discussion with @kevin-bates that can make me easier to get start work 😄

PR #780, opened 2020-02-12
Cap yarn-api-client at < 1.0

kevin-bates:
Significant changes are required to move to the 1.x release of yarn-api-client so we should cap it.
Note this change is only for the 1.x branch!


kevin-bates:
Merging w/o approvals since its a trivial change and the build issue is unrelated (and requires major lifting).

PR #781, opened 2020-02-12
Update sbt-launch location to fix travis build issue

kevin-bates:


kevin-bates:
Merging w/o approvals since its a trivial change and the build issue is unrelated (and requires major lifting).

Issue #797, opened 2020-03-28
KeyError: 'jinja2_env' if kernel does not exist


bsdz:
##  Description

When client requests missing kernel from enterprise gateway, the gateway raise a 404 error that relies on notebook error handling routine which uses jinja2 templates. However, jinja is not set up for enterprise gateway.

## Steps to replicate

1. Used the following vscode launch.json config:
```json
{
    "version": "0.2.0",
    "configurations": [
        {
            "name": "EntGW",
            "type": "python",
            "request": "launch",
            "pythonPath": "/pyenvs/jupyter/.venv/bin/python",
            "module": "enterprise_gateway",
            "args": [
                "--ip=0.0.0.0",
                "--port=8887",
                "--port_retries=0",
                "--log-level=DEBUG",
                "--EnterpriseGatewayApp.authorized_users=['guest']",
                "--EnterpriseGatewayApp.list_kernels=True",
                "--JupyterWebsocketPersonality.list_kernels=True",
            ],
            "justMyCode": false,
        },
        {
            "name": "JupyterLab",
            "type": "python",
            "request": "launch",
            "pythonPath": "/pyenvs/jupyter/.venv/bin/python",
            "module": "jupyterlab",
            "args": [
                "--gateway-url=http://192.168.1.100:8887",
                "--GatewayClient.http_user=guest",
                "--GatewayClient.http_pwd=guest-password",
                "--log-level=DEBUG",
            ],
            "justMyCode": false,
        }
    ]
}
```
2. Go to runtime dir and remove any kernel/nbserver files.
```bash
$ cd $(jupyter --runtime-dir)
$ rm nbserver-* kernel-*
```
3. Start Enterprise GW using launch config above.
4. Start Notebook using launch config above.
5. Create a new ipynb notebook in browser.
6. Stop Enterprise GW server.

7. In runtime dir, remove any kernel/nbserver files to simulate missing kernel.
```bash
$ rm nbserver-* kernel-* # simulate missing kernel
```
8. Start Enterprise GW again.

The following exception occurs:

```
[I 2020-03-28 11:26:48.597 EnterpriseGatewayApp] Jupyter Enterprise Gateway 2.1.0 is available at http://0.0.0.0:8887
[D 2020-03-28 11:26:48.601 EnterpriseGatewayApp] Initializing websocket connection /api/kernels/40e5c948-2367-4c07-979c-63051f7a1be7/channels
[...]

PR #798, opened 2020-03-28
fix 'jinja2_env' if kernel does not exist

bsdz:
Use APIHandler write_error routine for ZMQChannelsHandler and avoid jinja2 templates.
Added .vscode folder to .gitignore.

Fixes: #797 .

Tested manually.

Not the most elegant of fixes. In an ideal world perhaps ZMQChannelsHandler (from another package) might be subclassed rather than monkey patched.

### DIFF
--- a/enterprise_gateway/services/kernels/handlers.py
+++ b/enterprise_gateway/services/kernels/handlers.py
@@ -123,8 +123,5 @@ for path, cls in notebook_handlers.default_handlers:
         default_handlers.append((path, globals()[cls.__name__]))
     else:
         # Gen a new type with CORS and token auth
-        bases = (TokenAuthorizationMixin, CORSMixin, cls)
-        if cls.__name__ == 'ZMQChannelsHandler':
-            from notebook.base.handlers import APIHandler
-            cls.write_error = APIHandler.write_error
+        bases = (TokenAuthorizationMixin, CORSMixin, JSONErrorsMixin, cls)
         default_handlers.append((path, type(cls.__name__, bases, {})))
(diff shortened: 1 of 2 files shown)


enterprise_gateway/services/kernels/handlers.py
```diff
@@ -124,4 +124,7 @@ def get(self, kernel_id):
     else:
         # Gen a new type with CORS and token auth
         bases = (TokenAuthorizationMixin, CORSMixin, cls)
+        if cls.__name__ == 'ZMQChannelsHandler':
+            from notebook.base.handlers import APIHandler
+            cls.write_error = APIHandler.write_error
```
kevin-bates: One of the primary purposes of the gateway's set of mixins is to use `JSONErrorsMixin` for dealing with errors. I'll admit, I'm not up on how all the handlers work (outside of calling the python API for dealing with kernels, etc.), but I think we should either look into injecting JSONErrorsMixin into `bases` or add an override for `ZMQChannelsHandler` like we do for a couple of the kernel handlers.

Also, since `ZMQChannelsHandler` is completely inherited from notebook, why is this change necessary since `APIHandler` is not in its class hierarchy in the first place?  Neither `TokenAuthorizationMixin` or `CORSMixin` implement `write_error()`.

Thanks for your help on this.

bsdz: Mea culpa. I didn't see the existing JSONErrorsMixin solution used above so patched earlier without thinking. Updated as your suggestion. Manually tested again and resolves same issue. Not sure if will cause any regressions in other handlers from notebook_handlers.default_handlers; seems unlikely.

kevin-bates:
I'm sorry, I see you made that comment in the issue.

enterprise_gateway/services/kernels/handlers.py
kevin-bates: Thank you Blair.  Yeah, I think we'll be okay since EG uses very few of the NB handlers.

Issue #779, opened 2020-02-04
Custom API path /api/kernelspec


lucabem:
Hi @kevin-bates:

I want to modify the behaviour of the API call /api/kernelspec. Now, when we start a notebook server via JupyterHub, this server makes a call to /api/kernelspec and the response is a list of kernels in the NFS server.

I want to filter that kernels by user who have created them. Instead of calling /api/kernelspec, it should be /api/kernelspec/{username}. 
I have seen the [swager.yaml](https://github.com/jupyter/enterprise_gateway/blob/master/enterprise_gateway/services/api/swagger.yaml) but i dont find the python file of the API.

Thank you and have a nice day

kevin-bates:
Hi @lucabem - its good to hear from you again.  

Kernel specs are not "user owned" but are "system owned".  They merely contain the metadata used to start a kernel.  For example, User A and User B can each start an instance of kernel 'foo' using the 'foo' kernelspec.  You don't see the kernelspecs handlers explicitly defined in Enterprise Gateway because they're essentially [inherited from Notebook](https://github.com/jupyter/enterprise_gateway/blob/master/enterprise_gateway/services/kernelspecs/handlers.py#L5).

I think you're really talking about the `/api/kernels` endpoint, which lists the running kernels.  `/api/kernels/<kernel_id>` is what is used to direct requests (like interrupt, restart, and shutdown) to a specific kernel and `GET /api/kernels/<kernel_id>` returns the model corresponding to that running kernel.

Because there's already a `/api/kernels/<something>` endpoint and, although the pattern match is against a string with embedded dashes, I don't think there's a hard requirement that kernel-id will always be a guid, so I hesitate to try to overload that endpoint for filtering by user.

Perhaps the MainKernelHandler's `GET` implementation could look for a query parameter, so something like `GET /api/kernels?username=<user>` could work?  Then you'd probably want to copy the existing implementation of list-kernels into EG and perform the post filtering based on username - and in consultation with the `KernelSessionManager` (since that's where the user-to-kernel mappings would reside).  Or better yet, drive the kernel list from `KernelSessionManager` by creating a method similar to [`active_sessions()`](https://github.com/jupyter/enterprise_gateway/blob/master/enterprise_gateway/services/sessions/kernelsessionmanager.py#L242) which returns the list of sessions for a given user.

I would be happy to answer any questions you might have should you want to provide pull request for this.

lucabem:
[...]

PR #800, opened 2020-04-13
Add username filtering for each kernel

lucabem:
We have added a new path to JEG to allow users to filter their own kernels.

We come from #779. We have discussed about the possibility of add a new path to JEG API. It allows to filter kernels checking unauthorized/authorized lists.

### DIFF
--- a/enterprise_gateway/services/api/swagger.yaml
+++ b/enterprise_gateway/services/api/swagger.yaml
@@ -86,9 +86,17 @@ paths:
       summary: Get kernel specs
       tags:
         - kernelspecs
+      parameters:
+        - name: user
+          in: query
+          description: User value that need to be considered for filtering kernels
+          required: false
+          type: string
       responses:
         200:
-          description: Returns the available kernel specs to spawn new kernels.
+          description: |
+            Returns the available kernel specs to spawn new kernels. If no query param is used, it returns all
+            kernels available.
           schema:
             type: object
             properties:
--- a/enterprise_gateway/services/api/swagger.yaml
+++ b/enterprise_gateway/services/api/swagger.yaml
@@ -89,14 +89,14 @@ paths:
       parameters:
         - name: user
           in: query
-          description: User value that need to be considered for filtering kernels
+          description: When present, kernelspec results will be filtered based on the configured authorization of specified value.
           required: false
           type: string
       responses:
         200:
           description: |
-            Returns the available kernel specs to spawn new kernels. If no query param is used, it returns all
-            kernels available.
+            If no query parameter is specified, all kernel specs will be returned;
+            otherwise the result set is filtered based on the query parameter.
           schema:
             type: object
             properties:
--- a/enterprise_gateway/services/kernelspecs/handlers.py
+++ b/enterprise_gateway/services/kernelspecs/handlers.py
@@ -1,21 +1,177 @@
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
 """Tornado handlers for kernel specs."""
-
+import glob
+import json
 import notebook.services.kernelspecs.handlers as notebook_handlers
 import notebook.kernelspecs.handlers as notebook_kernelspecs_resources_handlers
+import asyncio
+import concurrent.futures
 
+import inspect
+import os
+import logging
+import traceback
+from pygments.lexer import default
+from tornado import web, gen
+from ipython_genutils import py3compat
+from urllib.parse import unquote
 from ...mixins import TokenAuthorizationMixin, CORSMixin, JSONErrorsMixin
+from ...base.handlers import APIHandler
+
+pjoin = os.path.join
+
+
+def url_unescape(path):
+    """Unescape special characters in a URL path
+    Turns '/foo%20bar/' into '/foo bar/'
+    """
+    return u'/'.join ([
+        py3compat.str_to_unicode (unquote (p), encoding='utf8')
+        for p in py3compat.unicode_to_str (path, encoding='utf8').split ('/')
+    ])
+
+
+def maybe_future(obj):
+    """Like tornado's deprecated gen.maybe_future
+    but more compatible with asyncio for recent versions
+    of tornado
+    """
+    if inspect.isawaitable (obj):
+        return asyncio.ensure_future (obj)
+    elif isinstance (obj, concurrent.futures.Future):
+        return asyncio.wrap_future (obj)
+    else:
+        # not awaitable, wrap scalar in future
+        f = asyncio.Future ()
+        f.set_result (obj)
+        return f
+
+
+def url_path_join(*pieces):
+    """Join components of url into a relative url
+    Use to prevent double slash when joining subpath. This will leave the
+    initial and final / in place
+    """
+    initial = pieces[0].startswith ('/')
+    final = pieces[-1].endswith ('/')
+    stripped = [s.strip ('/') for s in pieces]
+    result = '/'.join (s for s in stripped if s)
+    if initial: result = '/' + result
+    if final: result = result + '/'
+    if result == '//': result = '/'
+    return result
+
+
+def is_kernelspec_model(spec_dict):
+    """Returns True if spec_dict is already in proper form.  This will occur when using a gateway."""
+    return isinstance (spec_dict, dict) and 'name' in spec_dict and 'spec' in spec_dict and 'resources' in spec_dict
+
+
+def kernelspec_model(handler, name, spec_dict, resource_dir):
+    """Load a KernelSpec by name and return the REST API model"""
+    d = {
+        'name': name,
+        'spec': spec_dict,
+        'resources': {}
+    }
+
+    # Add resource files if they exist
+    resource_dir = resource_dir
+    for resource in ['kernel.js', 'kernel.css']:
+        if os.path.exists (pjoin (resource_dir, resource)):
+            d['resources'][resource] = url_path_join (
+                handler.base_url,
+                'kernelspecs',
+                name,
+                resource
+            )
+    for logo_file in glob.glob (pjoin (resource_dir, 'logo-*')):
+        fname = os.path.basename (logo_file)
+        no_ext, _ = os.path.splitext (fname)
+        d['resources'][no_ext] = url_path_join (
+            handler.base_url,
+            'kernelspecs',
+            name,
+            fname
+        )
+    return d
+
 
 # Extends the default handlers from the notebook package with token auth, CORS
 # and JSON errors.
 default_handlers = []
-for path, cls in notebook_handlers.default_handlers:
-    # Everything should have CORS and token auth
+
+
+def exists(obj, chain):
+    _key = chain.pop(0)
+    if _key in obj:
+        return exists(obj[_key], chain) if chain else obj[_key]
+
+def apply_user_filter(kernelspec_model, kernel_user = None):
+  if kernel_user:
+    # Check unauthorized list
+    if exists(kernelspec_model, ['spec', 'metadata', 'process_proxy', 'config', 'unauthorized_users']):
+      # Check if kernel_user in kernelspec_model
+      unauthorized_list = kernelspec_model['spec']['metadata']['process_proxy']['config']['unauthorized_users']
+      if kernel_user in unauthorized_list:
+        return None
+    # Now we have check that unauthorized list doesnt exist
+    if exists(kernelspec_model, ['spec', 'metadata', 'process_proxy', 'config', 'authorized_users']):
+      authorized_list = kernelspec_model['spec']['metadata']['process_proxy']['config']['authorized_users']
+      if authorized_list and kernel_user not in authorized_list:
+        return None
+  return kernelspec_model
+
+class ModifyKernelSpecHandler (APIHandler):
+    @web.authenticated
+    @gen.coroutine
+    def get(self, kernel_user = None):
+        ksm = self.kernel_spec_manager
+        km = self.kernel_manager
+        model = {}
+        model['default'] = km.default_kernel_name
+        model['kernelspecs'] = specs = {}
+        kspecs = yield maybe_future (ksm.get_all_specs ())
+        if kernel_user:
+            self.log.info("Searching kernels for user '%s' " %kernel_user)
+        else:
+            self.log.info("No user. All kernels given")
+
+        for kernel_name, kernel_info in kspecs.items ():
+            try:
+                if is_kernelspec_model (kernel_info):
+                    d = kernel_info
+                else:
+                    d = kernelspec_model(self, kernel_name, kernel_info['spec'], kernel_info['resource_dir'])
+
+                d = apply_user_filter(d, kernel_user)
+                if d is not None:
+                    self.log.info("Find kernel '%s'", d['name'])
+                    specs[kernel_name] = d
+
+            except Exception:
+                self.log.error("Failed to load kernel spec: '%s'", kernel_name)
+                continue
+        self.set_header("Content-Type", 'application/json')
+        self.finish(json.dumps(model))
+
+
+kernel_name_regex = r"(?P<kernel_name>[\w\.\-%]+)"
+kernel_user_regex = r"(?P<kernel_user>[\w%]+)"
+
+default_handlers_notebooks = [
+            (r"/api/kernelspecs", ModifyKernelSpecHandler),
+            (r"/api/kernelspecs/user/%s" % kernel_user_regex, ModifyKernelSpecHandler),
+            (r"/api/kernelspecs/%s" % kernel_name_regex, notebook_handlers.KernelSpecHandler)
+]
+
+for path, cls in default_handlers_notebooks:
     bases = (TokenAuthorizationMixin, CORSMixin, JSONErrorsMixin, cls)
     default_handlers.append((path, type(cls.__name__, bases, {})))
 
 for path, cls in notebook_kernelspecs_resources_handlers.default_handlers:
     # Everything should have CORS and token auth
     bases = (TokenAuthorizationMixin, CORSMixin, JSONErrorsMixin, cls)
-    default_handlers.append((path, type(cls.__name__, bases, {})))
+    default_handlers.append ((path, type(cls.__name__, bases, {})))
+
--- a/enterprise_gateway/services/kernelspecs/handlers.py
+++ b/enterprise_gateway/services/kernelspecs/handlers.py
@@ -109,7 +109,6 @@ class KernelSpecHandler(TokenAuthorizationMixin,
 
 
 kernel_name_regex = r"(?P<kernel_name>[\w\.\-%]+)"
-kernel_user_regex = r"(?P<kernel_user>[\w%]+)"
 
 # Extends the default handlers from the notebook package with token auth, CORS
 # and JSON errors.
--- a/enterprise_gateway/services/kernelspecs/handlers.py
+++ b/enterprise_gateway/services/kernelspecs/handlers.py
@@ -108,7 +108,8 @@ def exists(obj, chain):
     if _key in obj:
         return exists(obj[_key], chain) if chain else obj[_key]
 
-def apply_user_filter(kernelspec_model, kernel_user = None):
+
+def apply_user_filter(kernelspec_model, kernel_user=None):
     if kernel_user:
         # Check unauthorized list
         if exists(kernelspec_model, ['spec', 'metadata', 'process_proxy', 'config', 'unauthorized_users']):
@@ -123,7 +124,8 @@ def apply_user_filter(kernelspec_model, kernel_user = None):
                 return None
     return kernelspec_model
 
-class ModifyKernelSpecHandler (APIHandler):
+
+class ModifyKernelSpecHandler(APIHandler):
     @web.authenticated
     @gen.coroutine
     def get(self, kernel_user = None):
@@ -134,7 +136,7 @@ class ModifyKernelSpecHandler (APIHandler):
         model['kernelspecs'] = specs = {}
         kspecs = yield maybe_future (ksm.get_all_specs ())
         if kernel_user:
-            self.log.info("Searching kernels for user '%s' " %kernel_user)
+            self.log.info("Searching kernels for user '%s' " % kernel_user)
         else:
             self.log.info("No user. All kernels given")
 
@@ -173,5 +175,5 @@ for path, cls in default_handlers_notebooks:
 for path, cls in notebook_kernelspecs_resources_handlers.default_handlers:
     # Everything should have CORS and token auth
     bases = (TokenAuthorizationMixin, CORSMixin, JSONErrorsMixin, cls)
-    default_handlers.append ((path, type(cls.__name__, bases, {})))
+    default_handlers.append((path, type(cls.__name__, bases, {})))
 
--- a/enterprise_gateway/services/kernelspecs/handlers.py
+++ b/enterprise_gateway/services/kernelspecs/handlers.py
@@ -48,9 +48,8 @@ class MainKernelSpecHandler(TokenAuthorizationMixin,
         kernel_user = None
         if kernel_user_filter:
             kernel_user = kernel_user_filter[0].decode("utf-8")
-
-        if kernel_user:
-            self.log.debug("Searching kernels for user '%s' " % kernel_user)
+            if kernel_user:
+                self.log.debug("Searching kernels for user '%s' " % kernel_user)
 
         kspecs = await maybe_future(ksm.get_all_specs())
 
--- a/enterprise_gateway/services/kernelspecs/handlers.py
+++ b/enterprise_gateway/services/kernelspecs/handlers.py
@@ -2,12 +2,10 @@
 # Distributed under the terms of the Modified BSD License.
 """Tornado handlers for kernel specs."""
 import json
-
 import notebook.kernelspecs.handlers as notebook_kernelspecs_resources_handlers
 from notebook.services.kernelspecs.handlers import is_kernelspec_model, kernelspec_model
 from notebook.utils import maybe_future, url_unescape
 from tornado import web
-
 from ...base.handlers import APIHandler
 from ...mixins import TokenAuthorizationMixin, CORSMixin, JSONErrorsMixin
 
--- a/enterprise_gateway/services/kernelspecs/handlers.py
+++ b/enterprise_gateway/services/kernelspecs/handlers.py
@@ -109,19 +109,19 @@ def exists(obj, chain):
         return exists(obj[_key], chain) if chain else obj[_key]
 
 def apply_user_filter(kernelspec_model, kernel_user = None):
-  if kernel_user:
-    # Check unauthorized list
-    if exists(kernelspec_model, ['spec', 'metadata', 'process_proxy', 'config', 'unauthorized_users']):
-      # Check if kernel_user in kernelspec_model
-      unauthorized_list = kernelspec_model['spec']['metadata']['process_proxy']['config']['unauthorized_users']
-      if kernel_user in unauthorized_list:
-        return None
-    # Now we have check that unauthorized list doesnt exist
-    if exists(kernelspec_model, ['spec', 'metadata', 'process_proxy', 'config', 'authorized_users']):
-      authorized_list = kernelspec_model['spec']['metadata']['process_proxy']['config']['authorized_users']
-      if authorized_list and kernel_user not in authorized_list:
-        return None
-  return kernelspec_model
+    if kernel_user:
+        # Check unauthorized list
+        if exists(kernelspec_model, ['spec', 'metadata', 'process_proxy', 'config', 'unauthorized_users']):
+            # Check if kernel_user in kernelspec_model
+            unauthorized_list = kernelspec_model['spec']['metadata']['process_proxy']['config']['unauthorized_users']
+            if kernel_user in unauthorized_list:
+                return None
+        # Now we have check that unauthorized list doesnt exist
+        if exists(kernelspec_model, ['spec', 'metadata', 'process_proxy', 'config', 'authorized_users']):
+            authorized_list = kernelspec_model['spec']['metadata']['process_proxy']['config']['authorized_users']
+            if authorized_list and kernel_user not in authorized_list:
+                return None
+    return kernelspec_model
 
 class ModifyKernelSpecHandler (APIHandler):
     @web.authenticated
--- a/enterprise_gateway/services/kernelspecs/handlers.py
+++ b/enterprise_gateway/services/kernelspecs/handlers.py
@@ -25,7 +25,7 @@ def url_unescape(path):
     Turns '/foo%20bar/' into '/foo bar/'
     """
     return u'/'.join([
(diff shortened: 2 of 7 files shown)


kevin-bates:
@lucabem - you are applying changes to a codebase older than master and, because of that, you're inadvertently rolling back recent changes.

Please rebase your master branch with that from EG, then update your dev branch with that rebased master.  No changes should be present for `processproxy.py`, `kernels/handlers.py`,  or `remotemanager.py`.

Also, you should not change the version value in `_version.py`.  We will update that when a new release is necessary.  Any dev build should have 2.2.0.dev0 as their version indicator.

Thanks.

lucabem:
Updated, I was pulled my other version.

We need to add a config param --GatewayClient.kernel_user_filter_env to know if we will filter kernels or not.

By default, it should be off (or other value that means _no filtering_

enterprise_gateway/services/kernelspecs/handlers.py
```diff
@@ -1,17 +1,175 @@
...
+                return None
+        # Now we have check that unauthorized list doesnt exist
+        if exists(kernelspec_model, ['spec', 'metadata', 'process_proxy', 'config', 'authorized_users']):
+            authorized_list = kernelspec_model['spec']['metadata']['process_proxy']['config']['authorized_users']
+            if authorized_list and kernel_user not in authorized_list:
+                return None
+    return kernelspec_model
+
+
+class ModifyKernelSpecHandler(APIHandler):
```
kevin-bates, line 129: I don't think the `Modify` prefix is the right thing here.  We could keep the names the same, or change the prefix to `User` perhaps?

enterprise_gateway/services/kernelspecs/handlers.py
```diff
@@ -1,17 +1,175 @@
...
+    @web.authenticated
+    @gen.coroutine
+    def get(self, kernel_user=None):
```
kevin-bates, line 132: Whenever adding new async methods we should use async/await constructs rather than @gen.coroutine/yield.  The general guideline is...
```python
    @gen.coroutine
    def get(self, kernel_user=None):
```
becomes...
```python
    async def get(self, kernel_user=None):
```
And `yield` becomes `await`.

enterprise_gateway/services/kernelspecs/handlers.py
```diff
@@ -1,17 +1,175 @@
...
+        ksm = self.kernel_spec_manager
+        km = self.kernel_manager
+        model = {}
+        model['default'] = km.default_kernel_name
+        model['kernelspecs'] = specs = {}
+        kspecs = yield maybe_future(ksm.get_all_specs())
+        if kernel_user:
+            self.log.info("Searching kernels for user '%s' " % kernel_user)
```
kevin-bates, line 140: Please change to `debug`.

enterprise_gateway/services/kernelspecs/handlers.py
```diff
@@ -1,17 +1,175 @@
...
+        else:
+            self.log.info("No user. All kernels given")
```
kevin-bates, line 142: Because this is the default, I don't think this should be logged at all.  It can be inferred if the other isn't logged (when DEBUG is enabled).

enterprise_gateway/services/kernelspecs/handlers.py
```diff
@@ -1,17 +1,175 @@
...
+            try:
+                if is_kernelspec_model(kernel_info):
+                    d = kernel_info
+                else:
+                    d = kernelspec_model(self, kernel_name, kernel_info['spec'], kernel_info['resource_dir'])
+
+                d = apply_user_filter(d, kernel_user)
+                if d is not None:
+                    self.log.info("Find kernel '%s'", d['name'])
+                    specs[kernel_name] = d
```
kevin-bates, line 154: Not sure this should be logged as it will produce LOTS of logging.  I think the returned set will be indicative enough once we have confidence the general logic is working.

enterprise_gateway/services/kernelspecs/handlers.py
```diff
@@ -1,17 +1,175 @@
...
+        self.finish(json.dumps(model))
+
+
+kernel_name_regex = r"(?P<kernel_name>[\w\.\-%]+)"
+kernel_user_regex = r"(?P<kernel_user>[\w%]+)"
+
+default_handlers_notebooks = [
+    (r"/api/kernelspecs", ModifyKernelSpecHandler),
+    (r"/api/kernelspecs/user/%s" % kernel_user_regex, ModifyKernelSpecHandler),
+    (r"/api/kernelspecs/%s" % kernel_name_regex, notebook_handlers.KernelSpecHandler)
```
kevin-bates, line 169: Although this is a supported endpoint (`/api/kernelspecs/<kernel-name>`), I'm not sure there's anything that calls it from Notebook.  However, it is supported.  Do we need a similar handler that allows a user to be specified?

enterprise_gateway/services/kernelspecs/handlers.py
lucabem: Yes, ModifyKernelSpecHandler doesn't represent anything.
We could change it to FilterKernelSpecHandler or UserKernelSpecHandler, both could be ok and more representative.

kevin-bates: Both seem fine, but 'Filter' might imply a general, multi-field, filtering mechanism, so I lean toward `UserKernelSpecHandler`.  Thanks.

enterprise_gateway/services/kernelspecs/handlers.py
```diff
@@ -1,17 +1,175 @@
...
+
 
 # Extends the default handlers from the notebook package with token auth, CORS
 # and JSON errors.
 default_handlers = []
-for path, cls in notebook_handlers.default_handlers:
-    # Everything should have CORS and token auth
+
+
+def exists(obj, chain):
```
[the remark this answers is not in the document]
lucabem: yes, I think its better to call it key_exists rather than exists

enterprise_gateway/services/kernelspecs/handlers.py
lucabem: I need to take a look to `/api/kernelspecs/<kernel-name` function.

enterprise_gateway/services/kernelspecs/handlers.py
lucabem: I would construct a list with kernel_names, what do you think?

kevin-bates: This endpoint is called very frequently.  I think with juptyerLab it's called periodically.  Therefore, I would prefer we minimize the logging produced by this method.  If you can defer any logging until the end of the method, and only log once, that would be better than twice per call (and only if user-filtering was applied).

enterprise_gateway/services/kernelspecs/handlers.py
```diff
@@ -1,17 +1,85 @@
...
 
+from tornado import web
 from ...mixins import TokenAuthorizationMixin, CORSMixin, JSONErrorsMixin
+from ...base.handlers import APIHandler
 
 # Extends the default handlers from the notebook package with token auth, CORS
 # and JSON errors.
 default_handlers = []
-for path, cls in notebook_handlers.default_handlers:
-    # Everything should have CORS and token auth
```
kevin-bates, line 14: Could you please preserve (and update) this comment at the end of file?  That code isn't very intuitive and I've found this helpful.  By 'update' I mean something like: `# Everything should have CORS, token auth, and JSON error handling` - thanks.

enterprise_gateway/services/kernelspecs/handlers.py
lucabem: `/api/kernelspecs/<kernel-name` returns a spec model of available kernels with given kernel name.

I my opinion, I think we dont need to have an specified handler because I guess that this `kernel-name` will have this user authorized.

Is this call made after calling /api/kernelspecs?

enterprise_gateway/services/kernelspecs/handlers.py
lucabem: Ok, if its called periodically is better to silence it or check if that call returns different models maybe.

enterprise_gateway/services/kernelspecs/handlers.py
kevin-bates: >Is this call made after calling /api/kernelspecs?

I'm not sure either notebook front-ends (notebook, lab) call this endpoint.  I've set breakpoints in this handler and they are not hit, iirc.

>I my opinion, I think we dont need to have an specified handler because I guess that this kernel-name will have this user authorized.

I don't think that's a valid assumption. If I call /api/kernelspecs/foo, yet I am not authorized to run kernels of this type, should I be able to access this kernelspec - if there was a means to enforce authorization on this endpoint like you're adding on the other?

This PR should also add the /user endpoint definition in the swagger files in https://github.com/jupyter/enterprise_gateway/tree/master/enterprise_gateway/services/api.  I can help you with that although we should probably wait to see how this discussion turns out (since we may need to add another handler).

kevin-bates:
Hi @lucabem - I've been reading more about REST APIs and when to use a path variable over a query parameter and its seems for things like filtering, a query parameter is the preferred approach.  I think this would also keep the REST endpoints as they are today and we just handle the case where there's a ?user=foo parameter.  

What do you think?

lucabem:
> Hi @lucabem - I've been reading more about REST APIs and when to use a path variable over a query parameter and its seems for things like filtering, a query parameter is the preferred approach. I think this would also keep the REST endpoints as they are today and we just handle the case where there's a ?user=foo parameter.
> 
> What do you think?

I think that if we this query param added to /api/kernelspecs allow us to avoid adding new paths to the API, it could be great!

Im am thinking how we could implement it. The pattern looks like that `/api/kernelspecs?user={user}`
All [Jupyter Notebook Server API](https://github.com/jupyter/jupyter/wiki/Jupyter-Notebook-Server-API#Kernelspecs-API) is built without query-params

kevin-bates:
Correct. So because EG has their own handlers (and we might want to go ahead and copy the /api/kernelspecs/<name> handler as well) we simply look for the query parameter.  I believe I've seen that kind of thing around somewhere, so I'll try to find it.

What will be really nice about this approach is that if I go and hit the CURRENT EG with `/api/kernelspecs?user=foo`, I still get back all the kernelspecs!  This means that the gateway client code doesn't need to handle the case of hitting an older versioned server.

Here's a quick hit that might be worth looking into (I'm sure there are plenty of others as well): https://stackoverflow.com/questions/10726486/tornado-url-query-parameters
Since EG must also support downlevel clients, passing no query parameter should also be tolerated, preferably by the same handler.

kevin-bates:
Hi @lucabem - I think this is all that is needed.  We don't need to even modify the regular expression or endpoint declaration!

```python
        user_filters = self.request.query_arguments.get('user')
        if user_filters:  # There could be multiple (e.g., ?user=foo&user=bar), only take the first
            user_filter = user_filters[0].decode("utf-8")  # convert from bytes to str
```

lucabem:
@kevin-bates - If I undesntand correctly, with this approach we dont need to modify the KernelSpecHandler header 
```
@web.authenticated
    async def get(self):
```
It sounds great that we avoid adding new paths to the API. Futhermore, we need to modify path `/api/kernelspecs/{kernel-name} ` to  `/api/kernelspecs/{kernel-name}?user={user}`
And the path of API will be: 
```
default_handlers_notebooks = [
    (r"/api/kernelspecs", UserKernelSpecHandler),
    (r"/api/kernelspecs/<kernel-spec>", OtherKernelSpecHandler)
]
```

lucabem:
Added new handler to /api/kernelspecs/{kernel-name} to avoid unauthorized-users to get kernel-spec model.

**I dont test yet**

lucabem:
Added new handler to /api/kernelspecs/{kernel-name} to avoid unauthorized-users to get kernel-spec model.

**I dont test it yet**

lucabem:
Added new handler to /api/kernelspecs/{kernel-name} to avoid unauthorized-users to get kernel-spec model.

**I dont test it yet**

enterprise_gateway/services/kernelspecs/handlers.py
```diff
@@ -70,16 +72,52 @@ async def get(self, kernel_user=None):
...
+        except KeyError:
+            raise web.HTTPError(404, u'Kernel spec %s not found' % kernel_name)
+        if is_kernelspec_model(spec):
+            model = spec
+        else:
+            model = kernelspec_model(self, kernel_name, spec.to_dict(), spec.resource_dir)
+        d = apply_user_filter(model, kernel_user)
+
+        if d is None:
+            self.log.debu("You shouldnt been doing that")
```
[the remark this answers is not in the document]
lucabem: Perfect! I didnt notice to take an HTTP error.

Changed!

enterprise_gateway/services/kernelspecs/handlers.py
```diff
@@ -70,16 +72,52 @@ async def get(self, kernel_user=None):
         self.finish(json.dumps(model))
 
 
+class UserInfoKernelSpecHandler(APIHandler):
```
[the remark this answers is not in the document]
lucabem: Changed!

enterprise_gateway/services/kernelspecs/handlers.py
```diff
@@ -40,13 +36,19 @@ def apply_user_filter(kernelspec_model, kernel_user=None):
 
 class UserKernelSpecHandler(APIHandler):
     @web.authenticated
```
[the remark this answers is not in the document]
lucabem: Changed!

enterprise_gateway/services/kernelspecs/handlers.py
```diff
@@ -70,16 +72,52 @@ async def get(self, kernel_user=None):
...
 ]
 
+# Extends the default handlers from the notebook package with token auth, CORS
+# and JSON errors.
+default_handlers = []
+
 for path, cls in default_handlers_notebooks:
+    # Everything should have CORS and token auth
     bases = (TokenAuthorizationMixin, CORSMixin, JSONErrorsMixin, cls)
     default_handlers.append((path, type(cls.__name__, bases, {})))
```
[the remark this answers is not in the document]
lucabem: I dont undestand when you said `can be removed by adding the mixins into class definitions`

enterprise_gateway/services/kernelspecs/handlers.py
```diff
@@ -95,10 +93,8 @@ async def get(self, kernel_name):
         d = apply_user_filter(model, kernel_user)
 
         if d is None:
-            self.log.debu("You shouldnt been doing that")
-            # User is not Authorized to know kernel
-            # specs
-            # Do something
+            raise web.HTTPError(403, u'User %s is not authorized to use this kernel spec %s'
```
kevin-bates, line 96: I think we should drop 'this' or add a separator like "kernel spec: %s" that "introduces the variable kernel spec name.

enterprise_gateway/services/kernelspecs/handlers.py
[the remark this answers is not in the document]
kevin-bates: No worries.  The code in this block exists to extend the older handlers (previously defined only in notebook) with the mixin classes.  Now that we're defining the handlers in EG, we can simply update the declarations of `MainKernelSpecHandler` and `KernelSpecHandler` to include the mixin class definitions.  The result will look extremely similar to what we do in [`MainKernelHandler`](https://github.com/jupyter/enterprise_gateway/blob/master/enterprise_gateway/services/kernels/handlers.py#L15-L17) expect we're not overriding the notebook kernelspec handlers, we're replacing them.

enterprise_gateway/services/kernelspecs/handlers.py
lucabem: Changed to `User %s is not authorized to use kernel spec %s`

lucabem:
Hi @kevin-bates - For testing on Kubernetes Cluster, should I change `async/awai`t to `@gen.coroutine/yield`?

kevin-bates:
Hmm - are you seeing an error like: "<method> was not awaited"  somewhere?  If so can you provide the traceback and message?  Also, what notebook version are you using?  Is it a build from the master branch?  I may have stumbled over something similar. Thanks.

lucabem:
I didnt try it today. Tomorrow I will try (making changes you refer on [issue ](https://github.com/jupyter/enterprise_gateway/issues/806))

kevin-bates:
Please don't.  I'm running into some issues (with that issue) and I think we're going to need to pin Notebook to < 6.1 (knowing 6.1 is the async release).  Then, when we add support in EG for async kernel management, and given that Notebook 6.1 will be available, we'll pin EG to >= 6.1.

However, I'd like to understand your current issue because I'm still trying to understand what's going on.  If you could answer the questions, that would be appreciated.  If you don't have the error message handy, that's fine (although I need to see it).  I suspect you're building your own Notebook from the master branch - which contains the changes that are destined for NB 6.1.

For you to move forward, I recommend you run YOUR dev EG on top of Notebook 6.0.3 and YOUR dev Notebook client on master and you should be okay.  This will mean that the two envs (server, client) will likely need to be done in separate python/conda envs.

lucabem:
Perfect! I was working with notebook 6.0.0 so maybe thats why I m not habing issues. In addition, I am working with LocalProccessProxy.

lucabem:
Hi @kevin-bates - Shall we use an cofig param to allow kernel filtering as we have talked about?

kevin-bates:
I'd rather not.  I view the existence of the query parameter as the trigger.  If its not there, the logic isn't used.  We'll need a config parameter for the client side and we were going to use a parameter that specifies the name of the env variable from which to grab the user's name from.

Hmm.  I wonder if we even need that?  Seems like we could say if `KERNEL_USERNAME` is set, use that.  If folks hit an old EG, it should just work since that EG will ignore the query parameter that will be present.  If folks hit an updated EG, they'll still get all their kernelspecs **unless** they're already using the {un}authorized_users feature, in which case, they would want filtering performed on the fetch side of things since they're already applying authorization on kernel startup.

Let me know when this is ready for another review - thank you.

lucabem:
Perfect! - In addition I am working to allow JEG to use conda environments on other branch. It maybe useful in a future.

lucabem:
Hi @kevin-bates - Could you review the code from last commit? I have removed one debug statement in case we dont recive a user query param.

Should we debug [it](https://github.com/lucabem/enterprise_gateway/blob/user-filter/enterprise_gateway/services/kernelspecs/handlers.py#L53)?

enterprise_gateway/services/kernelspecs/handlers.py
```diff
@@ -52,8 +51,8 @@ async def get(self):
 
         if kernel_user:
```
kevin-bates, line 52: Let's move this `if` block into the one above since that's the only way `kernel_user` can have a value.

Should we check for a `None` `kernel_user` in that block?  Not sure if its possible to send something like: `GET /api/kernelspecs?user` or `GET /api/kernelspecs?user=''`, so just wondering if we should raise an error if `kernel_user` is `None` or has length == 0 when the query parameter is present?

lucabem: If we dont check if kernel_user is not empty, a call from `/api/kernelspecs?user=` will return all specs.

Is the same call `/api/kernelspecs?user=` and `/api/kernelspecs`

lucabem: We just check if kernel_user is not an empty string. If we call `/api/kernelspecs?user=''` or `/api/kernelspecs?user` , it is the same as `/api/kernelspecs`

lucabem:
Hi @kevin-bates, Shall I modify swagger files? Or you will do it

kevin-bates:
That sounds great. Please go ahead with that. Thanks.

lucabem:
@kevin-bates, could you check last commit in swagger.yaml? 

I have seen that in this file we dont have `/api/kernelspecs/{kernel-name}`

kevin-bates:
Your commit looks great - thank you.

Regarding the missing kernelspec endpoint, you're right.  I figured I had not brought it over from notebook when modifying these but I see that notebook only has the [one entry also](https://github.com/jupyter/notebook/blob/master/notebook/services/api/api.yaml#L509-L526).

Since we don't really see where that particular endpoint is used, I wonder if it was a late addition and the swagger docs just didn't get updated.  Seems like we should add that in as well, but I'll leave that up to you.  If you wanted to also add it to Notebook, I can help get it merged via a notebook PR.

lucabem:
Perfect! - I will open a pull requet on notebook to allow send that query param

kevin-bates:
Just to clarify.  If you open a PR on the notebook swagger docs, you should not reference the query parameter.  That only applies to the EG swagger.  However, we will need to update the Gateway code within Notebook to use the query parameter - if that's what you're referring to.

Is this PR ready for final review?

lucabem:
Yes, it is ready.

PR on Notebook is for updating Gateway code, not swagger

enterprise_gateway/services/kernelspecs/handlers.py
```diff
@@ -1,19 +1,117 @@
...
+        for kernel_name, kernel_info in kspecs.items():
+            try:
+                if is_kernelspec_model(kernel_info):
+                    d = kernel_info
+                else:
+                    d = kernelspec_model(self, kernel_name, kernel_info['spec'], kernel_info['resource_dir'])
+                d = apply_user_filter(d, kernel_user)
+                if d is not None:
+                    specs[kernel_name] = d
+                    list_kernels_found.append(d['name'])
```
kevin-bates, line 66: I think it would be good to add a DEBUG message in an `else` block for the unauthorized case.  This is because when debug is enabled, EG will produce the following kind of output...
```
[D 2020-04-28 11:55:46.982 EnterpriseGatewayApp] Searching kernels for user 'scala' 
[D 2020-04-28 11:55:46.983 EnterpriseGatewayApp] Found kernel python3 in /opt/anaconda2/envs/py3/share/jupyter/kernels
[D 2020-04-28 11:55:46.983 EnterpriseGatewayApp] Found kernel ir in /opt/anaconda2/envs/py3/share/jupyter/kernels
[D 2020-04-28 11:55:46.984 EnterpriseGatewayApp] Found kernel dask_python_yarn_remote in /usr/local/share/jupyter/kernels
[D 2020-04-28 11:55:46.984 EnterpriseGatewayApp] Found kernel spark_r_yarn_client in /usr/local/share/jupyter/kernels
[D 2020-04-28 11:55:46.984 EnterpriseGatewayApp] Found kernel spark_r_yarn_cluster in /usr/local/share/jupyter/kernels
[D 2020-04-28 11:55:46.984 EnterpriseGatewayApp] Found kernel spark_python_yarn_client in /usr/local/share/jupyter/kernels
[D 2020-04-28 11:55:46.984 EnterpriseGatewayApp] Found kernel spark_python_yarn_cluster in /usr/local/share/jupyter/kernels
[D 2020-04-28 11:55:46.984 EnterpriseGatewayApp] Found kernel spark_scala_yarn_client in /usr/local/share/jupyter/kernels
[D 2020-04-28 11:55:46.984 EnterpriseGatewayApp] Found kernel spark_scala_yarn_cluster in /usr/local/share/jupyter/kernels
```
In my example, user `scala` is in the unauthorized users list for the python-related YARN kernels and the proper filtering is applied.  However, because the "Found kernel" message is produced from `jupyter_client` this output can be misleading.

I think a DEBUG statement along the lines of: "User '%s' is not authorized to use kernel spec %s, excluding from results." would be helpful. 

Here's what the above would then look like...
```
[D 2020-04-28 11:55:46.982 EnterpriseGatewayApp] Searching kernels for user 'scala' 
[D 2020-04-28 11:55:46.983 EnterpriseGatewayApp] Found kernel python3 in /opt/anaconda2/envs/py3/share/jupyter/kernels
[D 2020-04-28 11:55:46.983 EnterpriseGatewayApp] Found kernel ir in /opt/anaconda2/envs/py3/share/jupyter/kernels
[D 2020-04-28 11:55:46.984 EnterpriseGatewayApp] Found kernel dask_python_yarn_remote in /usr/local/share/jupyter/kernels
[D 2020-04-28 11:55:46.984 EnterpriseGatewayApp] User 'scala' is not authorized to use kernel spec dask_python_yarn_remote, excluding from results.
[D 2020-04-28 11:55:46.984 EnterpriseGatewayApp] Found kernel spark_r_yarn_client in /usr/local/share/jupyter/kernels
[D 2020-04-28 11:55:46.984 EnterpriseGatewayApp] Found kernel spark_r_yarn_cluster in /usr/local/share/jupyter/kernels
[D 2020-04-28 11:55:46.984 EnterpriseGatewayApp] Found kernel spark_python_yarn_client in /usr/local/share/jupyter/kernels
[D 2020-04-28 11:55:46.984 EnterpriseGatewayApp] User 'scala' is not authorized to use kernel spec spark_python_yarn_client, excluding from results.
[D 2020-04-28 11:55:46.984 EnterpriseGatewayApp] Found kernel spark_python_yarn_cluster in /usr/local/share/jupyter/kernels
[D 2020-04-28 11:55:46.984 EnterpriseGatewayApp] User 'scala' is not authorized to use kernel spec spark_python_yarn_cluster, excluding from results.
[D 2020-04-28 11:55:46.984 EnterpriseGatewayApp] Found kernel spark_scala_yarn_client in /usr/local/share/jupyter/kernels
[D 2020-04-28 11:55:46.984 EnterpriseGatewayApp] Found kernel spark_scala_yarn_cluster in /usr/local/share/jupyter/kernels
```

enterprise_gateway/services/api/swagger.yaml
```diff
@@ -86,9 +86,17 @@ paths:
       summary: Get kernel specs
       tags:
         - kernelspecs
+      parameters:
+        - name: user
+          in: query
+          description: User value that need to be considered for filtering kernels
```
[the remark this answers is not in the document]
lucabem: Perfect, changed

enterprise_gateway/services/kernelspecs/handlers.py
```diff
@@ -1,19 +1,117 @@
...
 import notebook.kernelspecs.handlers as notebook_kernelspecs_resources_handlers
-
+from notebook.services.kernelspecs.handlers import is_kernelspec_model, kernelspec_model
+from notebook.utils import maybe_future, url_unescape
+from tornado import web
+from ...base.handlers import APIHandler
 from ...mixins import TokenAuthorizationMixin, CORSMixin, JSONErrorsMixin
 
+
+def key_exists(obj, chain):
```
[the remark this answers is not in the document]
lucabem: Changed!

lucabem:
Tomorrow I will update swagger json!

lucabem:
@kevin-bates - I have already updated swagger.json

lucabem:
Hi @kevin-bates - Should we wait until notebook 7.0.0 for opening PR to allow send user query param?

kevin-bates:
No - we plan on building a 6.1 release soon (I need to create an issue for that in fact), and I'd like this to get in there.  We can merge to master for the time being and have it picked up in 6.1.  

I'm hoping we can forgo the ENV indicating from where to pull the user value and always use `KERNEL_USERNAME` if it has a value.  Since all kernelspecs are returned if the authorization isn't configured in EG and since the query parameter is ignored on older APIs, it doesn't hurt to just do it all the time.

lucabem:
Yes, we could use KERNEL_USERNAME (it doesnt matter where his value comes from).

Before the user could modify this value setting KERNEL_USERNAME or HttpClient.user

lucabem:
> I'm hoping we can forgo the ENV indicating from where to pull the user value and always use KERNEL_USERNAME if it has a value. Since all kernelspecs are returned if the authorization isn't configured in EG and since the query parameter is ignored on older APIs, it doesn't hurt to just do it all the time.

Could it be like that?

    @gen.coroutine
    def get_kernel_spec(self, kernel_name, **kwargs):
        kernel_spec_url = '?user='.join([self._get_kernelspecs_endpoint_url(kernel_name=str(kernel_name)),
                                         str(os.environ['KERNEL_USERNAME'])])
        self.log.debug("Request kernel spec at: %s" % kernel_spec_url)
        try:
            response = yield gateway_request(kernel_spec_url, method='GET')
        except web.HTTPError as error:
            if error.status_code == 404:
                raise KeyError('kernelspec {kernel_name} not found on Gateway server at: {gateway_url}'.
                               format(kernel_name=kernel_name, gateway_url=GatewayClient.instance().url))
            else:
                raise
        else:
            kernel_spec = json_decode(response.body)

        raise gen.Return(kernel_spec)

kevin-bates:
That won't work if KERNEL_USERNAME is not set.  I would also recommend using `os.getenv()` instead.  Then you won't get a KeyError if not set.

We can only add the `?user=` clause if there's a value with length > 0.

Also, please reference this PR when opening your Notebook PR.  All changes must be within the gateway package as well - just a reminder.  Thanks!

lucabem:
Perfect! 

I will try to open It this weekend

kevin-bates:
@lucabem - that would be great.  We're hoping to get a bunch of Notebook PRs reviewed today and if you could open a PR this weekend, then we could get it into the next NB release.  Thank you!

kevin-bates:
Hi @lucabem - I was about to merge this and saw that this was still present and should be removed: https://github.com/jupyter/enterprise_gateway/blob/22b09d0f3c553a559900427b8b642fdf8e59d3cb/enterprise_gateway/services/kernelspecs/handlers.py#L112

Do you mind removing that line?  Also, I will be squashing the commits and will retain only the meaningful comments (as some commits were essentially reverted).  If you'd like to squash the commits yourself in order to have a better comment, please feel free to do so.  I'd like to merge this in the next day or two.

Looking forward to your update to the notebook gateway code.

Thanks.

kevin-bates:
@lucabem - would you mind rebasing your branch to master?  Once conflicts are resolved I'll merge this.  Thank you.

lucabem:
Sorry for responding late, I have been busy this week

Tomorrow I would rebase it!

lucabem:
Hi @kevin-bates! Everything is done.

Could you check it? I havent worked with rebase until today

kevin-bates:
There are still conflicts.  Please try the following...

1. fetch the appropriate remote name for the repo.  Mine is called 'upstream'.  Typical values are 'upstream' or 'origin', although many associate 'origin' to their fork.
```
git fetch upstream
```
2. I usually update my working master branch with upstream, then push to my origin's master:
```
git checkout master
git rebase upstream/master
git push origin master
```
3. Now that your working master reflects the EG repo, merge it with your branch...
```
git checkout user-filter
git merge --no-ff master
```
You should have a commit commented with "Merge branch 'master' into user-filter".
4. Push your branch or your remote name (mine would be origin) - which should update the PR
```
git push origin user-filter
```

GitHub should indicate the branch can be merged as conflicts are no longer present.

lucabem:
I have done what you have said and it looks like there is no conflicts with bas branch

kevin-bates:
Ok - I'm going to squash your commits unless you'd like to do that yourself.

lucabem:
Perfect! Im little bit busy (examns and proyects), feel free to do it

kevin-bates:
Merged - thank you very much for this excellent contribution @lucabem.  Good luck on your exams!

Issue #803, opened 2020-04-22
Making RemoteKernelManager more independent


golf-player:
Original thread: https://github.com/jupyter/nbclient/issues/26

Some config-related issues that crop up using RKM (RemoteKernelManager) for one-off executions with nbclient/nbconvert:

- RKM depends on `kernel_id` for pod name and connection file. As discussed on the other thread, the concept of `kernel_id` is only in the context of MultiKernelManager, so maybe there could be an option for oneoffs to make essentially the same thing as a kernel_id, but call it differently to avoid confusion?
- To make RKM usable without a JEG app or kernel_manager_factory, there's config that needs to be added that is common to JEG app and RemoteMappingKernelManager

@kevin-bates suggested using the kernel provider architecture here, but that seems to not be accepted, and there'd be some work to be done to get RKM to use the kernel provider. I'm gonna read through kernel providers to get some kind of understanding of how this would work.

Pasting in the code from the other thread here.
```
import os
import nbformat
from nbclient import NotebookClient
from enterprise_gateway.services.kernels.remotemanager import RemoteKernelManager, RemoteMappingKernelManager
from enterprise_gateway.enterprisegatewayapp import EnterpriseGatewayApp

eg = EnterpriseGatewayApp()
rmkm = RemoteMappingKernelManager(parent=eg,
                                  log=eg.log,
                                  connection_dir=eg.runtime_dir,)


class RKM(RemoteKernelManager):
    def __init__(self, **kwargs):
        kernel_id = rmkm.new_kernel_id()
        kwargs.update({
            "parent": rmkm,
            "log": rmkm.log,
            "kernel_name": "python_kubernetes",
            "connection_file": os.path.join(rmkm.connection_dir, "kernel-%s.json" % kernel_id)
        })
        super(RKM, self).__init__(**kwargs)
        self.kernel_id = kernel_id


with open("/home/ish/notebooks/Untitled.ipynb") as fp:
    test_notebook = nbformat.read(fp, as_version=4)

client = NotebookClient(nb=test_notebook, kernel_manager_class=RKM)
```

For me, ideally this would work by just setting the kernel_manager_class=RemoteKernelManager without the weird config stuff I'm forced to do otherwise.

[...]

PR #794, opened 2020-03-20
Add support for async kernel management

kevin-bates:
This is a redo of PR #580 due to branch changes - the description is essentially the same, however.

Updated RemoteKernelManager's methods to async/await and derive from jupyter_client's `AsyncIOLoopKernelManager` and `RemoteMappingKernelManager` now derives from notebook's `AsyncMappingKernelManager`.  As a result, this PR will NOT be back-ported to 1.x. 

**NOTE: Installation of this PR requires that BOTH related [Jupyter Client 428](https://github.com/jupyter/jupyter_client/pull/428) and [Notebook 4479](https://github.com/jupyter/notebook/pull/4479) PRs be merged.  Once merged, we need to update our dependencies to depend on those versions - thus the `[DEPENDENCY]` in the title.**

Fixes #86 
Fixes #592 

The PRs referenced above have been updated to reflect the recent PRs submitted that use a subclass-based approach.  Please see those PRs for a description the class hierarchy.

PR #810, opened 2020-04-24
First pass at making RemoteKernelManager independent

golf-player:
This is a WiP.

So first pass, I just wanted to get things working, and they do (with nbclient, not sure if gatewayapp still works, but it should). There's way too much asking for permission in the code right now, and there's probably a lot that could be done to separate code where mappingkernelmanagers are needed. Another option is to make a single kernel mapping manager (AKA a completely useless stupid class) just for convenience.

What do you think about doing traitlets like this? To properly share it with EG App, I'll need to add this mixin and remove the included traitlets from EG App.

### DIFF
--- a/docs/source/kernel-library.md
+++ b/docs/source/kernel-library.md
@@ -1,0 +1,22 @@
+## Standalone Remote Kernel Execution 
+##### a.k.a "Library Mode"
+
+Remote kernels can be executed by using the `RemoteKernelManager` class directly. This enables running kernels using `ProcessProxy`s without the Enterprise Gateway Webapp.
+
+This can be useful in niche situations, for example, using [nbconvert](https://nbconvert.readthedocs.io/) or [nbclient](https://nbclient.readthedocs.io/) to execute a kernel on a remote cluster.
+
+Sample code using nbclient 0.2.0:
+
+```python
+import nbformat
+from nbclient import NotebookClient
+from enterprise_gateway.services.kernels.remotemanager import RemoteKernelManager
+
+with open("my_notebook.ipynb") as fp:
+    test_notebook = nbformat.read(fp, as_version=4)
+
+client = NotebookClient(nb=test_notebook, kernel_manager_class=RemoteKernelManager)
+client.execute()
+```
+
+The above code will execute the notebook on a kernel using the configured `ProcessProxy` (defaults to Kubernetes).
--- a/enterprise_gateway/enterprisegatewayapp.py
+++ b/enterprise_gateway/enterprisegatewayapp.py
@@ -12,8 +12,6 @@ import sys
 import time
 import weakref
 
-from distutils.util import strtobool
-
 # Install the pyzmq ioloop. This has to be done before anything else from
 # tornado is imported.
 from zmq.eventloop import ioloop
@@ -21,13 +19,11 @@ ioloop.install()
 
 from tornado import httpserver
 from tornado import web
-from tornado.log import enable_pretty_logging, LogFormatter
+from tornado.log import enable_pretty_logging
 
-from traitlets import default, List, Set, Unicode, Type, Instance, Bool, CBool, Integer, observe
 from traitlets.config import Configurable
 from jupyter_core.application import JupyterApp, base_aliases
 from jupyter_client.kernelspec import KernelSpecManager
-from notebook.services.kernels.kernelmanager import MappingKernelManager
 from notebook.notebookapp import random_ports
 from notebook.utils import url_path_join
 
@@ -39,9 +35,11 @@ from .services.kernels.handlers import default_handlers as default_kernel_handle
 from .services.kernelspecs.handlers import default_handlers as default_kernelspec_handlers
 from .services.sessions.handlers import default_handlers as default_session_handlers
 
-from .services.sessions.kernelsessionmanager import KernelSessionManager, FileKernelSessionManager
+from .services.sessions.kernelsessionmanager import FileKernelSessionManager
 from .services.sessions.sessionmanager import SessionManager
 from .services.kernels.remotemanager import RemoteMappingKernelManager
+
+from .mixins import EnterpriseGatewayConfigMixin
 
 
 # Add additional command line aliases
@@ -56,7 +54,7 @@ aliases.update({
 })
 
 
-class EnterpriseGatewayApp(JupyterApp):
+class EnterpriseGatewayApp(EnterpriseGatewayConfigMixin, JupyterApp):
     """Application that provisions Jupyter kernels and proxies HTTP/Websocket
     traffic to the kernels.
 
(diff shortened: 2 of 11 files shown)


kevin-bates:
Thanks for submitting the pull request.  

I'd rather not take the trailet approach as this will break existing configurations.  Is there an issue with a property/accessor approach?

> There's way too much asking for permission in the code right now

Can you elaborate on this?  Are you referring to `if` statements?

> Another option is to make a single kernel mapping manager (AKA a completely useless stupid class) just for convenience.

You're welcome to use something like this in your wrapper code, but probably doesn't belong in EG.  So the idea would be to set a placeholder for `parent` and `parent.parent` items?

I think the kernelspec issue you brought up on #803 is something we need to address before proceeding.

golf-player:
I guess the only problem with the property approach is that it's a lot of repeated code. How does using traitlets like this break configuration? I'm not too familiar with traitlets.

> Can you elaborate on this? Are you referring to if statements?
Yeah basically there's too many points where we're checking if a property exists before doing something, when IMO there should be a clear divide between things that use a MultiKernelManager (MKM) and not (of course, this may be less useful if my use case isn't a common one). Or maybe that code belongs in the kernel_manager class.

> So the idea would be to set a placeholder for parent and parent.parent items?

The idea there would be to continue as is, removing parent and parent.parent items. `parent` is an MKM, so making something with the same interface, which basically is no-op on things like kernel connection count increments and then setting that as a property on RKM, which in this pull request is `RKM.multi_kernel_manager`. that code could live anywhere, true.

> I think the kernelspec issue you brought up on #803 is something we need to address before proceeding.

The problem being that the processproxy RKM uses is ignoring config from the kernelspec, right?

golf-player:
One thing I don't like is calling out to uuid.uuid4, but I see that done in several other places as well. Do you think it would make sense to make a function for everything to use, or keep it here, since technically the way RKM sets kernel_id should be agnostic of everything else.

kevin-bates:
Do you mind moving the conversation back to the issue?  I'm getting scattered with a bunch of other stuff going on and I'm kinda wondering if we should try to jump on a web session to get on the same page here.  Would you be available for something like that? If so, please drop me a note on my github email and I'll forward you a link.

golf-player:
I was going to ask about switching back to the issue haha.

I'll send you an email

enterprise_gateway/services/kernels/remotemanager.py
```diff
@@ -12,9 +12,11 @@
 from ipython_genutils.importstring import import_item
 from notebook.services.kernels.kernelmanager import MappingKernelManager
 from jupyter_client.ioloop.manager import IOLoopKernelManager
+from traitlets import log as traitlets_log
 
 from ..processproxies.processproxy import LocalProcessProxy, RemoteProcessProxy
 from ..sessions.kernelsessionmanager import KernelSessionManager
+from enterprise_gateway.mixins import EnterpriseGatewayConfigMixin
```
golf-player, line 19: Do y'all have an opinion on relative vs absolute imports? Mixing is bad, so I'll probably change this. That said, I really don't like the `...mixins` import

enterprise_gateway/services/kernels/remotemanager.py
```diff
@@ -418,8 +434,30 @@ def _get_process_proxy(self):
            instantiated.
         """
         process_proxy_cfg = get_process_proxy_config(self.kernel_spec)
+        print(self.kernel_spec)
         process_proxy_class_name = process_proxy_cfg.get('class_name')
         self.log.debug("Instantiating kernel '{}' with process proxy: {}".
                        format(self.kernel_spec.display_name, process_proxy_class_name))
         process_proxy_class = import_item(process_proxy_class_name)
         self.process_proxy = process_proxy_class(kernel_manager=self, proxy_config=process_proxy_cfg.get('config'))
+
```
golf-player, line 440: So looking into the code, seems kernelspec manager is already handled by the base class of RKM, and it gets invoked the same way as the one attached to RMKM

golf-player:
TODO: Fix tests, add tests, squash into nice commit

golf-player:
OK got it to work with mixins. A tag hacky maybe.

enterprise_gateway/services/kernels/remotemanager.py
```diff
@@ -325,26 +355,28 @@ def restart_kernel(self, now=False, **kwargs):
...
         """
         self.restarting = True
-        kernel_id = os.path.basename(self.connection_file).replace('kernel-', '').replace('.json', '')
+        kernel_id = self.kernel_id or os.path.basename(self.connection_file).replace('kernel-', '').replace('.json', '')
         # Check if this is a remote process proxy and if now = True. If so, check its connection count. If no
         # connections, shutdown else perform the restart.  Note: auto-restart sets now=True, but handlers use
         # the default value (False).
-        if isinstance(self.process_proxy, RemoteProcessProxy) and now:
-            if self.parent._kernel_connections.get(kernel_id, 0) == 0:
+        if isinstance(self.process_proxy, RemoteProcessProxy) and now and self.mapping_kernel_manager:
```
kevin-bates, line 362: Good catch on mapping_kernel_manager's potential for null.

enterprise_gateway/services/kernels/remotemanager.py
```diff
@@ -423,3 +455,24 @@ def _get_process_proxy(self):
                        format(self.kernel_spec.display_name, process_proxy_class_name))
         process_proxy_class = import_item(process_proxy_class_name)
         self.process_proxy = process_proxy_class(kernel_manager=self, proxy_config=process_proxy_cfg.get('config'))
+
+    @property
```
kevin-bates, line 459: I think we should add a comment block somewhere (thinking prior to these properties) that talks about when `self.parent` would not exist.  This might warrant an entry in the docs as well - although I don't mind keeping this on the down-low for the time being. :wink:

enterprise_gateway/services/kernels/remotemanager.py
golf-player: I retract that actually. Mixing is guaranteed to happen when you're using relative imports mostly given some of them are going to need a lot of dots. It's a cosmetic concern at worst.

enterprise_gateway/enterprisegatewayapp.py
```diff
@@ -56,7 +54,7 @@
 })
 
 
-class EnterpriseGatewayApp(JupyterApp):
+class EnterpriseGatewayApp(JupyterApp, EnterpriseGatewayConfigMixin):
```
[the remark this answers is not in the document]
golf-player: Good catch.

enterprise_gateway/services/kernels/remotemanager.py
```diff
@@ -224,6 +233,27 @@ def __init__(self, **kwargs):
         if hasattr(self, "cache_ports"):
             self.cache_ports = False
 
+        if not self.connection_file and not self.kernel_id:
```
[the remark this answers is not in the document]
golf-player: good catch. I'll remove the line setting it to None since it doesn't really help.

enterprise_gateway/services/kernels/remotemanager.py
golf-player: Will do. Is there somewhere specific in the docs where something like this might go?

enterprise_gateway/services/kernels/remotemanager.py
```diff
@@ -224,6 +233,27 @@ def __init__(self, **kwargs):
...
+            return
+        dependent_props = ["authorized_users",
+                           "unauthorized_users",
+                           "port_range",
+                           "impersonation_enabled",
+                           "max_kernels_per_user",
+                           "env_whitelist",
+                           "env_process_whitelist",
+                           ]
+        self._links = [directional_link((eg_instance, prop), (self, prop)) for prop in dependent_props]
```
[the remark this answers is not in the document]
golf-player: This is a traitlets thing actually. Any reason to not be a fan? I kind of like the ability to explicitly tie certain attributes to an upstream source.

enterprise_gateway/services/kernels/remotemanager.py
kevin-bates: I'm okay deferring this for now, although if we were to add it, the following locations might be worthy of updates:
- We should add an entry (or two?) to [Use Cases](https://jupyter-enterprise-gateway.readthedocs.io/en/latest/use-cases.html#use-cases)
- Perhaps add the details to the end of the deployments section (following [IBM Spectrum Conductor](https://jupyter-enterprise-gateway.readthedocs.io/en/latest/kernel-conductor.html#ibm-spectrum-conductor))

I'm trying to think of what we'd call the topic.  The term "Library Mode" popped into my head just now.  I'm not enamored with that, but I'd like the topic to convey a more direct capability that utilizes the heart of EG.  "Integration Mode" - something that allows third-parties to use RMK directly.  Ideas welcome! 

The topic should include an example - much like the one you opened the issue with - based on the final results.  (Btw, I'd like to have those details anyway so I could try things out against my Hadoop YARN cluster.)

golf-player: I like library mode. Just anything that conveys the meaning that this is JEG but without the gateway webapp.

Standalone Remote Kernel Execution if verbosity isn't an issue. 

The example actually becomes really simple.

```
import nbformat
from nbclient import NotebookClient
from enterprise_gateway.services.kernels.remotemanager import RemoteKernelManager

with open("/home/ish/notebooks/Untitled.ipynb") as fp:
    test_notebook = nbformat.read(fp, as_version=4)

client = NotebookClient(nb=test_notebook, kernel_manager_class=RemoteKernelManager)
client.execute()
```

If you'd like a full docker with kernelspecs and config, I can do that, but I don't have experience with Yarn so it may take a while.

kevin-bates: I like "Standalone Remote Kernel Execution" as the topic.  Perhaps in the opening statement, we can toss in an "a.k.a., Library Mode". :smile: This way we have a way to quickly identify the context when communicating this functionality or discussing it in general.

The sample is excellent - I'm really happy with how this is turning out.  Let's not worry about a docker image for the moment. Given the example, I can run this against my YARN cluster, although probably not today (sorry).

kevin-bates:
Hi - I finally got around to trying this, and I'm not able to make it off the launch pad.  Using nbclient 0.1.0 (0.2.0 switches to async kernel management, which EG isn't able to move to yet due to notebook), I get the following because `self.kernel_spec_manager` is None. 
```
python -m indieRMK.py 
  File "/opt/anaconda2/envs/py3/lib/python3.6/runpy.py", line 183, in _run_module_as_main
    mod_name, mod_spec, code = _get_module_details(mod_name, _Error)
  File "/opt/anaconda2/envs/py3/lib/python3.6/runpy.py", line 109, in _get_module_details
    __import__(pkg_name)
  File "/home/gateway/indieRMK.py", line 9, in <module>
    client.execute()
  File "/opt/anaconda2/envs/py3/lib/python3.6/site-packages/nbclient/client.py", line 357, in execute
    with self.setup_kernel(**kwargs):
  File "/opt/anaconda2/envs/py3/lib/python3.6/contextlib.py", line 81, in __enter__
    return next(self.gen)
  File "/opt/anaconda2/envs/py3/lib/python3.6/site-packages/nbclient/client.py", line 339, in setup_kernel
    self.start_new_kernel_client(**kwargs)
  File "/opt/anaconda2/envs/py3/lib/python3.6/site-packages/nbclient/client.py", line 312, in start_new_kernel_client
    self.km.start_kernel(extra_arguments=self.extra_arguments, **kwargs)
  File "/opt/anaconda2/envs/py3/lib/python3.6/site-packages/enterprise_gateway/services/kernels/remotemanager.py", line 274, in start_kernel
    self._get_process_proxy()
  File "/opt/anaconda2/envs/py3/lib/python3.6/site-packages/enterprise_gateway/services/kernels/remotemanager.py", line 458, in _get_process_proxy
    process_proxy_cfg = get_process_proxy_config(self.kernel_spec)
  File "/opt/anaconda2/envs/py3/lib/python3.6/site-packages/jupyter_client/manager.py", line 84, in kernel_spec
    self._kernel_spec = self.kernel_spec_manager.get_kernel_spec(self.kernel_name)
AttributeError: 'NoneType' object has no attribute 'get_kernel_spec'
```
Are you using a config file of any sorts to set that up?  I was able to run EG fine with this build, but get this in 'library mode'.  My driver is essentially identical to your [example code](https://github.com/jupyter/enterprise_gateway/pull/810#discussion_r418412886).

Also, I think we're going to want some form of documentation for this.  We're you still planning on adding that?

golf-player:
@kevin-bates apologies for getting back to you so late. Got wrapped up in another project. I will have time this week to debug this issue. I'm guessing it's the nbclient version difference; I'm using the latest nbclient (which works I guess surprisingly, given JEG isn't async)

What's the timeline for the JEG deploy btw? I may be able to reprioritize things depending in order to have this ready by then if possible.

I will be writing documentation for this.

kevin-bates:
Hi @golf-player - thanks for the update.  Regarding the timeline, as soon as Notebook 6.1 is available, we will merge #794 and start building the next EG release.  I'm hopeful that some of the front-end notebook issues will be reviewed/merged in the next week or two so that release can move forward.

I would love to get this into the next EG release, so, should things get to that point, I don't mind waiting to ensure this is ready, although I don't expect that to be necessary.

golf-player:
Looks like nbclient 0.3.0 also just dropped. I've gotten both 0.3.0 and 0.1.0 to _run_ the kernel, but there's another issue creeping up on both versions where they're not shutting down the kernels after.

Will update the MR for now with the changes to just add a kernelspec_manager in the constructor if one isn't gotten from the parent (RMKM explicitly passes a kernelspec_manager)

kevin-bates:
@golf-player - thanks for the updates.  I took this for another spin against my YARN cluster and got the following...
```
  File "/opt/anaconda2/envs/py3/lib/python3.6/runpy.py", line 183, in _run_module_as_main
    mod_name, mod_spec, code = _get_module_details(mod_name, _Error)
  File "/opt/anaconda2/envs/py3/lib/python3.6/runpy.py", line 109, in _get_module_details
    __import__(pkg_name)
  File "/home/gateway/indieRMK.py", line 9, in <module>
    client.execute()
  File "/opt/anaconda2/envs/py3/lib/python3.6/site-packages/nbclient/client.py", line 357, in execute
    with self.setup_kernel(**kwargs):
  File "/opt/anaconda2/envs/py3/lib/python3.6/contextlib.py", line 81, in __enter__
    return next(self.gen)
  File "/opt/anaconda2/envs/py3/lib/python3.6/site-packages/nbclient/client.py", line 339, in setup_kernel
    self.start_new_kernel_client(**kwargs)
  File "/opt/anaconda2/envs/py3/lib/python3.6/site-packages/nbclient/client.py", line 312, in start_new_kernel_client
    self.km.start_kernel(extra_arguments=self.extra_arguments, **kwargs)
  File "/opt/anaconda2/envs/py3/lib/python3.6/site-packages/enterprise_gateway/services/kernels/remotemanager.py", line 279, in start_kernel
    self._get_process_proxy()
  File "/opt/anaconda2/envs/py3/lib/python3.6/site-packages/enterprise_gateway/services/kernels/remotemanager.py", line 468, in _get_process_proxy
    self.process_proxy = process_proxy_class(kernel_manager=self, proxy_config=process_proxy_cfg.get('config'))
  File "/opt/anaconda2/envs/py3/lib/python3.6/site-packages/enterprise_gateway/services/processproxies/yarn.py", line 44, in __init__
    kernel_manager.parent.parent.yarn_endpoint)
AttributeError: 'NoneType' object has no attribute 'parent'
```
I went ahead and added a few more entries to `dependent_props` and was then able to run a notebook using nbclient.  I had nbclient 0.1.0 installed and noticed the kernel had not been terminated.  However, after updating to 0.2.0, I found the kernel terminated, but with a state of `KILLED`...
image
when "graceful" termination should produce a state of `FINISHED`.  This implies that the something isn't quite right with the shutdown sequence.  You might not see this using k8s, but it would be good to understand the why the message-based termination is not occurring.

In the YARN application log, we should see a shutdown indicator like the last line here.  Instead, we see all but that line.
```
Log Type: stdout
Log Upload Time: Sat May 16 07:20:33 -0700 2020
Log Length: 1772

[D 2020-05-16 07:17:44,825.825 launch_ipykernel] Using connection file '/tmp/kernel-6e7f4f81-1ffc-4c09-8a3b-0920f7fd2272_I_lTUb.json'.
[I 2020-05-16 07:17:44,826.826 launch_ipykernel] Signal socket bound to host: 0.0.0.0, port: 50501
[D 2020-05-16 07:17:44,830.830 launch_ipykernel] JSON Payload '{"stdin_port": 44132, "pgid": "3249", "ip": "0.0.0.0", "pid": "3299", "control_port": 47874, "hb_port": 40748, "signature_scheme": "hmac-sha256", "key": "cfc3a76f-964f-4f87-9bc8-1cbaaf8dd39f", "comm_port": 50501, "kernel_name": "", "shell_port": 53084, "transport": "tcp", "iopub_port": 39304}
[D 2020-05-16 07:17:44,836.836 launch_ipykernel] Encrypted Payload 'u5T6rmMK+4S6e659vVYcSooB9AA+yQhmrIXbTXBaMBdNjVGV3KaNTi1wHvxxLHZ/4J4xnw3j8ARbyEx0qEJ53LRUsphakrvh+IG2UtRloSgR9oBO+tUKf9LfXC0XHzorxuEdTr/byIEcQ6ChHgqqkxT4qnzum/TASyf8+MV3LBxW5M6gteoeTFeibvqD4hKa6UQm/A7+QT+aabxP5/flD1hPXPA+2AbDwSLyoLQpu+wHagihjEH6BqcH9YfuZ7ihErRQp4xb0ignUZRu7xvdvpg+KntBzq/rWvnlmLP+t9SUnc2/72pEJ6cYxat3SLpfg8ZsHUzhBlk8lJVlo9l0j6dqX7t+jDNzcUEwj/bL2Mk0E0RlQseb5IqDWcKYzjJjpYaooA8aOieLM3+zm6t8hg==

[I 2020-05-16 07:20:33,493.493 launch_ipykernel] gateway_listener got request: {u'shutdown': 1}
```

I've pushed the `parent.parent` removal to my repo - feel free to grab them or I can push to yours.  https://github.com/kevin-bates/enterprise_gateway/tree/remove-parent-parent

kevin-bates:
Looking at nbclient 0.2.0, if I add this line following [L335](https://github.com/jupyter/nbclient/blob/0.2.0/nbclient/client.py#L332-L349)
```
await ensure_async(self.km.shutdown_kernel(now=False))
```
I get a graceful shutdown with a `FINISHED` status.

golf-player:
Sorry about that, I didn't go through the other processproxy subclasses. 

Seems like the problem here is that graceful shutdown (kernel client shutdown) isn't shutting down the process. 

Looking into it.

kevin-bates:
The process is shutdown but nbclient isn't giving the kernel manager an initial chance to do it gracefully. This is a difference in behavior from what notebook does.

golf-player:
https://github.com/jupyter/nbclient/blob/0.2.0/nbclient/client.py#L334-335

Here the client is given the chance to shutdown gracefully rather than the manager

kevin-bates:
Right. This was my very comment from earlier today: https://github.com/jupyter/enterprise_gateway/pull/810#issuecomment-630270703

golf-player:
Sorry I'm a bit confused here. Is `self.kc.shutdown` not the correct way to shut down a kernel?

kevin-bates:
That sends the shutdown message to the kernel, but we leverage the kernel manager for interacting with the hosting resource manager and the shutdown method on the kernel manager includes both graceful and immediate options.

For the graceful route we send a comparable shutdown request to the listener but for immediate we enlist the help of the resource manager to do whatever is necessary. On YARN that equates to calling their KILL API.

golf-player:
So in the case of a docker container running a kernel process, the kernel gets the shutdown message, and at that point the process should exit 0, stopping the running container. 

I guess what's happening here is that the kill request happens while it's shutting down. I'm still trying to see that that's what's happening in the logs

golf-player:
Yeah that's what's happening; the shutdown message is received, then the pod gets killed.

kevin-bates:
We have the listener as well. It's a sibling to the kernel to handle things like interrupts. I haven't looked at the container process proxies lately but they probably terminate the container when the immediate route is taken. This is probably ok in container land but not in YARN where the status is killed rather than finished.

golf-player:
i think this warrants a discussion of its own with the nbclient devs. nbclient 0.3.0 I believe also does a force shutdown

> This is probably ok in container land but not in YARN where the status is killed rather than finished.

Agreed. distributed mode as well. And even in container land, if it's being `KILL`ed, it's possible some things don't stop gracefully, like [shutdown hooks](https://ipython.readthedocs.io/en/stable/api/generated/IPython.core.hooks.html#IPython.core.hooks.shutdown_hook)

golf-player:
https://github.com/jupyter/nbclient/issues/60

kevin-bates:
Thanks @golf-player.  I guess I'm okay with the docs.  They're really just an example and the code won't change even after nbclient fixes things.  At most I guess we'd update the nbclient version in the example, but since there's already 0.3.0, even that isn't a big deal.

I guess one of us should submit a PR to nbclient if we don't hear back from them soon.

Aside from that, how are we doing here?  Was the conductor endpoint commit the test-fix you were referring to?

golf-player:
Well as of nbclient 0.3.1, it works at least as well as 0.2.0, so that's good.

I'm working on a rough implementation to get some conversation going with them.

I think this PR is looking done more or less. Yeah it was a minor config change breaking tests, but no issues now.

kevin-bates:
Sounds good. Thanks for opening the nbclient PR.

I just ran this both as EG and Independent RMK (like [Independent George](https://www.youtube.com/watch?v=3afZip4BTRc)) and things look great.  I plan on performing a squash and merge first thing tomorrow.  If you'd prefer to squash to get the comment as you'd like please feel free, otherwise, I'll likely leave the meaningful commit comments in place or derive one myself.

I think is going to be really useful!  Thank you.

golf-player:
I've pushed a squashed commit. 

Thanks so much for all the help! Hopefully the nbclient PR goes well.

docs/source/kernel-library.md
```diff
@@ -0,0 +1,22 @@
+## Standalone Remote Kernel Execution 
+#####a.k.a "Library Mode"
```
kevin-bates, line 2: Shoot - just noticed this needs a space between the heading and `a.k.a`.

golf-player: Ugh my bad...

I blame pycharm.

This is fixed.

PR #812, opened 2020-04-27
Fix python-based images relative to recent changes

kevin-bates:
This commit addresses some recent changes and ensures the `future`
package is installed and replaces `pycrypto` with `pycryptodomex`.

Once merged, I think we can move forward with cutting 2.1.1.


kevin-bates:
Thanks Alan - merging.

PR #815, opened 2020-05-25
[Bugfix] Update cleanup method signature

jalpan-randeri:
Update cleanup method signature to match jupyter_client pkg

- https://github.com/jupyter/jupyter_client/pull/548#pullrequestreview-417826199

### DIFF
--- a/enterprise_gateway/services/kernels/remotemanager.py
+++ b/enterprise_gateway/services/kernels/remotemanager.py
@@ -447,11 +447,11 @@ class RemoteKernelManager(EnterpriseGatewayConfigMixin, IOLoopKernelManager):
         else:
             raise RuntimeError("Cannot signal kernel. No kernel is running!")
 
-    def cleanup(self, restart=False):
+    def cleanup(self, connection_file=True):
         """Clean up resources when the kernel is shut down"""
         warnings.warn("Method cleanup(connection_file=True) is deprecated, use cleanup_resources(restart=False).",
-                      DeprecationWarning)
-        self.cleanup_resources(restart)
+                      FutureWarning)
+        self.cleanup_resources(not connection_file)
 
     def cleanup_resources(self, restart=False):
         """Clean up resources when the kernel is shut down"""
@@ -463,7 +463,13 @@ class RemoteKernelManager(EnterpriseGatewayConfigMixin, IOLoopKernelManager):
         if self.process_proxy:
             self.process_proxy.cleanup()
             self.process_proxy = None
-        return super(RemoteKernelManager, self).cleanup_resources(restart)
+
+        import jupyter_client as jc
+        from distutils.version import LooseVersion
+        if LooseVersion(jc.__version__) < LooseVersion('6.2'):
+            return super(RemoteKernelManager, self).cleanup(connection_file=not restart)
+        else:
+            return super(RemoteKernelManager, self).cleanup_resources(restart)
 
     def write_connection_file(self):
         """Write connection info to JSON dict in self.connection_file if the kernel is local.
--- a/enterprise_gateway/services/kernels/remotemanager.py
+++ b/enterprise_gateway/services/kernels/remotemanager.py
@@ -6,6 +6,7 @@ import os
 import signal
 import re
 import uuid
+import warnings
 
 from tornado import gen, web
 from ipython_genutils.py3compat import unicode_type
@@ -430,6 +431,12 @@ class RemoteKernelManager(EnterpriseGatewayConfigMixin, IOLoopKernelManager):
 
     def cleanup(self, restart=False):
         """Clean up resources when the kernel is shut down"""
+        warnings.warn("Method cleanup(connection_file=True) is deprecated, use cleanup_resources(restart=False).",
+                      DeprecationWarning)
+        self.cleanup_resources(restart)
+
+    def cleanup_resources(self, restart=False):
+        """Clean up resources when the kernel is shut down"""
 
         # Note we must use `process_proxy` here rather than `kernel`, although they're the same value.
         # The reason is because if the kernel shutdown sequence has triggered its "forced kill" logic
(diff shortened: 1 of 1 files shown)


kevin-bates:
@jalpan-randeri - thank you for adding this PR.  Let's wait to see how https://github.com/jupyter/jupyter_client/pull/548 shakes out.  At a minimum, we'll want to have a backport to 5.3.x that is compatible with existing EG installations.

jalpan-randeri:
sure

jalpan-randeri:
@kevin-bates shall i update this pr with new interface?

kevin-bates:
Hi @jalpan-randeri - yes, please do.  Thank you.

enterprise_gateway/services/kernels/remotemanager.py
```diff
@@ -446,7 +447,13 @@ def signal_kernel(self, signum):
         else:
             raise RuntimeError("Cannot signal kernel. No kernel is running!")
 
-    def cleanup(self, connection_file=True):
+    def cleanup(self, restart=False):
+        """Clean up resources when the kernel is shut down"""
+        warnings.warn("Method cleanup(connection_file=True) is deprecated, use cleanup_resources(restart=False).",
```
kevin-bates, line 452: I don't think this is going to work without also modifying the jupyter_client dependency otherwise the call to `super().cleanup_resources()` will produce _AttributeError: 'super' object has no attribute 'cleanup_resources'_.  If we wanted to support older jupyter_client's we'd need to "plumb" `cleanup_resources()` as its own method chain, leaving `cleanup()` as-is.

I think it would be best to simply add a dependency on jupyter_client == 6.1.5 and then this looks good.

enterprise_gateway/services/kernels/remotemanager.py
```diff
@@ -446,7 +447,13 @@ def signal_kernel(self, signum):
...
+                      DeprecationWarning)
```
kevin-bates, line 453: Could you please update this to `FutureWarning` since deprecation warnings aren't produced when used from this location?

enterprise_gateway/services/kernels/remotemanager.py
```diff
@@ -446,7 +447,13 @@ def signal_kernel(self, signum):
         else:
             raise RuntimeError("Cannot signal kernel. No kernel is running!")
 
-    def cleanup(self, connection_file=True):
+    def cleanup(self, restart=False):
```
kevin-bates, line 450: Actually, looking at the test failures, you'll need to change this signature back to what it was and call `cleanup_resources()` with the inverse parameter value...
```
======================================================================
ERROR: Verify authorized users can start a kernel, unauthorized users cannot
----------------------------------------------------------------------
  File "/home/travis/build/jupyter/enterprise_gateway/enterprise_gateway/tests/test_gatewayapp.py", line 102, in tearDown
    self.app.shutdown()
  File "/home/travis/build/jupyter/enterprise_gateway/enterprise_gateway/enterprisegatewayapp.py", line 302, in shutdown
    self.kernel_manager.shutdown_kernel(kid, now=True)
  File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/notebook/services/kernels/kernelmanager.py", line 302, in shutdown_kernel
    return super(MappingKernelManager, self).shutdown_kernel(kernel_id, now=now)
  File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/jupyter_client/multikernelmanager.py", line 211, in shutdown_kernel
    km.shutdown_kernel(now=now, restart=restart)
  File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/jupyter_client/manager.py", line 395, in shutdown_kernel
    self.cleanup(connection_file=not restart)
TypeError: cleanup() got an unexpected keyword argument 'connection_file'
```

enterprise_gateway/services/kernels/remotemanager.py
jalpan-randeri: Oh shoot. yes you are correct. i completely forgot about client version. nice catch.


This is how i am thinking
```
    def cleanup(self, restart=False):
        """Clean up resources when the kernel is shut down"""
        warnings.warn("Method cleanup(connection_file=True) is deprecated, use cleanup_resources(restart=False).",
                      FutureWarning)
        self.cleanup_resources(restart)

    def cleanup_resources(self, restart=False):
        """Clean up resources when the kernel is shut down"""

        # Note we must use `process_proxy` here rather than `kernel`, although they're the same value.
        # The reason is because if the kernel shutdown sequence has triggered its "forced kill" logic
        # then that method (jupyter_client/manager.py/_kill_kernel()) will set `self.kernel` to None,
        # which then prevents process proxy cleanup.
        if self.process_proxy:
            self.process_proxy.cleanup()
            self.process_proxy = None

        import jupyter_client as jc
        from distutils.version import LooseVersion
        if LooseVersion(jc.__version__) < LooseVersion('6.2'):
            return super(RemoteKernelManager, self).cleanup(connection_file=not restart)
        else:
            return super(RemoteKernelManager, self).cleanup_resources(restart)
```

enterprise_gateway/services/kernels/remotemanager.py
jalpan-randeri: Fixed

enterprise_gateway/services/kernels/remotemanager.py
jalpan-randeri: Fixed

enterprise_gateway/services/kernels/remotemanager.py
kevin-bates: I think you came up with a solution along lines of my "method chain" comment - although I'd prefer to leave `cleanup()` as-is (the superclass call will produce the deprecation warning, so no need here).

Could you please add comments to each of the methods along the lines of...

for cleanup(): # This method has been deprecated in jupyter_client 6.1.5 and remains here for pre-6.2.0 jupyter_client installations.

for cleanup_resources(): # This method was introduced in jupyter_client 6.1.5 and will not be called until jupyter_client 6.2.0 has been released.

I just want to be sure it's clear to readers/maintainers these methods are mutually exclusive of each other.

jalpan-randeri: Done reset cleanup method to original one and added comments

PR #825, opened 2020-06-11
Enable ability to set default kernel name in container envs

### LINKED ISSUE #823: Container-based deployments need way to set default kernel name
Because the container-based deployments (Kubernetes, DockerSwarm) use kernel white lists and configuration options are environment variable based, we need to provide a means of setting `--RemoteMappingKernelManager.default_kernel_name` to a value other than the default `python3` when that value is not in the kernel white list.

Currently, `python3` is the default kernel, yet it's not in the kernel white list.  However, a start kernel request that doesn't specify a kernel-name will still succeed by starting the python3 - which is somewhat okay - although that kernel-name is not in a whitelist.

Ideally, notebook server would detect this condition and issue a warning message and/or reset the default kernel name to, say, the first item in the white list - or something to that effect.

However, the fact that setting the default kernel name at all in these environments is non-trivial should be resolved.

## Environment
- Enterprise Gateway Version 2.1
- Platform: Kubernetes, Docker

kevin-bates:
Adds appropriate plumbing (via environment variable `EG_DEFAULT_KERNEL_NAME`) to adjust the default kernel name in container environments and initializes those configurations with the appropriate python kernel.

Fixes #823

### DIFF
--- a/docs/source/kernel-kubernetes.md
+++ b/docs/source/kernel-kubernetes.md
@@ -92,6 +92,9 @@ spec:
           value: "60"
         - name: EG_KERNEL_WHITELIST
           value: "['r_kubernetes','python_kubernetes','python_tf_kubernetes','scala_kubernetes','spark_r_kubernetes','spark_python_kubernetes','spark_scala_kubernetes']"
+        - name: EG_DEFAULT_KERNEL_NAME
+          value: "python_kubernetes"
+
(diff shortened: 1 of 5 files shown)


akchinSTC:
Might need to update the example yaml snippet here:
https://github.com/jupyter/enterprise_gateway/blob/master/docs/source/kernel-kubernetes.md

kevin-bates:
Thanks Alan - I've pushed an update that renders like this in the docs...
image

PR #831, opened 2020-06-29
 response_socket remains open

aviadr:
In some cases, the response_socket remains open.
I Identified a few cases and fix it
(currently, the response_socket is randomly selected each time a new kernel is created, therefore this issue was not detected)


kevin-bates:
Hi @aviadr - thank you for the PR!  These changes look reasonable on the surface - nice catch!  It may be a day or so before I can perform a full review, but I wanted to let you know I've seen the initial changes.  Thanks again.

aviadr:
Fix one more place where the response socket remains open

kevin-bates:
Based on the recent updates, I'm beginning to think we should override `log_and_raise()` on `RemoteProcessProxy` such that it calls `close_response_socket()` then `super().close_response_socket()` and let the only "success-oriented" call be in `_update_connection()`.  We could then make close_response_socket "private" with an underscore prefix.

PR #832, opened 2020-06-30
Rollback option to use ZMQ global context

kevin-bates:
This is now default behavior in jupyter_client versions 5.3.5 and
6.1.5 and makes the previous change obsolete. This rolls back #820.


lresende:
Do you need to adjust anything on the minimun required dependencies? Didn't look into what is currently on master.

kevin-bates:
Thanks for the reviews.

@lresende, no, since the PR that is being rolled back (#820) didn't add any, but that was a good question.

PR #861, opened 2020-08-31
Update Tests to Spark 2.4.6

lresende:


etc/docker/demo-base/Dockerfile
```diff
@@ -26,7 +27,7 @@ ENV SHELL=/bin/bash \
 ENV HOME=/home/$NB_USER \
     PATH=$ANACONDA_HOME/bin:$HADOOP_PREFIX/bin:$JAVA_HOME/bin:$SPARK_HOME/bin:$PATH
 
-ENV SPARK_VER 2.4.1
+ENV SPARK_VER $SPARK_VERSION
```
kevin-bates, line 30: Is it necessary to set and ENV here?  Just wondering if we can reference the ARG SPARK_VERSION instead and not have SPARK_VER set in the env of the container.

If it is necessary to set the env (SPARK_VER) can we expand it to SPARK_VERSION?

etc/docker/enterprise-gateway-demo/Dockerfile
```diff
@@ -4,6 +4,7 @@ ARG SPARK_VERSION
 FROM $HUB_ORG/demo-base:$SPARK_VERSION
 
 ENV NB_USER="jovyan"
+ENV SPARK_VER $SPARK_VERSION
```
kevin-bates, line 7: Same as previous.  Can we skip this?

I do think we will need to wrap $SPARK_VERSION as ${SPARK_VERSION} and I don't know if that only "available" on envs or can also be applied to ARGs.

etc/kernel-launchers/scala/toree-launcher/build.sbt
```diff
@@ -6,15 +6,14 @@
 name := "toree-launcher"
 
 version := sys.props.getOrElse("version", default = "1.0").replaceAll("dev[0-9]", "SNAPSHOT")
+val sparkVersion = sys.props.getOrElse("spark_version", default = "2.4.6")
```
kevin-bates, line 9: Does this buy us anything if we always have to update the `default` anyway?  If this works, and the default is `2.4.6` because its the first use - that's cool - I'm good with this.

kevin-bates (commented):
Thanks Luciano.  A few comments regarding SPARK_VER vs. SPARK_VERSION.

etc/docker/demo-base/Dockerfile
lresende: This was already there and is following the same pattern as others (such as HADOOP_VER)

PR #880, opened 2020-09-18
Kubernetes Deployment Typo

ashvinnihalani:
Specifically when a kernel spec image is specified, emptydir get generated which is not valid, emtyDir is.


kevin-bates (approved):
Great catch Ashvin - thank you for the contribution!

PR #891, opened 2020-09-29
Add support for Conductor Proxy to work with JEG outside of Conductor Cluster

kjdoyle:
Today the Conductor Proxy and all of JEG for Conductor must run through a Conductor EGO Service and be setup in a certain way. This new support allows for JEG to be run anywhere but offload workload to Conductor using JWT Tokens for authentication.


kevin-bates (commented):
Hi Kevin.  This seems useful. 

Have you tried simultaneous starts of kernels in Conductor since EG 2.2 has been available?  You should see interleaved logging, whereas prior to EG 2.2, one POST request would complete before the other could be started.

I'm a little concerned these changes may wind up "blocking" the kernel start requests via the ConductorProcessProxy unless pushed to an executor, thereby allowing other start requests to get some "slice time".

Are you in a position to toss several 3 to 5 kernel start requests at this to see how things go?  Since you may not have played with EG 2.2 much, you should see significant improvement - at least prior to these changes.  I suspect that improvement may be diminished without the suggested change.

kevin-bates (approved):
Looks good Kevin - thank you.

PR #892, opened 2020-09-30
Handle supporting using both pycryptdome and pycryptodomex

kjdoyle:
In some environment pycryptodome is being used instead of pycryptodomex and can't be easily changed. This try/catch handles supporting using both pycryptdome and pycryptodomex.


kevin-bates (approved):
Makes sense - thank you.

akshaykjain:
thanks, Kevin...I tested this locally and works perfectly.

kevin-bates:
Excellent - thanks for the update @akshaykjain.

PR #893, opened 2020-10-04
Fix badge for pypi in README

dimon222:
Randomly noticed that there's broken badge not showing proper version. Apparently due to typo in repo name on Pypi (or, if it was potentially renamed)


kevin-bates (approved):
Nice catch Dmitry - thank you.

PR #894, opened 2020-10-05
EG_KERNEL_WHITELIST: include brackets in start-enterprise-gateway.sh 

ricklamers:
to make `EG_KERNEL_WHITELIST` consistent with other environment variable lists such as `EG_ENV_PROCESS_WHITELIST`.

Currently `EG_ENV_PROCESS_WHITELIST` is just comma separated: 
`EG_ENV_PROCESS_WHITELIST=A,B,C`

While `EG_KERNEL_WHITELIST` is comma separated and surrounded by square brackets: 
`EG_KERNEL_WHITELIST=['docker_py','docker_r']`

This PR proposes to drop the brackers on `EG_KERNEL_WHITELIST`.
`EG_KERNEL_WHITELIST='docker_py','docker_r'`

I'm not an expert on bash environment variables and thus not sure whether this is the cleanest/safest way to add "[...]" brackets.

@kevin-bates suggested on Gitter to make a PR. Hopefully this can serve as a starting point for a PR, perhaps someone with more familiarity with the EG code base can check whether it's comprehensive. I used code search to find all occurrences of `EG_KERNEL_WHITELIST`.

Note: it should be considered whether consistency is more important than backwards compatibility as with the currently proposed PR existing bracket including values of EG_KERNEL_WHITELIST will break. Adding bash code that will check `start-enterprise-gateway.sh` is possible but feels polluting.

### DIFF
--- a/etc/docker/enterprise-gateway/start-enterprise-gateway.sh
+++ b/etc/docker/enterprise-gateway/start-enterprise-gateway.sh
@@ -26,7 +26,8 @@ export EG_LOG_LEVEL=${EG_LOG_LEVEL:-DEBUG}
 export EG_CULL_IDLE_TIMEOUT=${EG_CULL_IDLE_TIMEOUT:-43200}  # default to 12 hours
 export EG_CULL_INTERVAL=${EG_CULL_INTERVAL:-60}
 export EG_CULL_CONNECTED=${EG_CULL_CONNECTED:-False}
-export EG_KERNEL_WHITELIST=${EG_KERNEL_WHITELIST:-"'r_docker','python_docker','python_tf_docker','scala_docker','spark_r_docker','spark_python_docker','spark_scala_docker'"}
+EG_KERNEL_WHITELIST=${EG_KERNEL_WHITELIST:-"'r_docker','python_docker','python_tf_docker','scala_docker','spark_r_docker','spark_python_docker','spark_scala_docker'"}
+export EG_KERNEL_WHITELIST=`echo ${EG_KERNEL_WHITELIST} | sed 's/[][]//g'`
 export EG_DEFAULT_KERNEL_NAME=${EG_DEFAULT_KERNEL_NAME:-python_docker}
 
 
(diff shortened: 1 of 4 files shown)


etc/docker/enterprise-gateway/start-enterprise-gateway.sh
```diff
@@ -26,15 +26,15 @@ export EG_LOG_LEVEL=${EG_LOG_LEVEL:-DEBUG}
 export EG_CULL_IDLE_TIMEOUT=${EG_CULL_IDLE_TIMEOUT:-43200}  # default to 12 hours
 export EG_CULL_INTERVAL=${EG_CULL_INTERVAL:-60}
 export EG_CULL_CONNECTED=${EG_CULL_CONNECTED:-False}
-export EG_KERNEL_WHITELIST=${EG_KERNEL_WHITELIST:-"['r_docker','python_docker','python_tf_docker','scala_docker','spark_r_docker','spark_python_docker','spark_scala_docker']"}
+export EG_KERNEL_WHITELIST=${EG_KERNEL_WHITELIST:-"'r_docker','python_docker','python_tf_docker','scala_docker','spark_r_docker','spark_python_docker','spark_scala_docker'"}
 export EG_DEFAULT_KERNEL_NAME=${EG_DEFAULT_KERNEL_NAME:-python_docker}
```
kevin-bates, line 30: Because the helm charts produce a bracketed value, and this may be useful for b/c purposes anyway, I found I needed something like the following to 'scrub' the brackets...
```bash

# Remove potential brackets from kernel whitelist
export EG_KERNEL_WHITELIST=`echo ${EG_KERNEL_WHITELIST} | sed 's/\[//' | sed 's/\]//'`
```

kevin-bates (commented):
Hi Rick - this is helpful, thank you.  I just had the one comment regarding the helm charts.  I am also unable to take the Docker compose script for a spin, but based on the other pieces of information, it looks good to me.

etc/docker/enterprise-gateway/start-enterprise-gateway.sh
kevin-bates: A single `sed` statement can also be used:
```bash
export EG_KERNEL_WHITELIST=`echo ${EG_KERNEL_WHITELIST} | sed 's/[][]//g'`
```

ricklamers: Neat idea! I should look at `sed` more often when working with shell scripts.

ricklamers: I've updated the PR. I thought a multi-line approach would be nicer. And a single export more logical? Let me know if it needs any changes.

kevin-bates: This looks great.  Would you mind adding a comment prior to the `sed` statement, something to the effect of: "Strip off any surrounding brackets that may have been set already".  That particular `sed` statement isn't very obvious, especially the way the brackets are organized within it - which matters (eg, `'s/[[]]//g'` doesn't work, although more intuitive).

ricklamers: Good idea! The `sed` statement does requiring some expert parsing to become obvious.

kevin-bates (approved):
Thanks Rick.

PR #898, opened 2020-10-14
Protect load_session() functionality when persistence is disabled

### LINKED ISSUE #896: FileKernelSessionManager.load_session() is not checking persistence enablement
The FileKernelSessionManager.load_session() method doesn't check that session persistence has been enabled.  This results in an attempt to access the `kernel_sessions` directory. However, if that directory does not exist, the code will attempt to create the appropriate hierarchy (which normally takes place during creation) - which can lead to issues on systems where, say, the _service user_ doesn't have permission to create the `kernel_sessions` directory.

kevin-bates:
The FileKernelSessionManager.load_session() method was not checking that persistence functionality was enabled, resulting in issues on more protected systems.  This adds the necessary enablement check that is already present in other methods.

Fixes #896

### DIFF
--- a/enterprise_gateway/services/sessions/kernelsessionmanager.py
+++ b/enterprise_gateway/services/sessions/kernelsessionmanager.py
@@ -325,9 +325,10 @@ class FileKernelSessionManager(KernelSessionManager):
                 self._load_session_from_file(kernel_session_file)
 
     def load_session(self, kernel_id):
-        if kernel_id is not None:
-            kernel_session_file = "".join([kernel_id, '.json'])
-            self._load_session_from_file(kernel_session_file)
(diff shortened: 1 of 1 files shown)


kevin-bates:
Not that I'm aware of, please let me know if you suspect others.

The main "public" methods were all protected prior to the addition of `load_session()` (note its singular form).  This method was added to provide a more dynamic HA support (active/active vs. active/passive) and, when added, I failed to see that its contents didn't check for enablement.

Note that some aspects of kernel session management are used even when persistence is disabled.  It is through this facility in which we track the "kernels per user" functionality/enforcement.

lresende (approved):

PR #908, opened 2020-12-07
Add support for Kernelspec caching

kevin-bates:
This pull request adds support for Kernel Specification caching.  Since various applications tend to poll for kernelspecs, and EG needs to support multiple such applications, it makes sense to route kernelspec requests through a caching facility, which defers to the configured KernelSpecManager if the kernelspec is not found.

This implementation uses `watchdog` to monitor the directory hierarchies containing kernelspecs.  If new kernelspecs are added that exist in non-monitored directories a cache miss will ensue - triggering the monitoring of the newly discovered directory.

In current trials, this appears to have about a 70% performance increase in kernelspec retrievals.

For now, the cache is disabled by default.  In EG 3.0 (or sooner?) we will enable caching to be on by default.  Until then, it can be enabled via CLI option `--KernelSpecCache.cache_enabled=True` or env `EG_KERNELSPEC_CACHE_ENABLED=True`.

- [x] Update configuration section of docs
- [x] Copy resource handler to EG and use cache to handle those requests (though static file handling will still work as today)


enterprise_gateway/services/kernelspecs/kernelspec_cache.py
```diff
@@ -0,0 +1,282 @@
...
+
+    This class contains the configured KernelSpecManager instance upon
+    which it uses to populate the cache (when enabled) or as a pass-thru
+    (when disabled).
+
+    Note that the KernelSpecManager returns different formats from methods
+    get_all_specs() and get_kernel_spec().  The format in which cache entries
+    are stored is that of the get_all_specs() results.  As a result, some
+    conversion between formats is necessary, depending on which method is called.
+    """
```
lresende, line 33: Based on the comment above, does it make sense to actually have a CachedKernelSpecManager(KernelSpecManager)?

enterprise_gateway/services/kernelspecs/kernelspec_cache.py
```diff
@@ -0,0 +1,282 @@
...
+                if kernel_dir not in self.observed_dirs:
+                    if os.path.exists(kernel_dir):
+                        self.log.info("KernelSpecCache: observing directory: {kernel_dir}".
+                                      format(kernel_dir=kernel_dir))
+                        self.observed_dirs.add(kernel_dir)
+                        self.observer.schedule(KernelSpecChangeHandler(self), kernel_dir, recursive=True)
+                    else:
+                        self.log.warn("KernelSpecCache: kernel_dir '{kernel_dir}' does not exist"
+                                      " and will not be observed.".format(kernel_dir=kernel_dir))
+            self.observer.start()
```
lresende, line 187: What happens if I go to one of the `jupyter --paths` directory that can be home to a kernelspec and add a new one there (e.g. all kernels were in `/Users/lresende/opt/anaconda/envs/dev/share/jupyter/kernels/` and I add a new one to `/usr/local/share/jupyter/kernels/`? Will the cache/observer identify the new kernelspec?

lresende:
Overall it looks good, I had a couple of question around some corner cases. 

Having said that, I believe this solves the overhead load on the server-side, but the contents of the kernelspec still go over the network, and maybe we could add support for [ETAGS](https://en.wikipedia.org/wiki/HTTP_ETag) on a subsequent PR.

enterprise_gateway/services/kernelspecs/kernelspec_cache.py
kevin-bates: Because the kernel spec manager class can be substituted, the cache needs to _contain_ the kernelspec manager, not _be_ the kernelspec manager.  For example, `nb-conda-kernels` introduces its own `KernelSpecManager` implementation.  As a result, I choose to not make it a subclass of `KernelSpecManager`.

enterprise_gateway/services/kernelspecs/kernelspec_cache.py
kevin-bates: The `self.kernel_spec_manager.kernel_dirs` contains the list of directories in which kernelspecs can reside.  The rub comes if any of those directories do not exist - since a non-existent directory cannot be observed.  However, any kernelspecs added to existing (observed) directories will be detected and automatically added to the cache.

If one of the `kernel_dirs` directories gets created and a kernelspec is added, the kernelspec will not be added to the cache until a cache-miss occurs on that named kernelspec.  However, the problem there is that it doesn't get added to the list of kernelspecs until its found, so end users won't know the new kernelspec is present.  The cache can be 'tickled' by triggering a cache miss using `/api/kernelspecs/new-spec-name` but that probably won't happen.

Of the 4 or 5 default runtime directories, I find only `/usr/share/jupyter/kernels` not existing.  However the more common `/usr/local/share/jupyter/kernels` directory is probably where those kernelspecs are persisted.

We could have another 'monitor' for any `kernel_dirs` that do not exist and add them to the observed list.  This check could be made whenever `get_all_specs` is called - or perhaps use a periodic poll for that.  I suppose if EG is started prior to any kernelspecs being created, this could be a problem, so it's may be worthwhile handling this one scenario.

kevin-bates:
Correct, this only addresses the redundant gathering of kernelspecs on the server.  I wasn't aware of ETags, but that makes total sense - good idea.  I'm assuming the browser would handle the `304` response, etc. and that client applications wouldn't need to change.

kevin-bates:
So my last commit removed `f` strings thinking that was preventing python 3.5 support, but it looks like PosixPath handling is an issue as well.  Sigh.  I think we should look at dropping 3.5 and perhaps moving the `jupyter_server` as our base implementation (and away from notebook).  This probably implies setting up a 2.x branch and this functionality would be part of a 3.0 release.

What do you think?

lresende (approved):

kevin-bates:
Here's a gist to add code to create missing kernel_dir entries (rather than issue a warning) if we want that functionality: https://gist.github.com/kevin-bates/45ddbc94f49f5415bf2985394dbf6925

PR #920, opened 2021-01-06
Update TF kernelspecs to Tensorflow 2.4.0 release

### LINKED ISSUE #919: Add tensorflow 2.x kernel to enterprise gateway launcher
Hi, please add support for tensorflow 2.x kernel to the launcher.

lresende:
Fixes #919

### DIFF
--- a/etc/docker/kernel-tf-gpu-py/Dockerfile
+++ b/etc/docker/kernel-tf-gpu-py/Dockerfile
@@ -1,5 +1,5 @@
 # Ubuntu:xenial
-ARG BASE_CONTAINER=tensorflow/tensorflow:1.12.0-gpu-py3
+ARG BASE_CONTAINER=tensorflow/tensorflow:2.4.0-gpu-jupyter
 FROM $BASE_CONTAINER
 
 RUN apt-get update && apt-get install -yq \
--- a/etc/docker/kernel-tf-py/Dockerfile
+++ b/etc/docker/kernel-tf-py/Dockerfile
@@ -1,5 +1,7 @@
 # Ubuntu:Bionic
-ARG BASE_CONTAINER=jupyter/tensorflow-notebook:04f7f60d34a6
+# TensorFlow 2.4.0
+ARG BASE_CONTAINER=jupyter/tensorflow-notebook:703d8b2dcb88
+
 FROM $BASE_CONTAINER
 
 ENV KERNEL_LANGUAGE python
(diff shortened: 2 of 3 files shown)


etc/docker/kernel-tf-py/Dockerfile
```diff
@@ -1,5 +1,6 @@
 # Ubuntu:Bionic
-ARG BASE_CONTAINER=jupyter/tensorflow-notebook:04f7f60d34a6
+ARG BASE_CONTAINER=jupyter/tensorflow-notebook:703d8b2dcb88
```
kevin-bates, line 2: Should we add a comment that indicates the tensorflow version here?  That might be helpful to maintainers.

etc/kernelspecs/python_tf_docker/kernel.json
```diff
@@ -1,6 +1,6 @@
 {
   "language": "python",
-  "display_name": "Python on Docker with Tensorflow",
+  "display_name": "Python on Docker with Tensorflow 2.x",
```
kevin-bates, line 3: I'd really prefer we not "go here" as this introduces 4 file updates for each major version increment and implies we DO support multiple versions.  In addition, the display-name doesn't necessarily correspond to what's in the image and folks are free to update both the image and display_name as they wish.

kevin-bates (commented):
I think we should move to 2.4.0 for the TF gpu image (Dockerfile and README) as well.

The test failure appears to be related to Notebook 6.1.6 and I suspect this change: https://github.com/jupyter/notebook/pull/5908 (sigh). We'll need to understand why this is affecting the test.

etc/docker/kernel-tf-gpu-py/Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Ubuntu:xenial
-ARG BASE_CONTAINER=tensorflow/tensorflow:1.12.0-gpu-py3
+ARG BASE_CONTAINER=tensorflow/tensorflow:2.4.0-gpu-py3
```
kevin-bates, line 2: This tag doesn't exist.  There's `2.4.0-gpu` and `2.4.0-gpu-jupyter` as well as another: https://hub.docker.com/r/tensorflow/tensorflow/tags?page=1&ordering=last_updated&name=2.4.0.

requirements.yml
```diff
@@ -7,7 +7,7 @@ dependencies:
   - jinja2>=2.10
   - jupyter_client>=6.1
   - jupyter_core>=4.6.0
-  - notebook>=6.1.0
+  - notebook>=6.1.0,<6.1.6
```
kevin-bates, line 10: I've created https://github.com/jupyter/enterprise_gateway/issues/922 so we can remember to remove (or bump) this pin once resolved.

kevin-bates (changes requested):

requirements.yml
lresende: Thank you

lresende:
@kevin-bates Updated the readme, please let me know if I missed any other places.

kevin-bates (approved):
Thanks Luciano.  We should probably merge #923 prior to this to avoid test failures though.

PR #921, opened 2021-01-06
Pin notebook version to <=6.1.6 avoid build issues

lresende:


kevin-bates:
I've created #922 to keep this on the radar.

kevin-bates (approved):
LGTM - we'll treat this as a temporary update.

PR #923, opened 2021-01-07
Fix handlers test - increase response poll loop

### LINKED ISSUE #922: Investigate test failure when running with Notebook 6.1.6
`pytest enterprise_gateway/tests/test_handlers.py` fails with the following output when running against Notebook 6.1.6 and is likely related to https://github.com/jupyter/notebook/pull/5908...
```
$ pytest enterprise_gateway/tests/test_handlers.py 
=============================================================================== test session starts ================================================================================
platform darwin -- Python 3.8.5, pytest-6.1.0, py-1.9.0, pluggy-0.13.1
rootdir: /Users/kbates/repos/oss/jupyter/enterprise_gateway
plugins: tornasync-0.6.0.post2, console-scripts-1.1.0, jupyter-0.0.1, shutil-1.7.0, jupyterlab-3.0.0rc10, virtualenv-1.7.0, jupyter-server-1.0.7, jupyterlab-server-2.0.0rc3
collected 23 items                                                                                                                                                                 

enterprise_gateway/tests/test_handlers.py ...........F...........                                                                                                            [100%]

===================================================================================== FAILURES =====================================================================================
__________________________________________________________________________ TestDefaults.test_kernel_comm ___________________________________________________________________________

self = <enterprise_gateway.tests.test_handlers.TestDefaults testMethod=test_kernel_comm>

    @gen_test
    def test_kernel_comm(self):
        """Default kernel should launch and accept commands."""
        ws = yield self.spawn_kernel()
    
        # Send a request for kernel info
        ws.write_message(json_encode({
            'header': {
                'username': '',
                'version': '5.0',
                'session': '',
                'msg_id': 'fake-msg-id',
                'msg_type': 'kernel_info_request'
            },
            'parent_header': {},
            'channel': 'shell',
            'content': {},
            'metadata': {},
            'buffers': {}
        }))
    
        # Assert the reply comes back. Test will timeout if this hangs.
        for _ in range(3):
            msg = yield ws.read_message()
            msg = json_decode(msg)
            if(msg['msg_type'] == 'kernel_info_reply'):
                break
        else:
>           self.assertTrue(False, 'never received kernel_info_reply')
E           AssertionError: False is not true : never received kernel_info_reply

enterprise_gateway/tests/test_handlers.py:404: AssertionError
```

kevin-bates:
Looks like the changes for jupyter/notebook#5908 have probably introduced an additional status response that pushed the actual `kernel_info_reply` out of the loop's range.  Increasing it by one resolves the issue.

Resolves #922

### DIFF
--- a/enterprise_gateway/tests/test_handlers.py
+++ b/enterprise_gateway/tests/test_handlers.py
@@ -395,7 +395,7 @@ class TestDefaults(TestHandlers):
         }))
 
         # Assert the reply comes back. Test will timeout if this hangs.
-        for _ in range(3):
+        for _ in range(4):
             msg = yield ws.read_message()
             msg = json_decode(msg)
             if(msg['msg_type'] == 'kernel_info_reply'):
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,7 +15,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        python-version: [ '3.5', '3.6', '3.7', '3.8' ]
+        python-version: [ '3.6', '3.7', '3.8', '3.9' ]
     steps:
     - name: Checkout
       uses: actions/checkout@v2
--- a/requirements.yml
+++ b/requirements.yml
@@ -17,6 +17,7 @@ dependencies:
   - requests>=2.7,<3.0
   - tornado>=4.2.0
   - traitlets>=4.2.0
+  - watchdog
   - yarn-api-client>=1.0
 
   # Test Requirements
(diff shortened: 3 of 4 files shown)


enterprise_gateway/tests/test_handlers.py
```diff
@@ -395,7 +395,7 @@ def test_kernel_comm(self):
         }))
 
         # Assert the reply comes back. Test will timeout if this hangs.
-        for _ in range(3):
+        for _ in range(4):
```
lresende, line 398: If you force Notebook to be 6.1.x other then 6.1.6, would this still work?

kevin-bates: Yes.  Prior versions will terminate the loop on the third iteration (rather than the fourth).

lresende (approved):
LGTM

PR #924, opened 2021-01-09
Remove NB2KG image and references

kevin-bates:
SInce NB2KG is only used for Notebook versions < 6.0 and 6.0 has been out for quite a while, it's time to update/remove such references from the docs and build.  Note that some references remain relative to using older versions of Notebook, but we will no longer build an image for convenience - although the previously built images will remain in docker hub.

Many of the references to _"NB2KG"_ have been replaced with _"gateway-enabled Notebook server"_ as that terminology is more generic.


lresende (approved):
LGTM, thanks for the cleanup

PR #925, opened 2021-01-10
Properly configure tzdata for TF GPU Kernel image

lresende:


etc/docker/kernel-tf-gpu-py/Dockerfile
```diff
@@ -9,11 +9,12 @@ RUN apt-get update && apt-get install -yq \
     libxrender1 \
     netcat \
     python3-dev \
-    tzdata \
-    unzip \
-    && rm -rf /var/lib/apt/lists/*
-
-RUN pip install future pycryptodomex
+    unzip && \
+    ln -fs /usr/share/zoneinfo/America/New_York /etc/localtime && \
```
kevin-bates, line 13: Could you explain why this is necessary?  It doesn't feel right as it seems to pin the TZ to EST.  A brief comment would be sufficient.

lresende: Newer version of tzdata now requires some setup in order to be properly installed, and this is just `a choice` of timezone. Another question could be, do we need to install data ?

lresende: I found a way to disable the requirement to set a timezone while installing tzdata using non-interactive env var, should have the same behavior as before now.

kevin-bates (approved):
LGTM - thanks Luciano!

PR #928, opened 2021-01-18
docs: remove misleading information on the default k8s in case of lib…

echarles:
Remove misleading information present in the documentation on the default kubernetes in case of library usage.


kevin-bates (approved):
Thanks Eric!

PR #937, opened 2021-02-17
Update build status to reflect GH actions

kevin-bates:
This updates the build status badge on the README to reflect the GitHub actions results. I've also removed the code coverage badge (and run) since they weren't being used.

I suspect this may need to be merged prior to seeing the final result because the `name` property in the workflow file was updated as well.  This value becomes the label for the badge.


Zsailer (approved):

Zsailer:
Thanks!

lresende (approved):

PR #961, opened 2021-05-01
fixes #960: lazy import of launch_instance in top-level __init__.py

### LINKED ISSUE #960: Imports in top-level __init__.py cause eager loading of most enterprise_gateway code
The top level `enterprise_gateway.__init__.py` contains a couple of `import`s that get fired whenever someone imports `enterprise_gateway` or anything below:

https://github.com/jupyter/enterprise_gateway/blob/443708ffb4cac466fb43e6daccddcdf24aed1945/enterprise_gateway/__init__.py#L4-L5

The `from ._version...` line is harmless, but the `from .enterprisegatewayapp import launch_instance` is problematic. Basically, the import from `enterprisegatewayapp` in turn imports most the rest of the code in the `enterprise_gateway` pkg, which causes some issues:

- Can help cause large slowdowns when fetching install-wide pkg metadata, eg a particular group of entry points
  - discussed extensively in this [ongoing PR](https://github.com/jupyter/jupyter_core/pull/209) about extending entry points to expose jupyter plugins and other resources
- it breaks monkey-patching
  - while not best practice, is currently required for eg overriding builtin `TokenAuthorizationMixin`

and last (maybe least?)

- it goes against the zen of python

telamonian:
Fixes #960 by making the problem import lazy. Also removed redundant `"scripts"` field from `setup.py` (noticed that since I had to make a small tweak to both that and `"entry_points"`).

If making the `launch_instance` import lazy seems too convoluted, I think another valid fix would be to remove `launch_instance` from the module top-level entirely. It's sufficient to be able to do:

```python
from enterprise_gateway.enterprisegatewayapp import launch_instance
```

### DIFF
--- a/enterprise_gateway/__init__.py
+++ b/enterprise_gateway/__init__.py
@@ -1,5 +1,8 @@
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
-"""Entrypoint for the enterprise gateway package."""
-from .enterprisegatewayapp import launch_instance
 from ._version import __version__
+
+"""Lazy-loading entrypoint for the enterprise gateway package."""
+def launch_instance(*args, **kwargs):
+    from enterprise_gateway.enterprisegatewayapp import launch_instance
+    launch_instance(*args, **kwargs)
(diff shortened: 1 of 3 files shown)


kevin-bates (approved):
These changes look good and appear to work fine - thank you for pointing this out Max!

PR #977, opened 2021-07-08
Update Watchdog dependency to version 2.1.3

lresende:
Update WatchDog dependency version to avoid issues while installing other projects such as Elyra:

```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
jupyter-enterprise-gateway 2.5.0 requires watchdog==0.10.3, but you have watchdog 2.1.3 which is incompatible.
```


kevin-bates (approved):
Thanks @lresende.

PR #981, opened 2021-07-21
Add support for YARN API non-anonymous access

### LINKED ISSUE #979: yarn_api_client error occurs when anonymous and kerberos are both disabled
## Description

I'm now testing enterprise gateway with spark_python_cluster_kernel at aliyun EMR Cluster and a yarn_api_client error occurs when the kernel is starting,  because my hadoop configuration `hadoop.http.authentication.simple.anonymous.allowed` is false while kerberos is not enabled.

I check configuration doc and find nothing about the way to set a username in api（simply add a user.name param of endpoint just work when anonymous and kerberos are both disabled, such as http://{rm_endpoint}/ws/v1/cluster?user.name=hadoop）.

Maybe it's nice to have the case supported when such configuration comes.

## Screenshots / Logs

截屏2021-07-14 下午9 39 00

## Environment

- Enterprise Gateway Version [2.5]
- Platform: [YARN 3.1.2]

kevin-bates:
If anonymous access is disallowed on the Hadoop YARN cluster,  and kerberos is not enabled, requests against the YARN API will fail with a `401` indicating authentication is required.  This change uses the newly added `SimpleAuth` from the yarn-api-client package to satisfy such configurations.  Since the specified username is not required to be an actual user, this fits in nicely with the notion of `KERNEL_USERNAME`, so we will set the `SimpleAuth`'s username to whatever we have deemed to be the kernel username at the time of the launch.  This resulted in a repositioning of the establishment of the resource manager connection to the actual kernel launch, rather than the `YarnClusterProcessProxy` constructor.

The implementation tolerates older versions of yarn-api-client since allowing anonymous access is the default behavior.  Users requiring non-anonymous access to their YARN clusters will need to ensure they have the appropriate version of the yarn-api-client package installed.

Special thanks to @dimon222 for enabling support for `SimpleAuth` in https://github.com/CODAIT/hadoop-yarn-api-python-client/pull/106 and @abzymeinsjtu for validating the solution! :pray:

Resolves #979

### DIFF
--- a/enterprise_gateway/services/processproxies/yarn.py
+++ b/enterprise_gateway/services/processproxies/yarn.py
@@ -13,6 +13,7 @@ import time
 from jupyter_client import localinterfaces
 from yarn_api_client.resource_manager import ResourceManager
 
+from ..sessions.kernelsessionmanager import KernelSessionManager
 from .processproxy import RemoteProcessProxy
 
 # Default logging level of yarn-api and underlying connectionpool produce too much noise - raise to warning only.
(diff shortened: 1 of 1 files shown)


abzymeinsjtu (approved):
LGTM

dimon222:
The long-waited dependency `yarn-api-client` that supports new Auth this is now released on Pypi under version 1.0.3.

lresende (approved):

PR #982, opened 2021-07-21
use idiomatic key checking

bgerrity:
mildly [unidiomatic](https://docs.python.org/3/glossary.html#term-eafp)


kevin-bates (approved):
These changes look good and, I agree, much easier to follow - thank you.

PR #988, opened 2021-08-06
Normalize kernel-launcher arguments

kevin-bates:
Prior to this change, all kernel-launcher arguments were _namespaced_ with a `RemoteProcessProxy` prefix.  With the introduction of [_Kernel Provisioning_](https://jupyter-client.readthedocs.io/en/latest/provisioning.html) in Jupyter Client 7.0 we can easily use the same launchers based on process proxies as will be based on [Remote Provisioners](https://github.com/kevin-bates/remote_provisioners).  As a result, it would be good to normalize the various arguments used by the kernel launchers so things like kernel images can be used by both Enterprise Gateway and remote kernel provisioners that are based on a similar framework.

Once Enterprise Gateway is updated to also be based on remote provisioners, its use of process proxies will be removed.

With this change, we can also share kernelspec definitions between process proxies and remote kernel provisioners.  I have not updated those kernelspec arguments to remove the `RemoteProcessProxy` prefix, only the argument parsers themselves such that they can tolerate both and use the older version as the fallback, preferring the new form of the arguments.  Such a _shared_ kernelspec would look something like the following...
```JSON
{
  "language": "python",
  "display_name": "Python on Kubernetes",
  "metadata": {
    "process_proxy": {
      "class_name": "enterprise_gateway.services.processproxies.k8s.KubernetesProcessProxy",
      "config": {
        "image_name": "elyra/kernel-py:dev"
      }
    },
    "kernel_provisioner": {
      "provisioner_name": "kubernetes-provisioner",
      "config": {
        "image_name": "elyra/kernel-py:dev"
      }
    }
  },
  "env": {
  },
  "argv": [
    "python",
    "/usr/local/share/jupyter/kernels/python_kubernetes/scripts/launch_kubernetes.py",
    "--kernel-id",
    "{kernel_id}",
    "--port-range",
    "{port_range}",
    "--response-address",
    "{response_address}",
    "--public-key",
    "{public_key}"
  ]
}
```

These launchers have been manually tested in YARN, Docker, and Kubernetes environments, and I have also confirmed updates to the kernelspecs that use the shorter forms of the arguments (i.e., `RemoteProcessProxy` has been removed) also work.


kevin-bates:
@lresende @akchinSTC - this could use a review.  There's a pending PR (#991) in which I'd prefer it not reference the `RemoteProcessProxy.` prefix on launch parameters.

PR #991, opened 2021-08-12
add crd process proxy to support some crd based k8s application like spark operator

### LINKED ISSUE #984: support of submit spark k8s kernel with gcp spark operator
hi @kevin-bates kevin, is there any plan about submit spark k8s kernel with [gcp spark operator](https://github.com/GoogleCloudPlatform/spark-on-k8s-operator), a popular k8s operator with a CRD including plenty feature of a spark application . 

p.s. i'm now working with the operator against aliyun EMR product, and i would be glad to contribute if possible.

abzymeinsjtu:
eg now has supported raw pod based k8s process proxy but maybe not so appropriate i wanna use some integrated k8s operator including some custom resources.

this PR implement a new process proxy, which makes it easy to define some kernel based on a certain k8s custom resource, such as spark application inside gcp spark operator. 

Resolves #984

### DIFF
--- a/docs/source/system-architecture.md
+++ b/docs/source/system-architecture.md
@@ -300,15 +300,24 @@ located.
 See [Enabling IBM Spectrum Conductor Support](kernel-conductor.html#enabling-ibm-spectrum-conductor-support) for details.
 
 ###### CustomResourceProcessProxy
-Enterprise Gateway also provides a implementation of a process proxy derived from KubernetesProcessProxy 
-called CustomResourceProcessProxy. 
+Enterprise Gateway also provides a implementation of a process proxy derived from `KubernetesProcessProxy` 
+called `CustomResourceProcessProxy`. 
 
-Instead of creating kernel based on Kubernetes pod, CustomResourceProcessProxy
-manages kernel by implementation of custom resource definition(CRD), e.g. SparkApplication is a CRD which include
-plenty component of a Spark on Kubernetes application.
+Instead of creating kernels based on a Kubernetes pod, `CustomResourceProcessProxy`
+manages kernels via a custom resource definition (CRD). For example,  `SparkApplication` is a CRD that includes
+many components of a Spark-on-Kubernetes application.
 
-If you want to include a new CRD, just wrt [spark_operator.py](https://github.com/jupyter/enterprise_gateway/blob/master/etc/kernel-launchers/kubernetes/scripts/launch_custom_resource.py)
-and [spark_python_operator](https://github.com/abzymeinsjtu/enterprise_gateway/blob/feat_crd_processproxy_support/etc/kernelspecs/spark_python_operator) kernelspec.
+If you are going to extend `CustomResourceProcessProxy`, just follow steps below:
+
+- override custom resource related variables(i.e. `group`, `version` and `plural`
+and `get_container_status` method, wrt [spark_operator.py](https://github.com/jupyter/enterprise_gateway/blob/master/etc/kernel-launchers/kubernetes/scripts/launch_custom_resource.py). 
+
+- define a jinja template like
+[sparkoperator.k8s.io-v1beta2.yaml.j2](https://github.com/jupyter/enterprise_gateway/blob/master/etc/kernel-launchers/kubernetes/scripts/sparkoperator.k8s.io-v1beta2.yaml.j2).
+As a generic design, the template file should be named as {crd_group}-{crd_version} so that you can reuse
+[launch_custom_resource.py](https://github.com/jupyter/enterprise_gateway/blob/master/etc/kernel-launchers/kubernetes/scripts/launch_custom_resource.py) in the kernelspec.
+
+- define a kernelspec like [spark_python_operator/kernel.json](https://github.com/jupyter/enterprise_gateway/blob/master/etc/kernelspecs/spark_python_operator/kernel.json).
 
 
 #### Process Proxy Configuration
--- a/docs/source/system-architecture.md
+++ b/docs/source/system-architecture.md
@@ -112,6 +112,14 @@ within a Docker Swarm cluster.
 within Docker configuration.  Note: because these kernels will always run local to the corresponding Enterprise Gateway instance, these process proxies are of limited use.
 - `ConductorClusterProcessProxy` - is responsible for the discovery and management of kernels hosted
 within an IBM Spectrum Conductor cluster.
+- `SparkOperatorProcessProxy` - is responsible for the descovery and management of kernels hosted 
+within a Kubernetes cluster but created as a SparkApplication instead of a Pod. The SparkApplication is a Kubernetes custom resource 
+defined inside the project [spark-on-k8s-operator](https://github.com/GoogleCloudPlatform/spark-on-k8s-operator), which 
+makes all kinds of spark on k8s components better organized and easy to configure.
+
+Before you run this kernel, please ensure that spark operator is installed under a certain namespace of your Kubernetes
+cluster.
+
 
 You might notice that the last five process proxies do not necessarily control the *launch* of the kernel.  This is 
 because the native jupyter framework is utilized such that the script that is invoked by the framework is what 
@@ -290,6 +298,18 @@ option or the `EG_CONDUCTOR_ENDPOINT` environment variable to determine where th
 located.  
 
 See [Enabling IBM Spectrum Conductor Support](kernel-conductor.html#enabling-ibm-spectrum-conductor-support) for details.
+
+###### CustomResourceProcessProxy
+Enterprise Gateway also provides a implementation of a process proxy derived from KubernetesProcessProxy 
+called CustomResourceProcessProxy. 
+
+Instead of creating kernel based on Kubernetes pod, CustomResourceProcessProxy
+manages kernel by implementation of custom resource definition(CRD), e.g. SparkApplication is a CRD which include
+plenty component of a Spark on Kubernetes application.
+
+If you want to include a new CRD, just wrt [spark_operator.py](https://github.com/jupyter/enterprise_gateway/blob/master/etc/kernel-launchers/kubernetes/scripts/launch_custom_resource.py)
+and [spark_python_operator](https://github.com/abzymeinsjtu/enterprise_gateway/blob/feat_crd_processproxy_support/etc/kernelspecs/spark_python_operator) kernelspec.
+
 
 #### Process Proxy Configuration
 Each kernel.json's `process-proxy` stanza can specify an optional `config` stanza that is converted 
--- a/enterprise_gateway/services/processproxies/crd.py
+++ b/enterprise_gateway/services/processproxies/crd.py
@@ -6,13 +6,13 @@ from .k8s import KubernetesProcessProxy
 from kubernetes import client
 
 
-class CustomResourceProxy(KubernetesProcessProxy):
+class CustomResourceProcessProxy(KubernetesProcessProxy):
     group = version = plural = None
     custom_resource_template_name = None
     kernel_resource_name = None
 
     def __init__(self, kernel_manager, proxy_config):
-        super(CustomResourceProxy, self).__init__(kernel_manager, proxy_config)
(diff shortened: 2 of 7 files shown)


kevin-bates:
Hi @abzymeinsjtu - this is great!  I've taken a quick glance at things and you definitely have the right idea.  I do have a few suggestions...

- Rather than derive from `RemoteProcessProxy` I think this should derive from `K8sProcessProxy`.  The reason is for all of the namespace management code.  We still want the ability for users to bring their own namespace, use a namespace-per container, or reside within the EG namespace, AND delete the namespace when the namespace is created on the fly.
- If derived from `K8sProcessProxy` I believe the resource-named methods would be replaced with the pod and container-named methods.  Those name changes could be easily "documented" via help strings when overridden in the `CustomResourceProxy` and `SparkOperatorProcessProxy` classes.  
- Its not clear to me if `CustomResourceProxy` would be necessary if derived from `K8sProcessProxy`.  Do you see an immediate possibility for supporting other operators?  If so, let's keep this subclass if we feel it provides usefulness.
- I believe the `{get,load}_process_info()` methods will go away since `ContainerProcessProxy` handles `assigned_node_ip` and `K8sProcessProxy` even adds support for dealing with the deletion of the namespace (in failover situations).
- Per the previous comments,  I think the `kernel.json` should add the `image_name` attribute into the config stanza, rather than define the env `KERNEL_IMAGE`.  That env will be set up by the `ContainerProcessProxy` superclass.
- Does the Spark Operator have the ability to specify an executor image as well?  If so, we should add `executor_image_name:` to the config stanza as well.  The superclasses would then handle that also.

This is really exciting - thank you for taking this on!

abzymeinsjtu:
thx @kevin-bates 

> Hi @abzymeinsjtu - this is great! I've taken a quick glance at things and you definitely have the right idea. I do have a few suggestions...
> 
> * Rather than derive from `RemoteProcessProxy` I think this should derive from `K8sProcessProxy`.  The reason is for all of the namespace management code.  We still want the ability for users to bring their own namespace, use a namespace-per container, or reside within the EG namespace, AND delete the namespace when the namespace is created on the fly.

yep, it really make sense if i make the `K8sProcessProxy` as base class for the ability to reuse much base function. But i've a concern that for gcp spark operator, one must install the spark operator under a certain namespace before they submit a ipykernel Spark Application. Apparently, it's not EG's job to install those components under a new namespace.

but it doesn't matter, i could make it a special case in `SparkOperatorProcessProxy`

> * Its not clear to me if `CustomResourceProxy` would be necessary if derived from `K8sProcessProxy`.  Do you see an immediate possibility for supporting other operators?  If so, let's keep this subclass if we feel it provides usefulness.

i think it's necessary for some basic operation are all the same among different custom resources such  as terminating custom resource and get custom resource detail.


> * Per the previous comments,  I think the `kernel.json` should add the `image_name` attribute into the config stanza, rather than define the env `KERNEL_IMAGE`.  That env will be set up by the `ContainerProcessProxy` superclass.
> * Does the Spark Operator have the ability to specify an executor image as well?  If so, we should add `executor_image_name:` to the config stanza as well.  The superclasses would then handle that also.

it's ok, spark operator support a wide range of spark on k8s configuration, even the webUI ingress setting. i believe i can put much more into the template.

abzymeinsjtu:
@kevin-bates  

some updates according to our discussion earlier, check again plz

enterprise_gateway/services/processproxies/crd.py
```diff
@@ -0,0 +1,45 @@
...
+        await super(CustomResourceProxy, self).launch_process(kernel_cmd, **kwargs)
+        return self
+
+    def kill(self):
+        result = None
+
+        if self.kernel_resource_name:
+            result = self.terminate_custom_resource()
+
+        return result
```
kevin-bates, line 32: Does this require a call to its superclass in order to terminate the container resources - like namespace, or does `terminate_custom_resource()` handle all of that?

How is the namespace lifecycle managed with custom resources?

enterprise_gateway/services/processproxies/crd.py
```diff
@@ -0,0 +1,45 @@
...
+        try:
+            delete_status = client.CustomObjectsApi().delete_cluster_custom_object(self.group, self.version,
+                                                                                   self.plurals,
+                                                                                   self.kernel_resource_name)
+
+            result = delete_status and delete_status.get('status', None) == 'Success'
+
+        except Exception as err:
+            result = isinstance(err, client.rest.ApiException) and err.status == 404
+
```
kevin-bates, line 44: Should this (or `kill()`) _unset_ `self.kernel_resource_name`?

kevin-bates (commented):
@abzymeinsjtu - this will be really good - thank you!

I had a couple of questions about class hierarchy calls and namespace management wrt custom resources.

enterprise_gateway/services/processproxies/crd.py
abzymeinsjtu: sorry for forgot it, namespace should be removed when such action required

kevin-bates:
Thanks for the update.  I'm beginning to wonder if we should tie `self.kernel_resource_name` to `self.container_name`.  If we did that, we could move the contents of crd.kill() into crd.terminate_container_resources() and essentially remove `terminate_custom_resource()` or make it a local method and crd.kill() goes away since termination is triggered completely from `ContainerProcessProxy.kill()`.  That approach feels a little forced (in the overload of `self.container_name`) but I like not having to introduce a new `terminate_xxx` method.

I can go either way, so if you're opposed to that, I'd be okay also.

abzymeinsjtu (commented):
> Thanks for the update. I'm beginning to wonder if we should tie `self.kernel_resource_name` to `self.container_name`. If we did that, we could move the contents of crd.kill() into crd.terminate_container_resources() and essentially remove `terminate_custom_resource()` or make it a local method and crd.kill() goes away since termination is triggered completely from `ContainerProcessProxy.kill()`. That approach feels a little forced (in the overload of `self.container_name`) but I like not having to introduce a new `terminate_xxx` method.

i suggest to keep the terminiate_container_resource() alone in case that custom resource maybe something far more than a single container.

meanwhile, tie `kernel_resource_name` and `container_name` together make it easy to reuse the base function in k8s process proxy.

consider it's ok to keep the current version

kevin-bates:
This is great.  One last thing (and I apologize for not mentioning this sooner but I just now realized) is that we should add entries to the docs where we list the various process proxy classes.  I think we only need to reference `SparkOperatorProcessProxy` [here](https://jupyter-enterprise-gateway.readthedocs.io/en/latest/system-architecture.html#process-proxy), but I think it would also be good to reference `CustomResourceProxy` in the slightly more detailed section [here](https://jupyter-enterprise-gateway.readthedocs.io/en/latest/system-architecture.html#remoteprocessproxy).

This is all looking really good. Thank you!

I'm not able to test this, but I trust that you're finding things working well?

abzymeinsjtu:
thx kevin, i will update the doc wrt

abzymeinsjtu:
thx kevin, have updated the doc and plz make a check

docs/source/system-architecture.md
```diff
@@ -112,6 +112,14 @@ within a Docker Swarm cluster.
 within Docker configuration.  Note: because these kernels will always run local to the corresponding Enterprise Gateway instance, these process proxies are of limited use.
 - `ConductorClusterProcessProxy` - is responsible for the discovery and management of kernels hosted
 within an IBM Spectrum Conductor cluster.
+- `SparkOperatorProcessProxy` - is responsible for the descovery and management of kernels hosted 
+within a Kubernetes cluster but created as a SparkApplication instead of a Pod. The SparkApplication is a Kubernetes custom resource 
+defined inside the project [spark-on-k8s-operator](https://github.com/GoogleCloudPlatform/spark-on-k8s-operator), which 
+makes all kinds of spark on k8s components better organized and easy to configure.
+
+Before you run this kernel, please ensure that spark operator is installed under a certain namespace of your Kubernetes
```
kevin-bates, line 120: I'm not sure I understand this statement: _ensure that spark operator is installed under a certain namespace_.

Does the `SparkOperatorProcessProxy` need to know _that_ namespace or are you saying that the spark operator just needs to be deployed with a known namespace?  How is this different from today's kernel pods?

docs/source/system-architecture.md
```diff
@@ -291,6 +299,18 @@ located.
 
 See [Enabling IBM Spectrum Conductor Support](kernel-conductor.html#enabling-ibm-spectrum-conductor-support) for details.
 
+###### CustomResourceProcessProxy
+Enterprise Gateway also provides a implementation of a process proxy derived from KubernetesProcessProxy
```
kevin-bates, line 303: Can we inclose all of the ProcessProxy class references in back-ticks: `KubernetesProcessProxy`, `CustomResourceProcessProxy`.

docs/source/system-architecture.md
```diff
@@ -291,6 +299,18 @@ located.
 
 See [Enabling IBM Spectrum Conductor Support](kernel-conductor.html#enabling-ibm-spectrum-conductor-support) for details.
 
+###### CustomResourceProcessProxy
+Enterprise Gateway also provides a implementation of a process proxy derived from KubernetesProcessProxy 
+called CustomResourceProcessProxy. 
+
+Instead of creating kernel based on Kubernetes pod, CustomResourceProcessProxy
+manages kernel by implementation of custom resource definition(CRD), e.g. SparkApplication is a CRD which include
+plenty component of a Spark on Kubernetes application.
```
kevin-bates, line 308:
```suggestion
Instead of creating kernels based on a Kubernetes pod, `CustomResourceProcessProxy`
manages kernels via a custom resource definition (CRD). For example,  `SparkApplication` is a CRD that includes
many components of a Spark-on-Kubernetes application.
```

docs/source/system-architecture.md
```diff
@@ -291,6 +299,18 @@ located.
...
+
+If you want to include a new CRD, just wrt [spark_operator.py](https://github.com/jupyter/enterprise_gateway/blob/master/etc/kernel-launchers/kubernetes/scripts/launch_custom_resource.py)
```
kevin-bates, line 310: I don't understand the _just wrt_ portion of this statement.  _Just with respect to `spark_operator.py` and `spark_python_operator` kernelspec. _ doesn't really parse for me.  What are you trying to say about this?

docs/source/system-architecture.md
```diff
@@ -291,6 +299,18 @@ located.
...
+and [spark_python_operator](https://github.com/abzymeinsjtu/enterprise_gateway/blob/feat_crd_processproxy_support/etc/kernelspecs/spark_python_operator) kernelspec.
```
kevin-bates, line 311: This link is pointing to your fork.

kevin-bates (commented):
Thanks for the updates.  Had some minor comments (mostly grammar-related) and a question.

docs/source/system-architecture.md
abzymeinsjtu: yep, the spark operator includes not only crd(SparkApplication), but also an operator pod which intend to inspect the running status of all running application as well as hook some action when events(start, end, error) happened to any application.

the pod must be deployed under a certain namespace first for start a spark application, so you understanding is correct.

docs/source/system-architecture.md
abzymeinsjtu: my fault

here, i should write down much more detail information about how to quickly extend the crd process proxy

docs/source/system-architecture.md
abzymeinsjtu: ok, updated

abzymeinsjtu:
hi kevin, just update again

kevin-bates (approved):
Really great stuff @abzymeinsjtu - thank you!

kevin-bates:
@telamonian - I'd love to have you review this if that's practical for you - given you're somewhat familiar with CRDs.  Given there's a `CustomResourceProcessProxy` base class, I'm sure we have _wiggle room_ for future extensions, etc., but just want to be sure there isn't anything obvious we're missing.

Issue #992, opened 2021-08-13
kernel image puller support k8s cluster with containerd instead of docker


abzymeinsjtu:
## Description

k8s is now removing docker support and replace it with containerd as default container engine. In EG, the kernel_image_puller script doesn't seem to support k8s cluster with containerd. 

so some improvement may be applied to current script and eg yaml file.

## Screenshots / Logs

![截屏2021-08-13 下午4 05 17](https://user-images.githubusercontent.com/4393648/129325179-9e736fbd-498e-4dae-9f73-ab6cdd69d450.png)

![截屏2021-08-13 下午4 05 59](https://user-images.githubusercontent.com/4393648/129325246-df3d742c-fb54-46a8-b56b-5483f34da658.png)

![截屏2021-08-13 下午8 46 17](https://user-images.githubusercontent.com/4393648/129359158-1487828a-6e88-4a9d-a546-086f7d385ba7.png)

## Environment

- Enterprise Gateway Version 2.5
- Platform: Kubernetes

kevin-bates:
Sigh.  Not sure what to do other than not deploy KIP and go back to having initial kernel invocations timeout due to their image upload.  

Is it possible to detect the container runtime (or convey it via helm values [which is certainly possible]) and update KIP to use the right mechanism?  I don't know anything about `containerd` at the moment (and don't have time to learn new stuff).  At a minimum, we could convey via an environment variable and only have KIP active if Docker is conveyed - just to prevent crash loops. 

I'm sorry about this.  Ideas and contributions are more than welcome!

abzymeinsjtu:
> Is it possible to detect the container runtime (or convey it via helm values [which is certainly possible]) and update KIP to use the right mechanism? I don't know anything about `containerd` at the moment (and don't have time to learn new stuff). At a minimum, we could convey via an environment variable and only have KIP active if Docker is conveyed - just to prevent crash loops.
> 

yes, V1Node object contains information about container runtime [object](https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/V1NodeSystemInfo.md), such as docker://{version}, containerd://{version}.


and there is also python [containerd](https://pypi.org/project/containerd/) lib

abzymeinsjtu:
@kevin-bates 

if so, i will be working on this later

abzymeinsjtu:
seems that containerd gRPC has no method to pull image, i can just failover when running without docker engine now

kevin-bates:
Addressed in #995

PR #995, opened 2021-08-17
make KIP sleep when container runtime is not docker

abzymeinsjtu:
#992 

i haven't find out a pull image api in containerd gRPC. so just failover first

### DIFF
--- a/etc/docker/kernel-image-puller/kernel_image_puller.py
+++ b/etc/docker/kernel-image-puller/kernel_image_puller.py
@@ -5,17 +5,6 @@ import requests
 import time
 
 from kubernetes import config, client
-
-config.load_incluster_config()
-first_node_name = client.CoreV1Api(client.ApiClient()).list_node().items[0].metadata.name
-node_info = client.CoreV1Api(client.ApiClient()).read_node(name=first_node_name)
-container_runtime = node_info.status.node_info.container_runtime_version
-
-if 'docker://' not in container_runtime:
-    while True:
-        # prevent a daemonset from restarting
-        time.sleep(10)
-
 from docker.client import DockerClient
 from docker.errors import APIError
 from docker.errors import NotFound
@@ -35,8 +24,6 @@ POLICY_ALYWAYS = "Always"
 policies = (POLICY_IF_NOT_PRESENT, POLICY_ALYWAYS)
 
 policy = os.getenv("KIP_PULL_POLICY", POLICY_IF_NOT_PRESENT)
-
-docker_client = DockerClient.from_env()
 
 logging.basicConfig(format='[%(levelname)1.1s %(asctime)s %(name)s.%(threadName)s] %(message)s')
 
@@ -163,6 +150,22 @@ if __name__ == "__main__":
     logger = logging.getLogger('kernel_image_puller')
     logger.setLevel(log_level)
 
+    config.load_incluster_config()
+    first_node_name = client.CoreV1Api(client.ApiClient()).list_node().items[0].metadata.name
+    node_info = client.CoreV1Api(client.ApiClient()).read_node(name=first_node_name)
+    container_runtime = node_info.status.node_info.container_runtime_version
+
+    if 'docker://' not in container_runtime:
+        while True:
+            logger.warning(
+                """
+                KernelInfoPuller only operates on Docker-based K8s systems and is unable to pull kernel images at this time.  
+                Please pre-seed your nodes with the necessary kernel images prior to their use.
+                """
+            )
+
+            time.sleep(60 * 5)
+
     # Determine pull policy.
     pulled_images = set()
     if policy not in policies:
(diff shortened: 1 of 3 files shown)


etc/docker/kernel-image-puller/kernel_image_puller.py
```diff
@@ -4,12 +4,23 @@
...
+
+config.load_incluster_config()
+first_node_name = client.CoreV1Api(client.ApiClient()).list_node().items[0].metadata.name
+node_info = client.CoreV1Api(client.ApiClient()).read_node(name=first_node_name)
+container_runtime = node_info.status.node_info.container_runtime_version
+
+if 'docker://' not in container_runtime:
+    while True:
+        # prevent a daemonset from restarting
+        time.sleep(10)
```
kevin-bates, line 17: Could we log a warning message every 5 minutes or so indicating that KIP isn't doing anything and it's recommended that kernel images be pulled manually prior to their use?  Something like:
```python
logger.warning("KernelInfoPuller only operates on Docker-based K8s systems and is unable to pull kernel images at this time.  Please pre-seed your nodes with the necessary kernel images prior to their use.")
```

The wording for _Docker-based K8s systems_ may not be correct, but that's an idea of what we should let the user/admin know is happening.

kevin-bates (commented):
This is a real drag (that `containerd` doesn't have a pull capability).  Thank you for working on this since this is better than crash loop stuff.

Just had the one comment about letting folks know KIP has decided to take a break. :smile:

etc/docker/kernel-image-puller/kernel_image_puller.py
abzymeinsjtu: ok, add the log and sleep longer

kevin-bates:
Thanks @abzymeinsjtu.  Since KIP will essentially be use-less in non-docker-based k8s configurations, I think we need a statement to that effect at the of its documentation: https://github.com/jupyter/enterprise_gateway/blob/master/docs/source/kernel-kubernetes.md#kernel-image-puller

Something like:
> Note: As of Kubernetes version X support for docker-based systems now defaults to containerd.  (Not sure what the proper wording for that should be.)  In such configurations, Kernel Image Puller will not attempt to pull kernel images down to the node on which it is running.  Instead, it will log a message at 5 minute intervals indicating that kernel images should be pulled prior to their use.  We are still looking for a solution and contributions are welcome.

Hmm.  I wonder if we should nicely log (one per line) the set of kernel image names as reflected in the kernelspecs when posting the log periodically.  This would be helpful as it prevents the admin from having to go and dig out the kernel image in every "visible" kernelspec and better integrates with the existing flow (where we simply not attempt the pull but replace with a log message instead).  

I believe this would amount to making [this block](https://github.com/jupyter/enterprise_gateway/blob/d6d6ff59fb33ddb9916f1e396f5e335d5a3dc6b8/etc/docker/kernel-image-puller/kernel_image_puller.py#L111-L119) conditional and just logging the image.  We'd produce the primary message about pre-seeding kernel images by adding a condition in [this else block](https://github.com/jupyter/enterprise_gateway/blob/d6d6ff59fb33ddb9916f1e396f5e335d5a3dc6b8/etc/docker/kernel-image-puller/kernel_image_puller.py#L192).  This way we automatically leverage the configured internal (which defaults to 5 minutes anyway).

abzymeinsjtu:
> Hmm. I wonder if we should nicely log (one per line) the set of kernel image names as reflected in the kernelspecs when posting the log periodically. This would be helpful as it prevents the admin from having to go and dig out the kernel image in every "visible" kernelspec and better integrates with the existing flow (where we simply not attempt the pull but replace with a log message instead).

the idea is fine, but means more changes  because the program fail when initializing docker client 
i.e. `docker_client = DockerClient.from_env()`.

abzymeinsjtu:
@kevin-bates

etc/docker/kernel-image-puller/kernel_image_puller.py
```diff
@@ -89,7 +89,20 @@ def pull_image(image_name):
...
     the caller handle others.
+
+    If container runtime is not docker(e.g. containerd), the method returns.
     """
+    if docker_client is None:
+        logger.warning(
+            """
+            KernelInfoPuller only operates on Docker-based K8s systems and is unable to pull kernel images at this time.  
+            Please pre-seed your nodes with the necessary kernel images prior to their use.
+            Image '{}' cannot be pulled.
```
kevin-bates, line 100: Thanks for the update, its appreciated.

I think we should issue the long warning once per iteration (in `fetch_image_names()` or the loop its called from`) and just issue "Image '{}' cannot be pulled."  per image, otherwise its too noisy.

You're missing an f-string or `.format()` suffix.

abzymeinsjtu: yes, really make sense

abzymeinsjtu: now i put it in `fetch_image_names` method, and when there is no docker runtime, nothing will be queued to be pulled

kevin-bates (approved):
LGTM - thank you @abzymeinsjtu!  Although it's a drag that KIP isn't effective in non-docker configurations, at least this provides some utility.

Issue #989, opened 2021-08-10
Configuration option to restrict requests to origin


bgerrity:
## Problem

Jupyter Enterprise Gateway [assumes](https://jupyter-enterprise-gateway.readthedocs.io/en/latest/getting-started-security.html#security-features) that all users issuing requests are authenticated. 

For my purposes, a reverse proxy is responsible for that user authentication. I'd like to be able to configure the gateway to reject any requests that don't originate from that origin.

## Proposition

I was wondering if there was any interest in adding this as a configuration option? Or perhaps there's already a (better) way to achieve the same effect? I was planning on monkey patching tornado for my specific use case but I wanted to ask if there were a shared interest, in which case, it'd be easiest to cut a PR.

kevin-bates:
Hi @bgerrity - I think this is a great idea and, since many of the maintainers are not experts in this area, would greatly appreciate your help (and contribution).  Thank you.

I'd like to cut a 2.6 release sometime this week or next, so it would be fantastic if this can make it in.

bgerrity:
Sorry about the hold up in getting the changes published: I've opened a PR #997.

kevin-bates:
Addressed in #997

PR #997, opened 2021-08-20
Add origin restriction config

bgerrity:
Addressing #989: It's a fairly simple SSL check and doesn't offer any support for running without authentication. Not sure if this is the best approach; open to any feedback or alternative implementations.

### DIFF
--- a/docs/source/config-options.md
+++ b/docs/source/config-options.md
@@ -114,6 +114,12 @@ EnterpriseGatewayApp(EnterpriseGatewayConfigMixin, JupyterApp) options
     carefully as it can dramatically limit who can launch kernels.
     (EG_AUTHORIZED_USERS env var - non-bracketed, just comma-separated)
     Default: set()
+--EnterpriseGatewayApp.authorized_origin=<Unicode>
+    Hostname (e.g. 'localhost', 'reverse.proxy.net') which the handler will
+    match against the request's SSL certificate.  An HTTP 403 (Forbidden) error
+    will be raised on a failed match.  This option requires TLS to be enabled.
+    It does not support IP addresses. (EG_AUTHORIZED_ORIGIN env var)
+    Default: ''
 --EnterpriseGatewayApp.base_url=<Unicode>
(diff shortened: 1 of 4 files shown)


kevin-bates (commented):
@bgerrity - these changes look good and helpful - thank you.

There are some linting issues with the `authorized_origin` help string that should get fixed up, but once fixed, the default case still appears to work.

We should also update the configuration section of the docs (near `authorized_users`): https://github.com/jupyter/enterprise_gateway/blob/master/docs/source/config-options.md

Are you able to adequately test `authorized_origin` having a value?  (Sorry I'm not able to do that.)

kevin-bates:
Thanks for the update.  Looks like a test is side-effected by this, but I'm not certain the test is that valid.

https://github.com/jupyter/enterprise_gateway/blob/d6d6ff59fb33ddb9916f1e396f5e335d5a3dc6b8/enterprise_gateway/tests/test_gatewayapp.py#L97

Seems like the test is just ensuring the SSL version.  Is there another way this could be triggered w/o a non-existent cert file?

bgerrity:
@kevin-bates Fixed the lint and added the option to the config docs; I'll fix that unit test.

For manual testing, I enabled TLS (`--certfile`, `--keyfile`, `--client-ca`) and ran with `--EnterpriseGatewayApp.authorized_origin="localhost"` and another origin.

bgerrity:
> Seems like the test is just ensuring the SSL version. Is there another way this could be triggered w/o a non-existent cert file?

Since the options build is being delegated to `ssl.SSLContext`, I don't think there's any need to validate it.

kevin-bates:
> Since the options build is being delegated to ssl.SSLContext, I don't think there's any need to validate it. IMO it's best to remove the second part of that test.

I'm okay with that.  I think version 5 is the default anyway and it's configurable as well.

bgerrity:
Cut down that unit test.

bgerrity:
My bad for the noise, will fix that other test as well.

kevin-bates (approved):
Looks good Brendan - thank you!

PR #1007, opened 2021-10-15
KIP: Add support for containerd-based installations

kevin-bates:
PR #995 was introduced to better tolerate installations configured with a non-docker container runtime.  This PR introduces the use of `crictl` (Containerd's CLI tool) to check image existence and pull images when a docker runtime is not configured.

Since `crictl` needs to mount the container runtime socket (to accomplish docker-in-docker, or rather, container-in-container) this change makes that socket reference a configurable value in the helm chart.

If a CRI socket is configured that is not detected as Docker or Containerd, KIP will proceed with attempting image management using `crictl` since it can take a _runtime endpoint_ that is composable from the mounted socket value.  That said, there's a good chance that another runtime may introduce issues, but I'm hoping this PR introduces an appropriate framework to add support beyond `critctl`.

This PR also introduces a _default container registry_ configurable that allows any non-qualified images to be pulled from alternate registries (e.g., quay.io).

This was tested against Kubernetes clusters configured with Docker and Containerd runtimes.

I also converted statements using `.format` to `f-strings` - which introduces some noise on the changes. (sorry)

@abzymeinsjtu - I'm hoping you might have a chance to review this as well. 🙏

### DIFF
--- a/docs/source/kernel-kubernetes.md
+++ b/docs/source/kernel-kubernetes.md
@@ -246,6 +246,10 @@ The Kernel Image Puller can be configured for the interval at which it checks fo
 
 If the Enterprise Gateway defines an authentication token (`EG_AUTH_TOKEN`) then that same token should be configured here as (`KIP_AUTH_TOKEN`) so that the puller can correctly authenticate its requests.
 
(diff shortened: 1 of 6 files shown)


akchinSTC (approved):
LGTM!

lresende (approved):

PR #1028, opened 2022-01-25
Drop Python 3.6 and add Python 3.10 to setup metadata and CI matrix

kevin-bates:
Update setup.py and CI workflows to no longer list Python 3.6 but also _add_ Python 3.10.  

- Took this opportunity to address what appears to be more status messages being returned in `test_kernel_comm` (increased range from 4 to 8 - looks like there are now 5 status messages prior to `kernel_info_reply`).
- Got rid of `get_new_ioloop()` override since we want a new loop created per test anyway.


akchinSTC (approved):

lresende (approved):

lresende:
This needs to be rebased before it can be merged.

PR #1031, opened 2022-01-26
Spark operator k8s updates to enable it in gateway on k8s

amarinder-bindra:
This adds the Spark Operator to K8s kernespecs package, updates the spark operator parameters to follow the new EG/Jupyter Client pattern, and fixes in the proxy to be able to run and submit jobs via it

### DIFF
--- a/etc/kernel-launchers/kubernetes/scripts/launch_custom_resource.py
+++ b/etc/kernel-launchers/kubernetes/scripts/launch_custom_resource.py
@@ -1,3 +1,4 @@
+#!/opt/conda/bin/python
 import os
 import yaml
 import argparse
--- a/etc/kernel-launchers/kubernetes/scripts/launch_custom_resource.py
+++ b/etc/kernel-launchers/kubernetes/scripts/launch_custom_resource.py
@@ -27,7 +27,7 @@ def launch_custom_resource_kernel(kernel_id, port_range, response_addr, public_k
     keywords['eg_response_address'] = response_addr
     keywords['kernel_id'] = kernel_id
     keywords['kernel_name'] = os.path.basename(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
-    keywords['kernel_spark_context_init_mode'] = spark_context_init_mode
+    keywords['spark_context_initialization_mode'] = spark_context_init_mode
 
     for name, value in os.environ.items():
         if name.startswith('KERNEL_'):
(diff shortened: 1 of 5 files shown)


lresende (approved):
Thanks @amarinder-bindra, very good to have Spark Operator updates to make it work out of the box with the EG images.

etc/kernel-launchers/kubernetes/scripts/launch_kubernetes.py
```diff
@@ -1,3 +1,4 @@
+#!/opt/conda/bin/python
```
kevin-bates, line 1: This is interesting.  What does this do?  Does `launch_custom_resource.py` need this as well?

When called via `python launch_kubernetes.py` does it have an influence on _which_ python is used?  (Just curious)

kevin-bates: Hmm - this does seem like a _fallback_ mechanism for when the .py script is not invoked using `python`: https://stackoverflow.com/questions/6908143/should-i-put-shebang-in-python-scripts-and-what-form-should-it-take#:~:text=14%20Answers&text=The%20shebang%20line%20in%20any,manager%20(when%20configured%20properly).

Was this needed for testing the script or something since every invocation from a `kernel.json` uses `python path-to-launch_kubernetes.py`?

At any rate, I guess it doesn't hurt, but we should be consistent (relative to the `launch_custom_resource.py`).

lresende: Good catch @kevin-bates we need this on the `launch_custom_resource.py` as well.

etc/kernel-launchers/kubernetes/scripts/launch_custom_resource.py
```diff
@@ -47,15 +47,15 @@ def launch_custom_resource_kernel(kernel_id, port_range, response_addr, public_k
 
 if __name__ == '__main__':
     parser = argparse.ArgumentParser()
-    parser.add_argument('--RemoteProcessProxy.kernel-id', dest='kernel_id', nargs='?',
+    parser.add_argument('--kernel-id', dest='kernel_id', nargs='?',
```
kevin-bates, line 50: If you're going to use the truncated options for this application, then you need to update https://github.com/jupyter/enterprise_gateway/blob/master/etc/kernelspecs/spark_python_operator/kernel.json to use the truncated forms as well.

kevin-bates (changes requested):
Hi @amarinder-bindra - thanks for the pull request.

Looks like the `spark_operator/kernel.json` requires an update to use the shortened arguments as well.

Other than that, and the question about the shebang, this looks good.  Thanks.

etc/kernel-launchers/kubernetes/scripts/launch_kubernetes.py
amarinder-bindra: yes, need to add in launch_custom_resource.py. 
Also, without the change, I got the error when invoking it via Enterprise Gateway.

etc/kernel-launchers/kubernetes/scripts/launch_custom_resource.py
amarinder-bindra: thanks @kevin-bates for catching this. Better will be to add an alias rather than removing it from the launch script so that deployment files with old arguments will also work.

kevin-bates: Sounds good @amarinder-bindra.  I think EG 3.0 would be a good time to move to the truncated forms.

etc/kernel-launchers/kubernetes/scripts/launch_custom_resource.py
```diff
@@ -48,15 +48,15 @@ def launch_custom_resource_kernel(kernel_id, port_range, response_addr, public_k
 
 if __name__ == '__main__':
     parser = argparse.ArgumentParser()
-    parser.add_argument('--RemoteProcessProxy.kernel-id', dest='kernel_id', nargs='?',
+    parser.add_argument('--kernel-id', '--RemoteProcessProxy.port-range', dest='kernel_id', nargs='?',
```
lresende, line 51: A small typo above, should be `--RemoteProcessProxy.kernel-id`

amarinder-bindra: thanks for catching it, fixed it

lresende:
With the required typo fix above, I have updated my dev environment and it all seems to work with kernelspecs using the `short-names` and the `RemoteProcessProxy. long-names`.

etc/kernel-launchers/kubernetes/scripts/launch_custom_resource.py
```diff
@@ -47,15 +48,15 @@ def launch_custom_resource_kernel(kernel_id, port_range, response_addr, public_k
 
 if __name__ == '__main__':
     parser = argparse.ArgumentParser()
-    parser.add_argument('--RemoteProcessProxy.kernel-id', dest='kernel_id', nargs='?',
+    parser.add_argument('--kernel-id', '--RemoteProcessProxy.kernel-id', dest='kernel_id', nargs='?',
```
kevin-bates, line 51: Very nice.  I didn't realize different _flavors_ for the same option could be used.  Thanks!

kevin-bates (approved):
Thanks for the update @amarinder-bindra and @lresende. 

This is a useful PR - thanks again.

PR #1035, opened 2022-02-01
Revert the RTD python version to 3.8

kevin-bates:
The RTD builds were breaking with the merge of https://github.com/jupyter/enterprise_gateway/pull/1034 due to the Python 3.9 reference: https://readthedocs.org/projects/jupyter-enterprise-gateway/builds/15945185/

This PR reverts the Python version to 3.8 for the time being.  We can revisit later (when the docs are reformatted).

### DIFF
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -5,6 +5,6 @@ channels:
   - free
 dependencies:
   - pip
-  - python=3.9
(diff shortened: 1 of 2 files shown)


kevin-bates:
Merging now so we can get to the RTD build.

PR #1040, opened 2022-02-16
Reorganize documentation

kevin-bates:
This pull request reorganizes the Enterprise Gateway documentation into a role-based model like that used in the Jupyter Server docs.  Nearly every file changed is for the docs. Much of the content was just "moved" although everything was updated to current operations.  

To build the docs...
image

There are five sections, four of which are "guides" (Users, Developers, Operators and Contributors) and a fifth section "Other" for things that either span the sections or just general stuff (like related references).

Each "Guide" starts with a _Use Cases_ note indicating the kinds of things that apply to folks of that persona or role.
image

Over time, we can probably improve on the flow of things.  I also had a hard time with the myst-parser getting certain links to work.


kevin-bates:
Hello @rahul26goyal and @kiersten-stokes, although I can't add you as "official" reviewers, you had previously stated an interest in Enterprise Gateway and I thought a review of the documentation might be a good introduction.  (Of course, I suppose it could also be a terrifying experience! :smile: )

kevin-bates:
I should probably add for folks new to EG, you'll need to run `make env` which creates a new conda environment named `enterprise-gateway-dev` by default.  If you'd like a different name you can run `ENV=my-env-name make env`, but doing so will require the `ENV=my-env-name` whenever `make` is run.  Perhaps, in that case, a good alias would be:
```bash
alias eg_make='ENV=my-env-name make'
```

docs/source/conf.py
```diff
@@ -57,7 +60,7 @@
 
 # General information about the project.
 project = 'Jupyter Enterprise Gateway'
-copyright = "2022, Jupyter Server."
+copyright = "2022, Jupyter"
```
lresende, line 63: Should we have this as "Project Jupyter"

lresende (approved):

lresende (approved):

Issue #968, opened 2021-05-05
There are some incorrect links in the DockerHub description.


taeyeopkim1:
## Description
Hello.
In fact, this is what I found while looking for a solution to [#962](https://github.com/jupyter/enterprise_gateway/issues/962).

The link for "enterprise-gateway-swarm.sh" and "enterprise-gateway-docker.sh" in "https://hub.docker.com/r/elyra/enterprise-gateway/" seems to be connected incorrectly.

https://github.com/jupyter/enterprise_gateway/blob/master/etc/docker/enterprise-gateway-swarm.sh
https://github.com/jupyter/enterprise_gateway/blob/master/etc/docker/enterprise-gateway-docker.sh

When I connect, I can see the NotFound page.
I don't know where to ask, so I post this issue.

## Screenshots / Logs
- Screenshot 1
![image](https://user-images.githubusercontent.com/43223381/117165273-4f623000-ae00-11eb-96a8-baefe0035ee4.png)

- Screenshot 2
![image](https://user-images.githubusercontent.com/43223381/117165443-74ef3980-ae00-11eb-9c2e-692898d32f3a.png)

## Environment

- Enterprise Gateway Version []
- Platform: [DockerHub]
- Others []

kevin-bates:
What version of the docs are you looking at?  I don't find these sections in any of the recent versions.
That said, we do have outstanding references to `enterprise-gateway-swarm.sh` in [`kernel-docker.md`](https://github.com/jupyter/enterprise_gateway/blob/master/docs/source/kernel-docker.md#dockerprocessproxy) and [`docker.md`](https://github.com/jupyter/enterprise_gateway/blob/master/docs/source/docker.md#elyraenterprise-gateway) that should be appropriately replaced with [`docker-compose.yml`](https://github.com/jupyter/enterprise_gateway/blob/master/etc/docker/docker-compose.yml).

taeyeopkim1:
Well..  What I found was just an overview.
Would this screenshot help?

![image](https://user-images.githubusercontent.com/43223381/117229550-31271f00-ae56-11eb-9f7d-4b3411d4bad7.png)

kevin-bates:
Ah - of course - this is completely helpful - thank you.  This file is located here: https://github.com/jupyter/enterprise_gateway/blob/master/etc/docker/enterprise-gateway/README.md, which we would then update on hub.

[...]

PR #1049, opened 2022-03-20
Docker docs

bloomsa:
Updating etc/docker/enterprise-gateway/README.md to account for switching from scripts to bring the containers up, to now using docker-compose

#968


kevin-bates (commented):
Thank you @bloomsa!  I think there are a couple of other GitHub links that do not reflect our move to the `jupyter-server` org that should probably get touched up while we're in here.  If you can make those corrections, that would be great.  Then I'll update the docker hub entry.

bloomsa:
sorry it took me a few days, pretty sure I got them all 👍

kevin-bates (approved):
Oh wow, I was just referring to the dockerhub-related README files, but this is fantastic - extra credit!  Thank you!

PR #1053, opened 2022-03-24
Add support for kernel-image-puller to pull images from private Docker registries

### LINKED ISSUE #759: kernel-image-puller is not supporting to pull images from private registries
kernel-image-puller is not supporting to pull images from private registries. I suppose that the best way would be to enable pulling images from private registries by passing KIP_IMAGE_REGISTRY_USER and KIP_IMAGE_REGISTRY_PASSWORD env variables... In a kubernetes setup this could be easily used in combination with a secret containing the user and login.

The python docker client supports this, using the auth_config parameter to docker.images.pull():

https://github.com/docker/docker-py/blob/master/docker/models/images.py

```

    def pull(self, repository, tag=None, **kwargs):
        Args:
            repository (str): The repository to pull
            tag (str): The tag to pull
            auth_config (dict): Override the credentials that are found in the
                config for this request.  ``auth_config`` should contain the
                ``username`` and ``password`` keys to be valid.
            platform (str): Platform in the format ``os[/arch[/variant]]``
```

mure:
Fixes #759

This adds a new value `kip.imagePullSecret` to the chart that is the name of a secret containing docker configuration. The .dockerconfigjson data from the secret is pulled out and mounted as a volume to the KIP container at /.docker. This allows the Python Docker client to automatically authenticate with a private registry. 

More info here: https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/

The value of `kip.imagePullSecret` can also be a template. This supports the use case where pull secrets are defined in global Helm values. In this case you could provide, for example, a value like:
```yaml
kip:
    imagePullSecret: "{{ Values.global.imagePullSecret }}"
```


etc/kubernetes/helm/enterprise-gateway/templates/daemonset.yaml
```diff
@@ -38,14 +38,26 @@ spec:
...
-        - name: KIP_AUTH_TOKEN   
+        - name: KIP_AUTH_TOKEN
           value: {{ .Values.authToken }}
         {{- end }}
         volumeMounts:
         - name: cri-socket
           mountPath: !!str {{ .Values.kip.criSocket }}  # see env KIP_CRI_ENDPOINT
+        {{- if .Values.kip.imagePullSecret }}
+        - name: dockerconfig
+          mountPath: "/root/.docker"
```
kevin-bates, line 49: I see that we're running this image as `root` (hadn't really noticed that before).  Does the mount have to be in the container user's HOME dir? (sorry, it seems the docs imply this, but I don't know if that's just an example). 
I wonder if this particular value should be templated as well via something like `.Values.kip.pullSecretMountPath` with this as the default value?

kevin-bates:
Hi @mure - this is really great - thank you!

Aside for the one comment I had, I'm hoping we could include whatever values come of this into the Kubernetes section of the [_Operators Guide_](https://jupyter-enterprise-gateway.readthedocs.io/en/latest/operators/deploy-kubernetes.html#configuration)?

I think it would also be helpful to add a small section titled something like `Private container registries` (I'm assuming this would work with other container types, is that correct?) to the [Kernel Image Puller](https://jupyter-enterprise-gateway.readthedocs.io/en/latest/operators/deploy-kubernetes.html#kernel-image-puller) section as well.

Thank you!

etc/kubernetes/helm/enterprise-gateway/templates/daemonset.yaml
```diff
@@ -38,14 +38,26 @@ spec:
...
       - name: cri-socket
         hostPath:
           path: {{ .Values.kip.criSocket }}
+      {{- if .Values.kip.imagePullSecret }}
+      - name: dockerconfig
+        secret:
+          secretName: {{ tpl .Values.kip.imagePullSecret . }}
+          items:
+          - key: .dockerconfigjson
+            path: config.json
```
rahul26goyal, line 61: Noob question: 
who is responsible for keeping this `config.json` file with valid token?

etc/kubernetes/helm/enterprise-gateway/templates/daemonset.yaml
rahul26goyal: this might help : https://www.devopsschool.com/blog/how-to-configure-docker-client-with-a-configuration-file/

kevin-bates: Thanks @rahul26goyal.  It looks like docker client supports a `DOCKER_CONFIG` env and I wonder if we should let that value dictate the `mountPath`?

etc/kubernetes/helm/enterprise-gateway/templates/daemonset.yaml
kevin-bates: @rahul26goyal - this will be provided by the user configuring `.Values.kip.imagePullSecret` and is a file in the mount location.

kevin-bates:
@mure - I've gone ahead and pushed the changes to resolve the conflicts.   Are you still working on this and can you please take a look at the current review comments when you get a chance?

mure:
@kevin-bates 
Thanks Kevin. I've read over the comments, but I haven't had a chance to update the PR yet.
We ended up running into another problem - if the image pull policy is set to 'Always', then the execution containers will still fail, due to this:

> Always
every time the kubelet launches a container, the kubelet queries the container image registry to resolve the name to an image [digest](https://docs.docker.com/engine/reference/commandline/pull/#pull-an-image-by-digest-immutable-identifier).

https://kubernetes.io/docs/concepts/containers/images/#image-pull-policy
So we need to either document the limitation or inject the pull secrets into new execution containers as well.

kevin-bates:
Thanks for the update @mure - no worries.
Regarding...
> inject the pull secrets into new execution containers as well.

By _execution containers_ are you referring to things like spark driver containers that need to spawn executor containers?  Could you please elaborate a bit?

darkstarmv:
@mure in kubernetes imagepull secrets are managed differently - see [doc](https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/).
Most of the time people configure service account to use imagepull secrets as it describer [here](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#add-image-pull-secret-to-service-account).
once  my pr https://github.com/jupyter-server/enterprise_gateway/pull/1073 is merged, you should be able to pull images from private registry.  I can also add creation of the k8s secret as an option

kevin-bates:
Hi @mure - are you in agree with @darkstarmv's https://github.com/jupyter-server/enterprise_gateway/pull/1053#issuecomment-1098430848?

kevin-bates:
Closing per https://github.com/jupyter-server/enterprise_gateway/pull/1053#issuecomment-1098430848.

Issue #1051, opened 2022-03-24
EG becomes unusable after receiving FD added twice error due to a race condition!


rahul26goyal:
## Description

This issue is related to #1047 

We are seeing ZMQStream FD leak happening on Jupyter Enterprise Gateway Server while running kernels on Kubernetes. We are running native spark Kubernetes kernels. 

Based on the analysis done below, the leak is happening at the jupyter application layer which integrates with Tornado IOLoop to manage the ZMQSocket Streams.  [add_handler](https://github.com/tornadoweb/tornado/blob/00c9e0ae31a5a0d12e09109fb77ffe391bfe1131/tornado/platform/asyncio.py#L148)

At an high level, the leak happens when a race condition between a Kernel Restart and Shutdown for the same kernel happens and this leads to a FD leak for a duration of 1 minute. 
- (1 min duration comes from `kernel_info_timeout`) [timeout](https://github.com/jupyter/notebook/blob/6.0.3/notebook/services/kernels/kernelmanager.py#L99)
- [restart_kernel](https://github.com/jupyter/notebook/blob/6.0.3/notebook/services/kernels/kernelmanager.py#L311)

This issue is seen only on **remote kernels** and not on the local kernels due to various differences that come with remote kernels.

**Sample exception trace from one such occurrence!**
```
[I 2022-02-23 04:23:06.531 EnterpriseGatewayApp] Kernel started: 4160c20e-c2ee-4df5-8c28-3716d135de5d
[I 220223 04:23:06 web:2243] 201 POST /api/kernels (127.0.0.1) 8313.27ms
[I 2022-02-23 04:23:07.566 EnterpriseGatewayApp] successfully validated request: /api/kernels/4160c20e-c2ee-4df5-8c28-3716d135de5d/channels
[W 2022-02-23 04:23:07.567 EnterpriseGatewayApp] No session ID specified
[E 220223 04:23:08 web:1793] Uncaught exception GET /api/kernels/4160c20e-c2ee-4df5-8c28-3716d135de5d/channels (172.18.96.19)
    HTTPServerRequest(protocol='https', host='localhost:18888', method='GET', uri='/api/kernels/4160c20e-c2ee-4df5-8c28-3716d135de5d/channels', version='HTTP/1.1', remote_ip='127.0.0.1')
    Traceback (most recent call last):
      File "/rhr/notebook-env/lib/python3.7/site-packages/tornado/websocket.py", line 954, in _accept_connection
[...]

PR #1054, opened 2022-03-25
Attempt to fix FD leak issue #1051

rahul26goyal:
## Description
This PR is raised wrt the fix proposed in #1051  to wait on the Kernel Shutdown request until the kernel Restart is completed. 
- the fix is made only on EG as it depends on the Kernel field set with in the EG code only: [kernel.restarting](https://github.com/jupyter-server/enterprise_gateway/blob/master/enterprise_gateway/services/kernels/remotemanager.py#L488)
-  There are some TODOs and open items which are open for discussion. (done)
- The time used to wait for restart to complete is inherited / taken from "KERNEL_LAUNCH_TIMEOUT" as this is precisely the time which even the kernel start would wait before giving up on the kernel.  

## How does this prevent the FD leak mentioned on #1051 ?
- As we are waiting on the restart to complete and no going to allow shutdown to happen in parallel, the chances of Shutdown flow terminating the kernel at the exact same time when Kernel restart is waiting on `kernel_info` is minimal. 
- This also prevents any other accidental leaks that might happen if a shutdown request is processed while Restart is happening. 

## TESTING 
- I have done some basic sanity testing with notebook and JupyterLab and  it looks to be working fine. 

- Tested the changes on local python kernel and changes seems to be working as expected. 
- when a restart request is followed by shutdown, the shutdown request waits until the restart is completed and then shuts down the kernel cleanly.
- when a restart request is followed by shutdown and the shutdown times out waiting, it proceeds to forceful shutdown. This was the same behaviour before the current changes were made.
- when multiple restart request is received for the same kernel, the first shutdown request goes the actual restart while all the other duplicate request wait for the `kernel.restarting` to become `False` and return. 
thanks!

### Test Scenario : Kernel is restarted via "auto restarter". And then kernel restart request is sent.

```
# the kernel was already running and we forcefully killed the kernel process.
# this trigger the auto-restarter 
 
AsyncIOLoopKernelRestarter: restarting kernel (1/5), keep random ports
kernel 0fb50ca9-b8ee-4030-9207-958ff3bab86a restarted
 
restarting kernel with value for now: True

Instantiating kernel 'Python 3 (ipykernel)' with process proxy: enterprise_gateway.services.processproxies.processproxy.LocalProcessProxy
Starting kernel (async): ['/Users/rhgoyal/miniconda3/envs/enterprise-gateway-dev/bin/python', '-m', 'ipykernel_launcher', '-f', '/Users/rhgoyal/Library/Jupyter/runtime/kernel-0fb50ca9-b8ee-4030-9207-958ff3bab86a.json']
Launching kernel: 'Python 3 (ipykernel)' with command: ['/Users/rhgoyal/miniconda3/envs/enterprise-gateway-dev/bin/python', '-m', 'ipykernel_launcher', '-f', '/Users/rhgoyal/Library/Jupyter/runtime/kernel-0fb50ca9-b8ee-4030-9207-958ff3bab86a.json']
BaseProcessProxy.launch_process() env: {'PATH': '/usr/local/opt/scala@2.12/bin:/Users/rhgoyal/miniconda3/envs/enterprise-gateway-dev/bin:/Users/rhgoyal/miniconda3/condabin:/Users/rhgoyal/miniconda3/bin:/Users/rhgoyal/learning/installs/apache-maven-3.6.3/bin:/Users/rhgoyal/.toolbox/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin', 'KERNEL_EXTRA_SPARK_OPTS': ' --conf spark.kubernetes.executor.label.component=worker ', 'KERNEL_LAUNCH_TIMEOUT': '150', 'KERNEL_WORKING_DIR': '/Users/rhgoyal/learning/k8s/jupyter', 'KERNEL_USERNAME': 'rhgoyal', 'KERNEL_GATEWAY': '1', 'KERNEL_ID': '0fb50ca9-b8ee-4030-9207-958ff3bab86a', 'KERNEL_LANGUAGE': 'python', 'EG_IMPERSONATION_ENABLED': 'False'}
Local kernel launched on '192.168.0.106', pid: 65658, pgid: 65658, KernelID: 0fb50ca9-b8ee-4030-9207-958ff3bab86a, cmd: '['/Users/rhgoyal/miniconda3/envs/enterprise-gateway-dev/bin/python', '-m', 'ipykernel_launcher', '-f', '/Users/rhgoyal/Library/Jupyter/runtime/kernel-0fb50ca9-b8ee-4030-9207-958ff3bab86a.json']'
Connecting to: tcp://127.0.0.1:51321
Refreshing kernel session for id: 0fb50ca9-b8ee-4030-9207-958ff3bab86a
activity on 0fb50ca9-b8ee-4030-9207-958ff3bab86a: status (busy)
activity on 0fb50ca9-b8ee-4030-9207-958ff3bab86a: status (idle)
AsyncIOLoopKernelRestarter: restart apparently succeeded



Websocket closed 0fb50ca9-b8ee-4030-9207-958ff3bab86a:d78a447e-56649b2aad7998e7efa537be
Starting buffering for 0fb50ca9-b8ee-4030-9207-958ff3bab86a:d78a447e-56649b2aad7998e7efa537be
Clearing buffer for 0fb50ca9-b8ee-4030-9207-958ff3bab86a
Initializing websocket connection /api/kernels/0fb50ca9-b8ee-4030-9207-958ff3bab86a/channels
No session ID specified
Opening websocket /api/kernels/0fb50ca9-b8ee-4030-9207-958ff3bab86a/channels
Getting buffer for 0fb50ca9-b8ee-4030-9207-958ff3bab86a
Clearing buffer for 0fb50ca9-b8ee-4030-9207-958ff3bab86a
Connecting to: tcp://127.0.0.1:51318
Connecting to: tcp://127.0.0.1:51317
Connecting to: tcp://127.0.0.1:51321
Connecting to: tcp://127.0.0.1:51319
Connecting to: tcp://127.0.0.1:51317
Connecting to: tcp://127.0.0.1:51321
Nudge: attempt 1 on kernel 0fb50ca9-b8ee-4030-9207-958ff3bab86a
Nudge: IOPub received: 0fb50ca9-b8ee-4030-9207-958ff3bab86a
Nudge: resolving iopub future: 0fb50ca9-b8ee-4030-9207-958ff3bab86a
activity on 0fb50ca9-b8ee-4030-9207-958ff3bab86a: status (busy)
activity on 0fb50ca9-b8ee-4030-9207-958ff3bab86a: status (idle)
Nudge: shell info reply received: 0fb50ca9-b8ee-4030-9207-958ff3bab86a
Nudge: resolving shell future: 0fb50ca9-b8ee-4030-9207-958ff3bab86a
activity on 0fb50ca9-b8ee-4030-9207-958ff3bab86a: status (busy)
activity on 0fb50ca9-b8ee-4030-9207-958ff3bab86a: status (idle)
activity on 0fb50ca9-b8ee-4030-9207-958ff3bab86a: status (busy)
activity on 0fb50ca9-b8ee-4030-9207-958ff3bab86a: status (idle)
activity on 0fb50ca9-b8ee-4030-9207-958ff3bab86a: status (busy)
activity on 0fb50ca9-b8ee-4030-9207-958ff3bab86a: comm_close
activity on 0fb50ca9-b8ee-4030-9207-958ff3bab86a: status (idle)
activity on 0fb50ca9-b8ee-4030-9207-958ff3bab86a: status (busy)
activity on 0fb50ca9-b8ee-4030-9207-958ff3bab86a: status (idle)




Websocket closed 0fb50ca9-b8ee-4030-9207-958ff3bab86a:93c6b3dd-fe2911af8651bfe9c58bd8f9
Starting buffering for 0fb50ca9-b8ee-4030-9207-958ff3bab86a:93c6b3dd-fe2911af8651bfe9c58bd8f9
Clearing buffer for 0fb50ca9-b8ee-4030-9207-958ff3bab86a

# received restart from User / UI
Current value of the kernel.restarting: False
Going ahead to process kernel restart request.
restarting kernel with value for now: False
activity on 0fb50ca9-b8ee-4030-9207-958ff3bab86a: status (busy)
activity on 0fb50ca9-b8ee-4030-9207-958ff3bab86a: shutdown_reply
Buffering msg on 0fb50ca9-b8ee-4030-9207-958ff3bab86a:iopub
activity on 0fb50ca9-b8ee-4030-9207-958ff3bab86a: status (idle)
Buffering msg on 0fb50ca9-b8ee-4030-9207-958ff3bab86a:iopub
Buffering msg on 0fb50ca9-b8ee-4030-9207-958ff3bab86a:iopub
Instantiating kernel 'Python 3 (ipykernel)' with process proxy: enterprise_gateway.services.processproxies.processproxy.LocalProcessProxy
Starting kernel (async): ['/Users/rhgoyal/miniconda3/envs/enterprise-gateway-dev/bin/python', '-m', 'ipykernel_launcher', '-f', '/Users/rhgoyal/Library/Jupyter/runtime/kernel-0fb50ca9-b8ee-4030-9207-958ff3bab86a.json']
Launching kernel: 'Python 3 (ipykernel)' with command: ['/Users/rhgoyal/miniconda3/envs/enterprise-gateway-dev/bin/python', '-m', 'ipykernel_launcher', '-f', '/Users/rhgoyal/Library/Jupyter/runtime/kernel-0fb50ca9-b8ee-4030-9207-958ff3bab86a.json']
BaseProcessProxy.launch_process() env: {'PATH': '/usr/local/opt/scala@2.12/bin:/Users/rhgoyal/miniconda3/envs/enterprise-gateway-dev/bin:/Users/rhgoyal/miniconda3/condabin:/Users/rhgoyal/miniconda3/bin:/Users/rhgoyal/learning/installs/apache-maven-3.6.3/bin:/Users/rhgoyal/.toolbox/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin', 'KERNEL_EXTRA_SPARK_OPTS': ' --conf spark.kubernetes.executor.label.component=worker ', 'KERNEL_LAUNCH_TIMEOUT': '150', 'KERNEL_WORKING_DIR': '/Users/rhgoyal/learning/k8s/jupyter', 'KERNEL_USERNAME': 'rhgoyal', 'KERNEL_GATEWAY': '1', 'KERNEL_ID': '0fb50ca9-b8ee-4030-9207-958ff3bab86a', 'KERNEL_LANGUAGE': 'python', 'EG_IMPERSONATION_ENABLED': 'False'}
Local kernel launched on '192.168.0.106', pid: 65678, pgid: 65678, KernelID: 0fb50ca9-b8ee-4030-9207-958ff3bab86a, cmd: '['/Users/rhgoyal/miniconda3/envs/enterprise-gateway-dev/bin/python', '-m', 'ipykernel_launcher', '-f', '/Users/rhgoyal/Library/Jupyter/runtime/kernel-0fb50ca9-b8ee-4030-9207-958ff3bab86a.json']'
Connecting to: tcp://127.0.0.1:51321
Refreshing kernel session for id: 0fb50ca9-b8ee-4030-9207-958ff3bab86a
Kernel restarted: 0fb50ca9-b8ee-4030-9207-958ff3bab86a
Connecting to: tcp://127.0.0.1:51317
Resetting kernel.restarting flag to False.
Initializing websocket connection /api/kernels/0fb50ca9-b8ee-4030-9207-958ff3bab86a/channels
No session ID specified
Opening websocket /api/kernels/0fb50ca9-b8ee-4030-9207-958ff3bab86a/channels
Getting buffer for 0fb50ca9-b8ee-4030-9207-958ff3bab86a
Clearing buffer for 0fb50ca9-b8ee-4030-9207-958ff3bab86a
Discarding 3 buffered messages for 0fb50ca9-b8ee-4030-9207-958ff3bab86a:93c6b3dd-fe2911af8651bfe9c58bd8f9
Connecting to: tcp://127.0.0.1:51318
Connecting to: tcp://127.0.0.1:51317
Connecting to: tcp://127.0.0.1:51321
Connecting to: tcp://127.0.0.1:51319
Connecting to: tcp://127.0.0.1:51317
Connecting to: tcp://127.0.0.1:51321
Nudge: attempt 1 on kernel 0fb50ca9-b8ee-4030-9207-958ff3bab86a
Nudge: attempt 2 on kernel 0fb50ca9-b8ee-4030-9207-958ff3bab86a
Nudge: IOPub received: 0fb50ca9-b8ee-4030-9207-958ff3bab86a
Nudge: resolving iopub future: 0fb50ca9-b8ee-4030-9207-958ff3bab86a
Nudge: shell info reply received: 0fb50ca9-b8ee-4030-9207-958ff3bab86a
Nudge: resolving shell future: 0fb50ca9-b8ee-4030-9207-958ff3bab86a
activity on 0fb50ca9-b8ee-4030-9207-958ff3bab86a: status (busy)
activity on 0fb50ca9-b8ee-4030-9207-958ff3bab86a: status (idle)
Kernel info reply received: 0fb50ca9-b8ee-4030-9207-958ff3bab86a
```

### DIFF
--- a/enterprise_gateway/services/kernels/remotemanager.py
+++ b/enterprise_gateway/services/kernels/remotemanager.py
@@ -20,7 +20,7 @@ from ..processproxies.processproxy import LocalProcessProxy, RemoteProcessProxy
 from ..sessions.kernelsessionmanager import KernelSessionManager
 
 default_kernel_launch_timeout = float(os.getenv("EG_KERNEL_LAUNCH_TIMEOUT", "30"))
-
+kernel_restart_finish_poll_internal = float(os.getenv("EG_RESTART_FINISH_POLL_INTERVAL", 1.0))
 
 def import_item(name):
     """Import and return ``bar`` given the string ``foo.bar``.
@@ -194,25 +194,18 @@ class RemoteMappingKernelManager(AsyncMappingKernelManager):
 
     async def restart_kernel(self, kernel_id):
         kernel = self.get_kernel(kernel_id)
-        self.log.debug(f"Current value of the kernel.restarting: {kernel.restarting}")
         if kernel.restarting:  # assuming duplicate request.
             await self.wait_for_restart_finish(kernel_id, "restart")
-            self.log.info(
-                f"Done with waiting for restart to complete. Current value of kernel.restarting: {kernel.restarting}. "
-                f"Skipping kernel restart."
-            )
+            self.log.info("Skipping kernel restart as this was duplicate request.")
             return
-        self.log.info("Going ahead to process kernel restart request.")
         try:
             kernel.restarting = True  # Moved in out of RemoteKernelManager
             await super().restart_kernel(kernel_id)
         finally:
-            self.log.debug("Resetting kernel.restarting flag to False.")
             kernel.restarting = False
 
     async def shutdown_kernel(self, kernel_id, now=False, restart=False):
         kernel = self.get_kernel(kernel_id)
-        self.log.debug(f"Current value of the Kernel Restarting: {kernel.restarting}")
         if kernel.restarting:
             await self.wait_for_restart_finish(kernel_id, "shutdown")
         try:
@@ -224,23 +217,19 @@ class RemoteMappingKernelManager(AsyncMappingKernelManager):
         kernel = self.get_kernel(kernel_id)
         start_time = float(time.time())  # epoc time
         timeout = kernel.kernel_launch_timeout
-        poll_time = 1  # TODO can be read from a new config if need be.
+        poll_time = kernel_restart_finish_poll_internal
         self.log.info(
-            f"kernel is restarting when {action} request received. Polling every {poll_time} "
-            f"seconds for next {timeout} seconds for kernel '{kernel_id}'"
-            f" to complete its restart."
+            f"Kernel '{kernel_id}' was restarting when {action} request received. Polling every {poll_time} "
+            f"seconds for next {timeout} seconds for kernel to complete its restart."
         )
         while kernel.restarting:
             now = float(time.time())
             if (now - start_time) > timeout:
                 self.log.info(
-                    f"Wait_Timeout: Existing out of the restart wait loop to {action} kernel."
+                    f"Timeout: Exiting restart wait loop in order to {action} kernel '{kernel_id}'."
                 )
                 break
             await asyncio.sleep(poll_time)
-        self.log.info(
-            f"Returning with current value of the kernel.restarting: {kernel.restarting}."
-        )
         return
 
     def _enforce_kernel_limits(self, username: str) -> None:
@@ -562,9 +551,7 @@ class RemoteKernelManager(EnterpriseGatewayConfigMixin, AsyncIOLoopKernelManager
             Any options specified here will overwrite those used to launch the
             kernel.
         """
-        # this is added for only auto-restarter as it directly call this method.
-        self.log.info(f"restarting kernel with value for now: {now}")
-        if now:
+        if now:  # if auto-restarting (when now is True), indicate we're restarting.
             self.restarting = True
         kernel_id = self.kernel_id or os.path.basename(self.connection_file).replace(
             "kernel-", ""
--- a/enterprise_gateway/services/kernels/remotemanager.py
+++ b/enterprise_gateway/services/kernels/remotemanager.py
@@ -562,6 +562,10 @@ class RemoteKernelManager(EnterpriseGatewayConfigMixin, AsyncIOLoopKernelManager
             Any options specified here will overwrite those used to launch the
             kernel.
         """
+        # this is added for only auto-restarter as it directly call this method.
+        self.log.info(f"restarting kernel with value for now: {now}")
+        if now:
+            self.restarting = True
         kernel_id = self.kernel_id or os.path.basename(self.connection_file).replace(
             "kernel-", ""
         ).replace(".json", "")
@@ -592,6 +596,8 @@ class RemoteKernelManager(EnterpriseGatewayConfigMixin, AsyncIOLoopKernelManager
         # Refresh persisted state.
         if self.kernel_session_manager:
             self.kernel_session_manager.refresh_session(kernel_id)
+        if now:
+            self.restarting = False
 
     async def signal_kernel(self, signum):
         """
--- a/enterprise_gateway/services/kernels/remotemanager.py
+++ b/enterprise_gateway/services/kernels/remotemanager.py
@@ -1,6 +1,9 @@
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
 """Kernel managers that operate against a remote process."""
+import time
+
+import asyncio
 
 import os
 import re
@@ -18,6 +21,7 @@ from enterprise_gateway.mixins import EnterpriseGatewayConfigMixin
 from ..processproxies.processproxy import LocalProcessProxy, RemoteProcessProxy
 from ..sessions.kernelsessionmanager import KernelSessionManager
 
+default_kernel_launch_timeout = float(os.getenv('EG_KERNEL_LAUNCH_TIMEOUT', '30'))
 
 def import_item(name):
     """Import and return ``bar`` given the string ``foo.bar``.
@@ -189,6 +193,50 @@ class RemoteMappingKernelManager(AsyncMappingKernelManager):
         self.parent.kernel_session_manager.create_session(kernel_id, **kwargs)
         return kernel_id
 
+    async def restart_kernel(self, kernel_id):
+        kernel = self.get_kernel(kernel_id)
+        self.log.debug(f"Current value of the Kernel Restarting: {kernel.restarting}")
+        if kernel.restarting:  # assuming duplicate request.
+            await self.wait_and_poll_for_restart_to_complete(kernel_id, "restart")
+            self.log.info(
+                f"Done with waiting for restart to complete. Current value of restarting: {kernel.restarting}. "
+                f"Skipping restarting kernel.")
+            return
+        self.log.info("Going ahead to process restart kernel request.")
+        try:
+            kernel.restarting = True  # Moved in out of RemoteKernelManager
+            await super(RemoteMappingKernelManager, self).restart_kernel(kernel_id)
+        finally:
+            self.log.debug("Resetting `restarting` flag to False.")
+            kernel.restarting = False
+
+    async def shutdown_kernel(self, kernel_id, now=False, restart=False):
+        kernel = self.get_kernel(kernel_id)
+        self.log.debug(f"Current value of the Kernel Restarting: {kernel.restarting}")
+        if kernel.restarting:
+            await self.wait_and_poll_for_restart_to_complete(kernel_id, "shutdown")
+        try:
+            await super(RemoteMappingKernelManager, self).shutdown_kernel(kernel_id, now, restart)
+        except KeyError as ke:  # this is hit for multiple shutdown request.
+            self.log.exception(f"Exception while shutting Kernel: {kernel_id}: {ke}")
+
+    async def wait_and_poll_for_restart_to_complete(self, kernel_id, action="shutdown"):
+        kernel = self.get_kernel(kernel_id)
+        start_time = int(time.time())  # epoc time in seconds
+        timeout = kernel.kernel_launch_timeout
+        poll_time = 1  # TODO can be read from a new config.
+        self.log.info(f"Kernel is restarting when {action} request received. Polling every {poll_time} "
+                      f"seconds for next {timeout} seconds for kernel '{kernel_id}'"
+                      f" to complete its restart.")
+        while kernel.restarting:
+            now = int(time.time())
+            if (now - start_time) > timeout:
+                self.log.info(f"Wait_Timeout: Existing out of the restart wait loop to {action} kernel.")
+                break
+            await asyncio.sleep(poll_time)
+        self.log.info(f"Returning with current value of the Kernel Restarting: {kernel.restarting}.")
+        return
+
     def _enforce_kernel_limits(self, username: str) -> None:
         """
         If MaxKernels or MaxKernelsPerUser are configured, enforce the respective values.
@@ -341,6 +389,7 @@ class RemoteKernelManager(EnterpriseGatewayConfigMixin, AsyncIOLoopKernelManager
         self.sigint_value = None
         self.kernel_id = None
         self.user_overrides = {}
+        self.kernel_launch_timeout = default_kernel_launch_timeout
         self.restarting = False  # need to track whether we're in a restart situation or not
 
         # If this instance supports port caching, then disable cache_ports since we don't need this
@@ -412,6 +461,9 @@ class RemoteKernelManager(EnterpriseGatewayConfigMixin, AsyncIOLoopKernelManager
         of the kernelspec env stanza that would have otherwise overridden the user-provided values.
         """
         env = kwargs.get("env", {})
+        # see if KERNEL_LAUNCH_TIMEOUT was included from user.  If so, override default
+        if env.get('KERNEL_LAUNCH_TIMEOUT', None):
+            self.kernel_launch_timeout = float(env.get('KERNEL_LAUNCH_TIMEOUT'))
         self.user_overrides.update(
             {
                 key: value
@@ -504,7 +556,6 @@ class RemoteKernelManager(EnterpriseGatewayConfigMixin, AsyncIOLoopKernelManager
             Any options specified here will overwrite those used to launch the
             kernel.
         """
-        self.restarting = True
         kernel_id = self.kernel_id or os.path.basename(self.connection_file).replace(
             "kernel-", ""
         ).replace(".json", "")
@@ -535,7 +586,6 @@ class RemoteKernelManager(EnterpriseGatewayConfigMixin, AsyncIOLoopKernelManager
         # Refresh persisted state.
         if self.kernel_session_manager:
             self.kernel_session_manager.refresh_session(kernel_id)
-        self.restarting = False
 
     async def signal_kernel(self, signum):
         """
--- a/enterprise_gateway/services/kernels/remotemanager.py
+++ b/enterprise_gateway/services/kernels/remotemanager.py
@@ -20,7 +20,9 @@ from ..processproxies.processproxy import LocalProcessProxy, RemoteProcessProxy
 from ..sessions.kernelsessionmanager import KernelSessionManager
 
 default_kernel_launch_timeout = float(os.getenv("EG_KERNEL_LAUNCH_TIMEOUT", "30"))
-kernel_restart_finish_poll_internal = float(os.getenv("EG_RESTART_FINISH_POLL_INTERVAL", 1.0))
+kernel_restart_status_poll_interval = float(os.getenv("EG_RESTART_STATUS_POLL_INTERVAL", 1.0))
+
+
 
 def import_item(name):
     """Import and return ``bar`` given the string ``foo.bar``.
@@ -211,13 +213,14 @@ class RemoteMappingKernelManager(AsyncMappingKernelManager):
         try:
             await super().shutdown_kernel(kernel_id, now, restart)
         except KeyError as ke:  # this is hit for multiple shutdown request.
-            self.log.exception(f"Exception while shutting Kernel: {kernel_id}: {ke}")
+            self.log.exception(f"Exception while shutting down kernel: '{kernel_id}': {ke}")
+            raise web.HTTPError(404, "Kernel does not exist: %s" % kernel_id)
 
     async def wait_for_restart_finish(self, kernel_id, action="shutdown"):
         kernel = self.get_kernel(kernel_id)
         start_time = float(time.time())  # epoc time
         timeout = kernel.kernel_launch_timeout
-        poll_time = kernel_restart_finish_poll_internal
+        poll_time = kernel_restart_status_poll_interval
         self.log.info(
             f"Kernel '{kernel_id}' was restarting when {action} request received. Polling every {poll_time} "
             f"seconds for next {timeout} seconds for kernel to complete its restart."
@@ -261,7 +264,7 @@ class RemoteMappingKernelManager(AsyncMappingKernelManager):
             if self.parent.max_kernels_per_user >= 0:
                 if self.parent.kernel_session_manager:
                     active_and_pending = (
-                        self.parent.kernel_session_manager.active_sessions(username) + pending_user
+                            self.parent.kernel_session_manager.active_sessions(username) + pending_user
                     )
                     if active_and_pending >= self.parent.max_kernels_per_user:
                         error_message = (
@@ -285,7 +288,7 @@ class RemoteMappingKernelManager(AsyncMappingKernelManager):
         self.parent.kernel_session_manager.delete_session(kernel_id)
 
     def start_kernel_from_session(
-        self, kernel_id, kernel_name, connection_info, process_info, launch_args
+            self, kernel_id, kernel_name, connection_info, process_info, launch_args
     ):
         """
         Starts a kernel from a persisted kernel session.
@@ -456,9 +459,8 @@ class RemoteKernelManager(EnterpriseGatewayConfigMixin, AsyncIOLoopKernelManager
         of the kernelspec env stanza that would have otherwise overridden the user-provided values.
         """
         env = kwargs.get("env", {})
-        # see if KERNEL_LAUNCH_TIMEOUT was included from user.  If so, override default
-        if env.get("KERNEL_LAUNCH_TIMEOUT", None):
-            self.kernel_launch_timeout = float(env.get("KERNEL_LAUNCH_TIMEOUT"))
+        # If KERNEL_LAUNCH_TIMEOUT is passed in the payload, override it.
+        self.kernel_launch_timeout = float(env.get("KERNEL_LAUNCH_TIMEOUT", default_kernel_launch_timeout))
         self.user_overrides.update(
             {
                 key: value
--- a/enterprise_gateway/services/kernels/remotemanager.py
+++ b/enterprise_gateway/services/kernels/remotemanager.py
@@ -23,7 +23,6 @@ default_kernel_launch_timeout = float(os.getenv("EG_KERNEL_LAUNCH_TIMEOUT", "30"
 kernel_restart_status_poll_interval = float(os.getenv("EG_RESTART_STATUS_POLL_INTERVAL", 1.0))
 
 
-
 def import_item(name):
     """Import and return ``bar`` given the string ``foo.bar``.
     Calling ``bar = import_item("foo.bar")`` is the functional equivalent of
@@ -264,7 +263,7 @@ class RemoteMappingKernelManager(AsyncMappingKernelManager):
             if self.parent.max_kernels_per_user >= 0:
                 if self.parent.kernel_session_manager:
                     active_and_pending = (
-                            self.parent.kernel_session_manager.active_sessions(username) + pending_user
+                        self.parent.kernel_session_manager.active_sessions(username) + pending_user
                     )
                     if active_and_pending >= self.parent.max_kernels_per_user:
                         error_message = (
@@ -288,7 +287,7 @@ class RemoteMappingKernelManager(AsyncMappingKernelManager):
         self.parent.kernel_session_manager.delete_session(kernel_id)
 
     def start_kernel_from_session(
-            self, kernel_id, kernel_name, connection_info, process_info, launch_args
+        self, kernel_id, kernel_name, connection_info, process_info, launch_args
     ):
         """
         Starts a kernel from a persisted kernel session.
@@ -460,7 +459,9 @@ class RemoteKernelManager(EnterpriseGatewayConfigMixin, AsyncIOLoopKernelManager
         """
         env = kwargs.get("env", {})
         # If KERNEL_LAUNCH_TIMEOUT is passed in the payload, override it.
-        self.kernel_launch_timeout = float(env.get("KERNEL_LAUNCH_TIMEOUT", default_kernel_launch_timeout))
+        self.kernel_launch_timeout = float(
+            env.get("KERNEL_LAUNCH_TIMEOUT", default_kernel_launch_timeout)
+        )
         self.user_overrides.update(
             {
                 key: value
--- a/enterprise_gateway/services/kernels/remotemanager.py
+++ b/enterprise_gateway/services/kernels/remotemanager.py
@@ -1,13 +1,11 @@
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
 """Kernel managers that operate against a remote process."""
-import time
-
 import asyncio
-
 import os
 import re
 import signal
+import time
 import uuid
 
 from jupyter_client.ioloop.manager import AsyncIOLoopKernelManager
@@ -21,7 +19,7 @@ from enterprise_gateway.mixins import EnterpriseGatewayConfigMixin
 from ..processproxies.processproxy import LocalProcessProxy, RemoteProcessProxy
 from ..sessions.kernelsessionmanager import KernelSessionManager
 
-default_kernel_launch_timeout = float(os.getenv('EG_KERNEL_LAUNCH_TIMEOUT', '30'))
+default_kernel_launch_timeout = float(os.getenv("EG_KERNEL_LAUNCH_TIMEOUT", "30"))
 
 
 def import_item(name):
@@ -201,12 +199,13 @@ class RemoteMappingKernelManager(AsyncMappingKernelManager):
             await self.wait_for_restart_finish(kernel_id, "restart")
             self.log.info(
                 f"Done with waiting for restart to complete. Current value of kernel.restarting: {kernel.restarting}. "
-                f"Skipping kernel restart.")
+                f"Skipping kernel restart."
+            )
             return
         self.log.info("Going ahead to process kernel restart request.")
         try:
             kernel.restarting = True  # Moved in out of RemoteKernelManager
-            await super(RemoteMappingKernelManager, self).restart_kernel(kernel_id)
+            await super().restart_kernel(kernel_id)
         finally:
             self.log.debug("Resetting kernel.restarting flag to False.")
             kernel.restarting = False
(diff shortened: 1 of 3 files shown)


kevin-bates (commented):
Thanks for the PR @rahul26goyal.  As I mentioned on the issue, it would be great if you could spend some time with EG 2.6 to reproduce this.  This PR is essentially a poll loop and strikes me as analogous to the "pending kernels" work in jupyter_client >= 7 contains (and EG can't leverage).  That said, if we find it helps your situation, it seems fairly harmless to include.

rahul26goyal:
Thanks a lot for the review comments @kevin-bates. I will go over the comments today and address those.
On Testing side: 
1. While testing the earlier fix, we found issue with duplicate "Restart" request which could cause the same race condition.  Added the code change to handle the same.
Thanks!

enterprise_gateway/services/kernels/handlers.py
```diff
@@ -133,6 +135,33 @@ def get(self, kernel_id):
...
+            self.log.info(f"Going to Poll every {poll_time} seconds for next {timeout} "
+                          f"seconds for Kernel to come out of restart.")
+            while kernel.restarting:
+                now = int(time.time())
+                if (now - start_time) > timeout:
+                    self.log.info("Existing out of the shutdown wait loop to terminate the kernel anyways.")
+                    break
+                self.log.info(f"going to sleep for {poll_time}")  # TODO remove  this.
+                await asyncio.sleep(poll_time)
+                time.sleep(5)
```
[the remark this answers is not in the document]
rahul26goyal: my bad..it was a test code which slipped into this PR.

enterprise_gateway/services/kernels/handlers.py
```diff
@@ -133,6 +135,33 @@ def get(self, kernel_id):
...
+        self.log.info(f"Received Shutdown for Kernel : {kernel_id}")
+        km = self.kernel_manager
+        km.check_kernel_id(kernel_id)
+        kernel = km.get_kernel(kernel_id)
+        if kernel.restarting:
+            start_time = int(time.time())  # epoc time in seconds
+            timeout = km.kernel_info_timeout  # this could be set to kernel launch timeout to be in sync!
+            poll_time = 5  # we can make this configurable
+            self.log.info(f"Going to Poll every {poll_time} seconds for next {timeout} "
+                          f"seconds for Kernel to come out of restart.")
```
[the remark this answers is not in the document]
rahul26goyal: sounds good 👍

enterprise_gateway/services/kernels/handlers.py
```diff
@@ -133,6 +135,33 @@ def get(self, kernel_id):
         model = km.kernel_model(kernel_id)
         self.finish(json.dumps(model, default=date_default))
 
+    @web.authenticated
+    async def delete(self, kernel_id):
+        self.log.info(f"Received Shutdown for Kernel : {kernel_id}")
```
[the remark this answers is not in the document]
rahul26goyal: sure

enterprise_gateway/services/kernels/handlers.py
```diff
@@ -133,6 +135,33 @@ def get(self, kernel_id):
...
+        km = self.kernel_manager
+        km.check_kernel_id(kernel_id)
+        kernel = km.get_kernel(kernel_id)
+        if kernel.restarting:
+            start_time = int(time.time())  # epoc time in seconds
+            timeout = km.kernel_info_timeout  # this could be set to kernel launch timeout to be in sync!
+            poll_time = 5  # we can make this configurable
```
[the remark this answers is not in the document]
rahul26goyal: agree. 
1. We can have smaller default value for `poll_time`  to detect restart sooner while providing an override option.

enterprise_gateway/services/kernels/handlers.py
```diff
@@ -133,6 +135,33 @@ def get(self, kernel_id):
...
 
+    @web.authenticated
+    async def delete(self, kernel_id):
+        self.log.info(f"Received Shutdown for Kernel : {kernel_id}")
+        km = self.kernel_manager
+        km.check_kernel_id(kernel_id)
+        kernel = km.get_kernel(kernel_id)
+        if kernel.restarting:
+            start_time = int(time.time())  # epoc time in seconds
+            timeout = km.kernel_info_timeout  # this could be set to kernel launch timeout to be in sync!
```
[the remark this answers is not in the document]
rahul26goyal: I have tried incorporated this fix by moving the `kernel_launch_timeout` as a Kernel property which seems the right place for it?

kevin-bates:
Hi @rahul26goyal.  Is there a reason you chose to add the duplicate restart logic in the `RemoteMappingKernelManager` rather than in the kernel action handler (similar to what you did for shutdown)?  There's a definite pattern taking shape here and I hope we can move this kind of code into a comment function of sorts.

rahul26goyal:
> Hi @rahul26goyal. Is there a reason you chose to add the duplicate restart logic in the `RemoteMappingKernelManager` rather than in the kernel action handler (similar to what you did for shutdown)? There's a definite pattern taking shape here and I hope we can move this kind of code into a common function of sorts.

I agree that its a duplicate code and a definite pattern being followed,
There are 2 reasons I wrote it this way: 
1. The `remove_kernel` method in `RemoteMappingKernelManager` or any other Manager is not a async function. So, I guess it would not allow any `asyncio` stuff there? Thus, I implemented the `polling` in the handler which is an async method. 

2. The current list of Kernel handlers on EG did not extend `KernelActionHandler` which handles the restart request. So, I ended up writing it in the `RemoteMappingKernelManager::restart_kernel` which looked like a good place. 

If we can find a solution for Point 1, I could reuse the pattern and implement the logic to poll in `RemoteMappingKernelManager` which both shutdown and restart can use.

If Point 1 can not be solved, than I can extend the  `KernelActionHandler` in EG  and write the polling logic there.
 
Please let me know on how can proceed further.

kevin-bates:
I've been giving this some thought over the last couple of days.  Basically, this logic should be colocated and the question is where should that occur.

I view this PR as another implementation of _pending kernels_, but one that doesn't use a Future to determine that the _pending_ portion of things has completed.  Since EG cannot leverage that functionality (at this time) we're faced with rolling our own temporarily until we can leverage pending kernels.  (This will require EG's transition to kernel provisioners.)

So, if we were to roll our own temporary solution, it seems clear that we'd want to do this in similar locations as what is done in jupyter_client v7 (albeit in the KernelManager _subclasses_ we implement) and, preferably, in a nearly identical manner if possible.  This last comment implies the use of AsyncKernelManager.

The EG KernelManager seems to be right colocation IMO because its the EG KernelManager overrides that will essentially dissolve once kernel provisioner support is achieved.  Yes, we will likely still have subclasses to accomplish "enterprise" kinds of functionality - like HA/DR, load-balancing, etc.

This all said, I think it would be good to get together and hash this out further.  Let's chat via gitter and come up with a time we can meet.  Thank you.

enterprise_gateway/services/kernels/remotemanager.py
```diff
@@ -485,7 +537,6 @@ async def restart_kernel(self, now=False, **kwargs):
             Any options specified here will overwrite those used to launch the
             kernel.
         """
-        self.restarting = True
```
kevin-bates, line 488: I think these (set and unset) need to remain in the `RemoteKernelManager` since things like auto-restart can trigger a restart that is not initiated by the user.

enterprise_gateway/services/kernels/remotemanager.py
```diff
@@ -186,6 +190,50 @@ async def start_kernel(self, *args, **kwargs):
...
+        if kernel.restarting:
+            await self.wait_and_poll_for_restart_to_complete(kernel_id, "shutdown")
+        try:
+            await super(RemoteMappingKernelManager, self).shutdown_kernel(kernel_id, now, restart)
+        except KeyError as ke:  # this is hit for multiple shutdown request.
+            self.log.exception(f"Exception while shutting Kernel: {kernel_id}: {ke}")
+
+    async def wait_and_poll_for_restart_to_complete(self, kernel_id, action="shutdown"):
+        kernel = self.get_kernel(kernel_id)
+        start_time = int(time.time())  # epoc time in seconds
```
kevin-bates, line 222: We should tolerate milliseconds since we'll want to allow for a sub-second interval.
```suggestion
        start_time: float = time.time()  # epoc time
```

enterprise_gateway/services/kernels/remotemanager.py
```diff
@@ -186,6 +190,50 @@ async def start_kernel(self, *args, **kwargs):
...
+        kernel = self.get_kernel(kernel_id)
+        self.log.debug(f"Current value of the Kernel Restarting: {kernel.restarting}")
+        if kernel.restarting:
+            await self.wait_and_poll_for_restart_to_complete(kernel_id, "shutdown")
+        try:
+            await super(RemoteMappingKernelManager, self).shutdown_kernel(kernel_id, now, restart)
+        except KeyError as ke:  # this is hit for multiple shutdown request.
+            self.log.exception(f"Exception while shutting Kernel: {kernel_id}: {ke}")
+
+    async def wait_and_poll_for_restart_to_complete(self, kernel_id, action="shutdown"):
```
kevin-bates, line 220: This seems a little verbose, perhaps something more like:
```suggestion
    async def wait_for_restart(self, kernel_id: str, action:str = "shutdown"):
```

enterprise_gateway/services/kernels/remotemanager.py
```diff
@@ -186,6 +190,50 @@ async def start_kernel(self, *args, **kwargs):
         self.parent.kernel_session_manager.create_session(kernel_id, **kwargs)
         return kernel_id
 
+    async def restart_kernel(self, kernel_id):
+        kernel = self.get_kernel(kernel_id)
+        self.log.debug(f"Current value of the Kernel Restarting: {kernel.restarting}")
```
kevin-bates, line 195: Could we please lowercase `Kernel` and `Restarting` in its various places?  Might as well reference the variable names as well.
```suggestion
        self.log.debug(f"Current value of the 'kernel.restarting': {kernel.restarting}")
```

enterprise_gateway/services/kernels/remotemanager.py
rahul26goyal: > 

@kevin-bates : we tested this on v2.6.0 and still this issue is seen with Kubernetes  Kernels.

rahul26goyal:
hi @kevin-bates : 
we tested the race condition on v2.6.0 without this fix and we still see the FD leak happening with Kubernetes Kernels.

enterprise_gateway/services/kernels/remotemanager.py
```diff
@@ -189,6 +192,57 @@ async def start_kernel(self, *args, **kwargs):
...
 
+    async def restart_kernel(self, kernel_id):
+        kernel = self.get_kernel(kernel_id)
+        self.log.debug(f"Current value of the kernel.restarting: {kernel.restarting}")
+        if kernel.restarting:  # assuming duplicate request.
+            await self.wait_for_restart_finish(kernel_id, "restart")
+            self.log.info(
+                f"Done with waiting for restart to complete. Current value of kernel.restarting: {kernel.restarting}. "
+                f"Skipping kernel restart."
+            )
```
kevin-bates, line 203: Let's remove this.  We can infer this from any logging the `wait_for_restart_finish()` does.

enterprise_gateway/services/kernels/remotemanager.py
```diff
@@ -189,6 +192,57 @@ async def start_kernel(self, *args, **kwargs):
...
+            return
+        self.log.info("Going ahead to process kernel restart request.")
```
kevin-bates, line 205: Please remove.

enterprise_gateway/services/kernels/remotemanager.py
```diff
@@ -189,6 +192,57 @@ async def start_kernel(self, *args, **kwargs):
...
+        try:
+            kernel.restarting = True  # Moved in out of RemoteKernelManager
+            await super().restart_kernel(kernel_id)
+        finally:
+            self.log.debug("Resetting kernel.restarting flag to False.")
```
kevin-bates, line 210: I don't think this is very helpful, please remove.

enterprise_gateway/services/kernels/remotemanager.py
```diff
@@ -189,6 +192,57 @@ async def start_kernel(self, *args, **kwargs):
...
+            kernel.restarting = False
+
+    async def shutdown_kernel(self, kernel_id, now=False, restart=False):
+        kernel = self.get_kernel(kernel_id)
+        self.log.debug(f"Current value of the Kernel Restarting: {kernel.restarting}")
```
kevin-bates, line 215: Can be inferred.  Please remove.

enterprise_gateway/services/kernels/remotemanager.py
```diff
@@ -189,6 +192,57 @@ async def start_kernel(self, *args, **kwargs):
...
+        if kernel.restarting:
+            await self.wait_for_restart_finish(kernel_id, "shutdown")
+        try:
+            await super().shutdown_kernel(kernel_id, now, restart)
+        except KeyError as ke:  # this is hit for multiple shutdown request.
+            self.log.exception(f"Exception while shutting Kernel: {kernel_id}: {ke}")
```
kevin-bates, line 221:
```suggestion
            self.log.exception(f"Exception while shutting down kernel: '{kernel_id}': {ke}")
```
Shouldn't this be re-raised?  Or perhaps only if 'restarting' == False?

enterprise_gateway/services/kernels/remotemanager.py
```diff
@@ -189,6 +192,57 @@ async def start_kernel(self, *args, **kwargs):
...
+
+    async def wait_for_restart_finish(self, kernel_id, action="shutdown"):
+        kernel = self.get_kernel(kernel_id)
+        start_time = float(time.time())  # epoc time
+        timeout = kernel.kernel_launch_timeout
+        poll_time = 1  # TODO can be read from a new config if need be.
```
kevin-bates, line 227: Let's tie this to an env at a minimum.  How about `EG_RESTART_FINISH_POLL_INTERVAL`?  A default of 1.0 seems reasonable.

enterprise_gateway/services/kernels/remotemanager.py
```diff
@@ -189,6 +192,57 @@ async def start_kernel(self, *args, **kwargs):
...
+        self.log.info(
+            f"kernel is restarting when {action} request received. Polling every {poll_time} "
+            f"seconds for next {timeout} seconds for kernel '{kernel_id}'"
+            f" to complete its restart."
+        )
```
kevin-bates, line 232:
```suggestion
        self.log.info(
            f"Kernel '{kernel_id}' was restarting when {action} request received. Polling every {poll_time} "
            f"seconds for next {timeout} seconds for kernel to complete its restart."
        )
```

enterprise_gateway/services/kernels/remotemanager.py
```diff
@@ -189,6 +192,57 @@ async def start_kernel(self, *args, **kwargs):
...
+        while kernel.restarting:
+            now = float(time.time())
+            if (now - start_time) > timeout:
+                self.log.info(
+                    f"Wait_Timeout: Existing out of the restart wait loop to {action} kernel."
+                )
```
kevin-bates, line 238: Let's change this to debug...
```suggestion
                self.log.debug(
                    f"Timeout: Exiting restart wait loop in order to {action} kernel '{kernel_id}'."
                )
```

enterprise_gateway/services/kernels/remotemanager.py
```diff
@@ -189,6 +192,57 @@ async def start_kernel(self, *args, **kwargs):
...
+                break
+            await asyncio.sleep(poll_time)
+        self.log.info(
+            f"Returning with current value of the kernel.restarting: {kernel.restarting}."
+        )
```
kevin-bates, line 243:
```suggestion
        self.log.debug(
            f"Returning from restart-wait with kernel.restarting value: {kernel.restarting} for kernel '{kernel_id}'."
        )
```

enterprise_gateway/services/kernels/remotemanager.py
```diff
@@ -412,6 +467,9 @@ def _capture_user_overrides(self, **kwargs):
         of the kernelspec env stanza that would have otherwise overridden the user-provided values.
         """
         env = kwargs.get("env", {})
+        # see if KERNEL_LAUNCH_TIMEOUT was included from user.  If so, override default
+        if env.get("KERNEL_LAUNCH_TIMEOUT", None):
+            self.kernel_launch_timeout = float(env.get("KERNEL_LAUNCH_TIMEOUT"))
```
kevin-bates, line 472: We can scrap the if statement...
```suggestion
        self.kernel_launch_timeout = float(env.get("KERNEL_LAUNCH_TIMEOUT", default_kernel_launch_timeout))
```

enterprise_gateway/services/kernels/remotemanager.py
```diff
@@ -504,7 +562,10 @@ async def restart_kernel(self, now=False, **kwargs):
             Any options specified here will overwrite those used to launch the
             kernel.
         """
-        self.restarting = True
+        # this is added for only auto-restarter as it directly call this method.
+        self.log.info(f"restarting kernel with value for now: {now}")
+        if now:
+            self.restarting = True
```
kevin-bates, line 568: I don't think this logging is that helpful.  When possible, I'd rather have the comment specific to the action. Where it is makes it tough for someone reading the code know what "this is added" is referring to.  They might think its the larger block of code.
```suggestion
        if now:  # if auto-restarting (when now is True), indicate we're restarting.
            self.restarting = True
```

kevin-bates (commented):
Hi Rahul - thanks for doing this and the tremendous troubleshooting efforts this took - this is not easy to diagnose!

I had some relatively minor comments regarding the logging of information.  If you have found certain statements I asked to be removed to be essential, please feel free to push back and we can discuss.  Thanks.

enterprise_gateway/services/kernels/remotemanager.py
rahul26goyal: sure

enterprise_gateway/services/kernels/remotemanager.py
rahul26goyal: sure

enterprise_gateway/services/kernels/remotemanager.py
rahul26goyal: yes

enterprise_gateway/services/kernels/remotemanager.py
rahul26goyal: this usually happens when we have sent duplicate kernel shutdown request while the kernel was still restarting.
 I will testing this by raising the exception and get back with the behaviour.

enterprise_gateway/services/kernels/remotemanager.py
rahul26goyal: agree. I will  make this change.
- does EG_RESTART_STATUS_POLL_INTERVAL sounds better?
- will this also need a doc update for env variables?

enterprise_gateway/services/kernels/remotemanager.py
rahul26goyal: sounds good. ll update.

enterprise_gateway/services/kernels/remotemanager.py
rahul26goyal: this also can be inferred ..so removing this log line completely.

enterprise_gateway/services/kernels/remotemanager.py
rahul26goyal: L398, I have already initialised with the default value. So, I skipped setting this again. 
let me know if we still need to add this here?

enterprise_gateway/services/kernels/remotemanager.py
rahul26goyal: yes. agree kevin.

enterprise_gateway/services/kernels/remotemanager.py
rahul26goyal: I can change this to debug but I think this is an important log line that needs to be visible by default. Pls give this another thought and let me know .

enterprise_gateway/services/kernels/remotemanager.py
kevin-bates: That name sounds good - thanks.  Yeah, we should probably document this.  I think placing it into the [_Environment variables that assist in troubleshooting_](https://jupyter-enterprise-gateway.readthedocs.io/en/latest/operators/config-env-debug.html) section is a good spot.

enterprise_gateway/services/kernels/remotemanager.py
kevin-bates: I understand. It just seems cleaner and more maintainable to not have the extra `if` statement IMO.

enterprise_gateway/services/kernels/remotemanager.py
kevin-bates: Ok.  Leaving at INFO seems ok.  I'm not sure how useful it is other than to perhaps get an idea that a race condition occurred during restart/restart or restart/shutdown.  That said, if there is noise, this might be something for operators to look into.

enterprise_gateway/services/kernels/remotemanager.py
rahul26goyal: okay .. ll make the change

enterprise_gateway/services/kernels/remotemanager.py
rahul26goyal: yes @kevin-bates .. this will be helpful to debug any new issue that might arise due to this new code flow.

enterprise_gateway/services/kernels/remotemanager.py
rahul26goyal: Hi @kevin-bates : I tested various scenarios for this: 
1. The default behaviour of the JEG when multiple `shutdown` request is sent for the same kernel ID is except the first requests which returns 204 after shutting down the kernel, the other requests get en exception while `popping` out the `kernel_id ` from `MultiKernelManager._kernel` dict. This leads to `KeyError: 'b0b149e8-8a22-48db-b8b6-ed90e441ac41'` and the final response thrown to the clients is HTTP 500: `500 DELETE /api/kernels/b0b149e8-8a22-48db-b8b6-ed90e441ac41`.

2. With the current code change in place, we are handling the exception gracefully and returning a HTTP response 204. This behaviour is different from the point 1 above but this is similar to to the scenario where if you had sent a `shutdown` request for the `kernel_id` which does not exist. https://github.com/jupyter-server/jupyter_server/blob/main/jupyter_server/services/kernels/kernelmanager.py#L654  

let me know how to proceed forward with this. 
I feel in order to keep the behaviour same, we can skip the graceful handling and let it throw `500` to user as before.

enterprise_gateway/services/kernels/remotemanager.py
```diff
@@ -18,6 +19,9 @@
 from ..processproxies.processproxy import LocalProcessProxy, RemoteProcessProxy
 from ..sessions.kernelsessionmanager import KernelSessionManager
 
+default_kernel_launch_timeout = float(os.getenv("EG_KERNEL_LAUNCH_TIMEOUT", "30"))
+kernel_restart_finish_poll_internal = float(os.getenv("EG_RESTART_FINISH_POLL_INTERVAL", 1.0))
```
kevin-bates, line 23: typo
```suggestion
kernel_restart_finish_poll_interval = float(os.getenv("EG_RESTART_FINISH_POLL_INTERVAL", 1.0))
```

enterprise_gateway/services/kernels/remotemanager.py
kevin-bates: Could we please update the text of the log statement to the suggested value? It doesn't adhere to the conventions of the others (lower-case `kernel`, quoted `kernel_id`, and adds 'down' to complete the action).

enterprise_gateway/services/kernels/remotemanager.py
kevin-bates: Gentle ping regarding the suggested change.

kevin-bates (commented):
Getting closer.  One typo and a couple of suggestions.

enterprise_gateway/services/kernels/remotemanager.py
kevin-bates: Sorry, I just realized I didn't answer your question above.  

Hmm.  The timing issue you describe goes beyond the current checks made in the [`_check_kernel_id()` override](https://github.com/jupyter-server/jupyter_server/blob/main/jupyter_server/services/kernels/kernelmanager.py#L493-L497) in which we'll return a `404` if the kernel_id is unknown to the `MultiKernelManager` - correct?  (I had [fixed one such race condition a while okay related to the `_kernel_connections` list](https://github.com/jupyter/notebook/pull/5875).)

So I guess a similar issue still exists wrt to `_kernels` and the `try/except KeyError` block is a catch for that.  (You're using the latest server code with `AsyncKernelManagement` - correct?)

I like the fact that we'd no longer raise a 500 in this case, and most cases of this nature will result in a `404` due to the `_check_kernel_id()` override.  I think it could be misleading to return `204` when such requests did not delete the resource (kernel) and believe the best status, in this case, would be  `404` - since that's the truth and essentially exhibits similar behavior to when the kernel-id is not managed (which is also true).

rahul26goyal: yes @kevin-bates ..your observations are correct and I am testing with `AsyncKernelManagement`. 
The next step here is to raise a `404` similar to [check_kernel_id](https://github.com/jupyter-server/jupyter_server/blob/main/jupyter_server/services/kernels/kernelmanager.py#L493-L497).

```
raise web.HTTPError(404, "Kernel does not exist: %s" % kernel_id)
```

Also, should I further raise a CR in `AsyncMappingKernelManager::shutdown_kernel` to handle the  `KeyError` exception in jupyter server repo?

enterprise_gateway/services/kernels/remotemanager.py
rahul26goyal: sure. ll make this change as well.

enterprise_gateway/services/kernels/remotemanager.py
kevin-bates: Under normal circumstances, I'd say, yeah, if you can reproduce it in that environment.  But _that environment_ (outside of EG) will be using the pending kernel support - in which case I'd be surprised this can be reproduced. So, it comes down to, "is it worth a patch to jupyter_client 6.x?" and I'd say, let the status quo flow and live with the fact that `500` will be returned in this relatively rare scenario.  If we find issues here when converting to provisions, we can tackle this then, but this particular area of the stack is still evolving (e.g., _pending kernels_ will likely be subsumed by the _state machine_ work).

kevin-bates:
Great - thanks for the response.

PR #1056, opened 2022-03-28
Fix links

blink1073:
- Fix link check
- Break docs and check-manifest into separate jobs


docs/doc-requirements.txt
```diff
@@ -1,6 +1,8 @@
-jinja2
+jinja2<3.1
+mistune<1
```
kevin-bates, line 2: Could we please provide comments as to why these are capped and if/when we should revisit the cap's removal?  I suspect the `jinja2` cap is due to its removal of deprecated methods that (I want to say) nbconvert relies on.  Not sure what the `mistune` cap is about though.

blink1073: Done!

PR #1061, opened 2022-04-02
Adopt pre-commit

blink1073:


blink1073:
I'm not sure why the integration tests are timing out...

kevin-bates:
> I'm not sure why the integration tests are timing out...

I just tried running `make itest-yarn` locally (so I could tail the container logs) and found the timeout is occurring when it is downloading the Spark jars to HDFS.  I then hit the increased timeout (120), but noticed the download completed and EG started shortly after that.  Not sure why those downloads are taking longer right now, but could you go ahead and add another minute to the `PREP_TIMEOUT`?

.github/workflows/build.yml
```diff
@@ -1,9 +1,8 @@
 name: Builds
 on:
   push:
-    branches: '*'
+    branches: ['master']
```
kevin-bates, line 4: I believe we had `branches: '*'` so that when folks push work to their fork (prior to creating the pull request) they can ensure that the CI is passing prior to PR creation.  Assuming that's the way one accomplishes that particular behavior, what is the "standard workflow"?  Do folks need to first create a PR against the `master`/`main` branch of their fork in order to trigger CI, then, once confirmed, they point at the upstream branch?  I'm just curious and would like to do the right thing here.

Makefile
```diff
@@ -216,7 +216,7 @@ endif
 	($(SA) $(ENV) && GATEWAY_HOST=$(ITEST_DOCKER_HOST) LOG_LEVEL=$(LOG_LEVEL) KERNEL_USERNAME=$(ITEST_USER) KERNEL_LAUNCH_TIMEOUT=$(ITEST_KERNEL_LAUNCH_TIMEOUT) $(ITEST_DOCKER_KERNELS) ITEST_HOSTNAME_PREFIX=$(ITEST_USER) pytest -v -s $(TEST_DEBUG_OPTS) $(ITEST_DOCKER_TESTS))
 	@echo "Run \`docker service logs itest-docker\` to see enterprise-gateway log."
 
-PREP_TIMEOUT?=60
+PREP_TIMEOUT?=120
```
kevin-bates, line 219: Per my comment on the PR, let's go ahead and give 180 a shot here.  Should that fail, then perhaps 240, but I think if that too were to fail we'd want to get a better understanding of where the delay is actually occurring.
```suggestion
PREP_TIMEOUT?=180
```

kevin-bates (commented):
Hey Steve (@blink1073) - thanks for all the work you've been doing to bring EG in line with the rest of the server maintenance areas - it's greatly appreciated! 

These changes look good.  I didn't mark as approved simply because I'd like to get the CI passing.  I also had a general workflow question that I'm hoping you can answer.

Thanks again!

.github/workflows/build.yml
blink1073: Yeah fair, that works

blink1073:
Still timing out...

kevin-bates (approved):
LGTM - thanks @blink1073.  I'll need to look into the integration test timeout, but these changes should have no relation to that.

kevin-bates:
FWIW, I added a `docker logs` statement [during each poll for EG startup](https://github.com/kevin-bates/enterprise_gateway/runs/5821550018?check_suite_focus=true#step:12:288), and it looks like we've been bit by the recent `markupsafe` (2.1.1)  and `jinja2` (3.1.1) snafu...
```
Waiting for enterprise-gateway to start...
2022-04-04T18:16:08.380142789Z     from jinja2 import Environment, FileSystemLoader
2022-04-04T18:16:08.380146289Z   File "/opt/conda/lib/python3.7/site-packages/jinja2/__init__.py", line 12, in <module>
2022-04-04T18:16:08.380162289Z     from .environment import Environment
2022-04-04T18:16:08.380166389Z   File "/opt/conda/lib/python3.7/site-packages/jinja2/environment.py", line 25, in <module>
2022-04-04T18:16:08.380169989Z     from .defaults import BLOCK_END_STRING
2022-04-04T18:16:08.380300790Z   File "/opt/conda/lib/python3.7/site-packages/jinja2/defaults.py", line 3, in <module>
2022-04-04T18:16:08.380308190Z     from .filters import FILTERS as DEFAULT_FILTERS  # noqa: F401
2022-04-04T18:16:08.380311890Z   File "/opt/conda/lib/python3.7/site-packages/jinja2/filters.py", line 13, in <module>
2022-04-04T18:16:08.380315990Z     from markupsafe import soft_unicode
2022-04-04T18:16:08.380319590Z ImportError: cannot import name 'soft_unicode' from 'markupsafe' (/opt/conda/lib/python3.7/site-packages/markupsafe/__init__.py)
```
The base image used for this test has `jinja2==2.11.2` which isn't happy with the more recent `MarkupSafe`.  Since we require a floor of [`jinja2>=2.10`](https://github.com/kevin-bates/enterprise_gateway/blob/2307516d79d94d48eb9e004efe93831121e66b87/setup.py#L49), I think it might be best to raise it to `>=3.1`.  I believe 3.0.x may have had some issues here, so that's why I'm going with 3.1.  If this doesn't sound right we can iterate on the pending PR.

cc: @lresende

blink1073:
Yeah, 3.1 sounds good.  They're causing a whole lot of heartache by not backporting the fix to 2.x...

PR #1066, opened 2022-04-05
Add git-blame-ignore-revs

blink1073:


lresende:
Please let's make sure we have reviews and we are squashing to meaningful commits to avoid polluting the master history.

blink1073:
I don't understand this statement.   This was one commit that was a follow up to a previous PR that is procedural only.

lresende:
The eg community usually squash commits to meaningful commits/features/changes to avoid polluting the master history. In this particular case, I would prefer we had the "Add git-blame-ignore-revs" as one commit, instead of seeing "Add file" in the commit history as it is not very useful in the history. 

This could have been caught in a review, and usually, @kevin-bates is always on top of it, but I understand that sometimes we might need to provide a hot-fix and skip the review.

blink1073:
My mistake, I thought the default was squash-merge, as it is on most other PRs.  I'm disengaging from this conversation because it isn't productive.

Issue #795, opened 2020-03-25
a matlab kernel launcher is required


lazywhite:
I installed a [matlab jupyter kernel](https://github.com/Calysto/matlab_kernel)

It seems like different kernel need different launcher script, like this `etc/kernel-launchers/python/scripts/launch_ipykernel.py`

now I  need a kernel launcher  to run matlab kernel , How can I do it?

kevin-bates:
Yeah, although given the MatLab kernel is python, you could probably use most of it.  

What would be great is to plumb an option to specify a different kernel (e.g. `--kernel=matlab`).  Then, [`start_ipython()`](https://github.com/jupyter/enterprise_gateway/blob/master/etc/kernel-launchers/python/scripts/launch_ipykernel.py#L338) could be `start_kernel()` and we add a switch on the kernel to start, with `embed_kernel()` being the default.

In looking at MatLab, it (and other MetaKernels) appear to be strongly dependent on configuration files.  This could be a problem unless you only want to launch containerized versions since we'd need to figure out a way to get the configuration over to the node on which the kernel has landed.  With containerized versions, we could bake the configuration into the image or convey the various parameters via the [kernel-pod.yaml](https://github.com/jupyter/enterprise_gateway/blob/master/etc/kernel-launchers/kubernetes/scripts/kernel-pod.yaml.j2), etc.  What kind of environment are you targeting for this?

I'm happy to walk you through this and answer any questions, but I'm rather tied up with a few things at the moment.  Since we have yet to add support for a kernel outside of IPyKernel, IRKernel and Apache Toree (scala) this would be a great opportunity to "test those waters"!  Thank you for your interest.

lazywhite:
## environment
```
k8s: v1.13.12
jupyter-enterprise-gateway: 2.1.0
matlab: 2018b (it require python<3.7, I've build a py36 version of jupyter/base-image)
```
my goal is building an image like elyra/kernel-py:2.1.0 and a kernelspec for gateway to launch a pod with matlab kernel just like "python_kubernetes" etc...

lazywhite:
@kevin-bates   I tried as you said, here is [launch_ipykernel.py](https://gist.github.com/lazywhite/3d8a6730738f69b6019e7cfcb4bf929f)
I add an argument in line 384, now I can boot a notebook with this kernel

but  top right of notebook page show  kernel status, it's  always  a big black dot, not white annular.
please help me to figure out the wrong code.

[...]

Issue #1060, opened 2022-03-30
ipython startup scripts won't load


jaybubs:
Enterprise Gateway version: 2.6.0 (helm)

Ordinarily users put any startup scripts in their .ipython folder (eg /home/jovyan/.ipython/default_profile/startup) and they get executed during the kernel startup, however this is not the case when jupyter enterprise gateway(jeg) is used.

Some things I was able to find out:
When running a normal kernel `who` outputs `NamespaceMagics get_ipython`
When running jeg `who` outputs `Interactive namespace is empty`

Both in a normal and jeg kernel the output of `get_ipython().profile_dir.startup_dir` point correctly to `/home/jovyan/.ipython/default_profile/startup`

This got me digging some more until I backtraced it to `launch_ipykernel.py` where we have the following bits:
```python
def start_ipython(namespace, cluster_type="spark", **kwargs):
    from IPython import embed_kernel
    #other code processing namespace
    embed_kernel(local_ns=namespace, **kwargs)
```

Now the [embed_kernel function](https://github.com/ipython/ipython/blob/6d40fd892d5268a91a514d9dd532ba097dd345aa/IPython/__init__.py#L68) is then documented saying that:
```
    If you don't want the kernel to initialize the namespace
    from the scope of the surrounding function,
    and/or you want to load full IPython configuration,
    you probably want `IPython.start_kernel()` instead.
```
likewise, the complimentary [start_kernel function](https://github.com/ipython/ipython/blob/6d40fd892d5268a91a514d9dd532ba097dd345aa/IPython/__init__.py#L125) mentions:
```
    `IPython.embed_kernel()` puts a shell in a particular calling scope,
    such as a function or method for debugging purposes,
    which is often not desirable.
    `start_kernel()` does full, regular IPython initialization,
    including loading startup files, configuration, etc.
    much of which is skipped by `embed()`.
```

which then lead me to conclude the startup scripts are not loading precisely because it's using the "wrong" function. Digging even further into the code got me extremely confused because it's one hell of a rabbit hole.

I was keeping an eye out on what start_kernel does differently to embed_kernel, specifically looking for some sort of "glob_startup_folder" functionality but no luck thus far.

[...]

PR #1076, opened 2022-04-16
Support startup scripts and ipykernel subclasses

kevin-bates:
Issues #795 and #1060 both warranted the need to not use `embed_kernel()` when starting a python-based kernel.  `embed_kernel()` was the original approach used to launch the `ipykernel` from our launcher, since we wanted to control the SparkContext variables, but it turns out (at least in my initial testing) that that was not necessary.

This change uses the inspiration from the respective issues (and their authors @lazywhite and @jaybubs, respectively) to address both the startup script issue from #1060 as well as add support for ipykernel derivations like `MatlabKernel` in #795.

We also should try to support older python versions in the launcher, so I reverted the use of `f-strings` for now (as my YARN systems are using older versions of Python).

I installed a startup profile on my test environments with a simple `x = 42` statement and found that after the kernel started that variable and value were immediately available, for both Spark and non-Spark kernels.  @jaybubs, I'm hoping you can take this `launch_ipykernel.py` file for a spin in your environment to check things out.

Regarding support for ipykernel derivations, support for a `--kernel-class-name` option has been provided.  Its default value is `"ipykernel.ipkernel.IPythonKernel"` so that even the default case uses the same machinery.  @lazywhite, I'm hoping you can take  this `launch_ipykernel.py` file for a spin in your environment and, rather than set `--kernel=matlab`, instead use `--kernel-class-name=matlab_kernel.kernel.MatlabKernel`.  This approach avoids having to bake in a list of kernel name to class mappings, etc.

I found it was necessary to pass the `namespace` parameter into the `instance()` method, otherwise, https://github.com/jupyter-server/enterprise_gateway/issues/687 resurfaced.  So please check for this kind of thing using the following...
``` python
a = 123

def fun():
    print(a)
    
fun()
```

Resolves: #795, #1060

### DIFF
--- a/etc/kernel-launchers/python/scripts/launch_ipykernel.py
+++ b/etc/kernel-launchers/python/scripts/launch_ipykernel.py
@@ -14,7 +14,6 @@ from Cryptodome.Cipher import AES, PKCS1_v1_5
 from Cryptodome.PublicKey import RSA
 from Cryptodome.Random import get_random_bytes
 from Cryptodome.Util.Padding import pad
-from future.utils import raise_from
 from jupyter_client.connect import write_connection_file
 
 LAUNCHER_VERSION = 1  # Indicate to server the version of this launcher (payloads may vary)
@@ -149,14 +148,12 @@ class WaitingForSparkSessionToBeInitialized:
             self._init_thread.join(timeout=None)
             exc = self._init_thread.exc
             if exc:
-                raise_from(
-                    RuntimeError(
-                        "Variable: {} was not initialized properly.".format(
-                            self._spark_session_variable
-                        )
-                    ),
-                    exc,
-                )
+                raise RuntimeError(
+                    "Variable: {} was not initialized properly.".format(
+                        self._spark_session_variable
+                    )
+                ) from exc
+
             # now return attribute/function reference from actual Spark object
             return getattr(self._namespace[self._spark_session_variable], name)
 
--- a/etc/kernel-launchers/python/scripts/launch_ipykernel.py
+++ b/etc/kernel-launchers/python/scripts/launch_ipykernel.py
@@ -149,9 +149,7 @@ class WaitingForSparkSessionToBeInitialized:
             exc = self._init_thread.exc
             if exc:
                 raise RuntimeError(
-                    "Variable: {} was not initialized properly.".format(
-                        self._spark_session_variable
-                    )
+                    f"Variable: {self._spark_session_variable} was not initialized properly."
                 ) from exc
 
             # now return attribute/function reference from actual Spark object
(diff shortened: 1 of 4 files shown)


kevin-bates:
@lazywhite and @jaybubs, I'm hoping one or both of you can check this out.

kevin-bates:
@rahul26goyal - I'd appreciate your review as well.  Thank you.  

I was able to launch an instance of `Glue Pyspark` using this kernel.json:
```
{
  "display_name": "Glue Spark (Distributed)",
  "language": "python",
  "metadata": {
    "process_proxy": {
      "class_name": "enterprise_gateway.services.processproxies.distributed.DistributedProcessProxy"
    }
  },
  "argv": [
    "python",
    "/usr/local/share/jupyter/kernels/glue_pyspark/scripts/launch_ipykernel.py",
    "--RemoteProcessProxy.kernel-id",
    "{kernel_id}",
    "--RemoteProcessProxy.response-address",
    "{response_address}",
    "--RemoteProcessProxy.public-key",
    "{public_key}",
    "--RemoteProcessProxy.spark-context-initialization-mode",
    "none",
    "--kernel-class-name",
    "aws_glue_interactive_sessions_kernel.glue_pyspark.GlueKernel.GlueKernel"
  ]
}
```
I wasn't able to get past the successful launch since I didn't have a `Region` specified, but I suspect you'll know how to address that. :smile:

jaybubs:
@kevin-bates hello, apologies, the github notification was way too subtle!
So i've taken the new script and just replaced it in place of my modifications with the exception of [raise_from](https://github.com/jupyter-server/enterprise_gateway/blob/main/etc/kernel-launchers/python/scripts/launch_ipykernel.py#L17), which affects [this small chunk of code](https://github.com/jupyter-server/enterprise_gateway/blob/main/etc/kernel-launchers/python/scripts/launch_ipykernel.py#L150) and replaced it with
```py
if exc:
    raise RuntimeError("Variable: ........) from exc
```
which is in line with the other exceptions. The reason is because for some reason, our python 3.9 and 3.10 kernels fail to start with that line. Having said that, I make all our python kernels from scratch rather than the provided ones.

As far as the rest goes, it _seems_ to work properly now, startup scripts launch, cell magics work etc. I will have to wait maybe another week for my colleagues who use the kernels for actual work to report whether they notice any difference, and I'll let you know if they pick up anything. Otherwise it looks like the changes you made are good to go!

jaybubs (approved):

etc/kernel-launchers/python/scripts/launch_ipykernel.py
```diff
@@ -178,11 +180,15 @@ def _validate_port_range(port_range):
                 )
     except ValueError as ve:
         raise RuntimeError(
-            f"Port range validation failed for range: '{port_range}'.  Error was: {ve}"
+            "Port range validation failed for range: '{port_range}'.  Error was: {ve}".format(
```
lresende, line 183: Why are we moving back to format? instead of f-string?

jaybubs: Oh yea, I forgot to mention this too, any reason we're not using fstrings? They've been introduced in py 3.6 and previous versions of python are no longer supported. If you're doing this for backwards compatibility, you might be unintentionally creating maintenence hell for yourself!

kevin-bates: From the description:
> We also should try to support older python versions in the launcher, so I reverted the use of f-strings for now (as my YARN systems are using older versions of Python).

I felt that enterprises tend to (significantly) lag behind on support versions and not using `f-strings` in the launcher enables the use of Python < 3.5 (including 2.x, btw).  Since f-strings are purely for code readability, I didn't think it was that big a deal to not use them (once I figured out how to get the `pyupgrade` to ignore the launcher).

If others strongly believe we need `f-strings` over (significant) backward compatibility, I'm okay moving to `f-strings` in the python launcher.  I just felt this particular tradeoff was better served learning toward B/C.

kevin-bates:
Regarding the [`raise_from` comment](https://github.com/jupyter-server/enterprise_gateway/pull/1076#issuecomment-1107703719), addressing that to use `raise ... from` would likely force our hand at moving to `f-strings` since the purpose `raise_from` was added was to [retain support for Python 2.x](https://github.com/jupyter-server/enterprise_gateway/commit/7ab8e7bbb749f360d073ed33f45cd3150d9a3aab).

Since these changes are slated for a major-release boundary and with such a boundary being the time to make these kinds of changes, let's go ahead and move to `f-strings` as well and state Python 3.7+ support for EG - matching what our package states.  Folks deploying into older Python envs can "retro" their launcher scripts if necessary.

jaybubs (approved):

kevin-bates:
@lresende - are you okay moving this forward?  I think you're only comment was regarding the f-strings and I think we've cleared that up.

rahul26goyal (approved):
sorry for the late reply on this @kevin-bates. This is awesome!  
I did a quick test of the changes on the `spark python kubernetes` kernel and I did not notice any issue on my setup.  
👍  from my end.

lresende (approved):

Issue #1078, opened 2022-04-21
CI build and integration tests don't honor CI's python version


kevin-bates:
## Description
While working on another issue, I noticed that the GitHub actions to build EG and run its integration tests do not honor the version of Python in which the CI is meant to be run.   This is because the Makefile (and CI) doesn't specify the version of Python that it's running when creating its conda environment.  As a result, any CI steps that leverage the Makefile, which sets the conda environment, will use the default version of Python (currently 3.10).  This is true for the build of the wheel file and kernelspecs, as well as the integration tests.  Unit tests, on the other hand, are, for whatever reason, invoked from the GH action directly (which we should update to use the `Makefile`).

## Reproduce
It takes a little digging, but one can find a reference to the in-use Python on the [integration tests step](https://github.com/jupyter-server/enterprise_gateway/runs/6099992771?check_suite_focus=true#step:12:46) and see that it doesn't match the desired Python version.

## Expected behavior

We should be using the desired version of Python throughout the GitHub action.

It looks like GH sets the python version into the env `PYTHON_VERSION`, so we should be able to honor this value when creating the conda env.

kevin-bates:
Turns out that adding `python=$(PYTHON_VERSION)` to the `conda create` command is easier said than done.  I believe the `--file requirements.yml` parameter is interfering since it seems the python version is ignored only when that parameter is used.  When present, the python version within the conda env is always (and currently, since it's the default), `3.10.4`.

If I try to do this in two steps, installing from the file after the appropriately versioned env is created, I run into `conda init <shell>` required stuff.   This is actually a real PITA.  Yes, I'm sure we can add multiple `conda` commands and cobble something together, but...

We've talked (privately) about dropping the use of the conda env requirement from the Makefile since folks tend to develop in the appropriate env anyway, and I think this is a good time to propose such a change.  We already [bypass the conda env when running unit tests](https://github.com/jupyter-server/enterprise_gateway/blob/00d7376b932eacd347b3c32c863691bfbad53b86/.github/workflows/build.yml#L52-L58), and have to reset the [`SA` variable in the CI scripts](https://github.com/jupyter-server/enterprise_gateway/blob/00d7376b932eacd347b3c32c863691bfbad53b86/.github/workflows/build.yml#L44), so we're already leaking this pain elsewhere.

**As a result, the PR I produce to address this (serious) issue will be to remove the creation and activation of a conda env within the Makefile and GH workflow.**

[...]

PR #1079, opened 2022-04-21
Remove conda env config from build workflow

kevin-bates:
This pull request removes the use of a conda env from the EG build workflow (primarily in the Makefile).  This simplifies the working environment and appears to trim 3 minutes from each of the python-versioned CI jobs.

While doing this, I found that the former `make install` target was more like a test-install target, so I named it as such and split the two forms of installation it was performing into separate targets (so we can make use of one from our CI actions).

I also noticed that `make dev` was merely running whatever EG was installed, so I updated the target name to `run-dev` and had it depend on an actual build of EG to make it more inclusive.

The docs `Makefile` was also updated because the check for `sphinx-build` was interfering with the ability to install requirements using `make requirements`, which is part of the parent makefile's `make docs` target.

The other changes are mostly to remove the `$(SA) $(ENV)` references that activated the conda env and update the dev workflow docs to not depend on or reference `conda`.

Resolves: #1078

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,8 @@
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
 
-.PHONY: help build clean nuke dev dev-http docs install sdist test release clean-images clean-enterprise-gateway \
-    clean-demo-base clean-kernel-images clean-enterprise-gateway \
+.PHONY: help build clean nuke dev dev-http docs install bdist sdist test release check_dists \
+    clean-images clean-enterprise-gateway clean-demo-base clean-kernel-images clean-enterprise-gateway \
     clean-kernel-py clean-kernel-spark-py clean-kernel-r clean-kernel-spark-r clean-kernel-scala clean-kernel-tf-py \
     clean-kernel-tf-gpu-py clean-kernel-image-puller push-images push-enterprise-gateway-demo push-demo-base \
     push-kernel-images push-enterprise-gateway push-kernel-py push-kernel-spark-py push-kernel-r push-kernel-spark-r \
@@ -20,8 +20,11 @@ else
 endif
 
 
-WHEEL_FILE:=dist/jupyter_enterprise_gateway-$(VERSION)-py2.py3-none-any.whl
-WHEEL_FILES:=$(shell find . -type f ! -path "./build/*" ! -path "./etc/*" ! -path "./docs/*" ! -path "./.git/*" ! -path "./.idea/*" ! -path "./dist/*" ! -path "./.image-*" )
+#WHEEL_FILES:=$(shell find . -type f ! -path "./build/*" ! -path "./etc/*" ! -path "./docs/*" ! -path "./.git/*" ! -path "./.idea/*" ! -path "./dist/*" ! -path "./.image-*" )
+WHEEL_FILES:=$(shell find enterprise_gateway -type f )
+WHEEL_FILE:=dist/jupyter_enterprise_gateway-$(VERSION)-py3-none-any.whl
+SDIST_FILE:=dist/jupyter_enterprise_gateway-$(VERSION).tar.gz
+DIST_FILES=$(WHEEL_FILE) $(SDIST_FILE)
 
 HELM_CHART:=dist/jupyter_enterprise_gateway_helm-$(VERSION).tgz
 HELM_CHART_FILES:=$(shell find etc/kubernetes/helm -type f)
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -50,50 +40,43 @@ clean: ## Make a clean source tree
...
+	pip uninstall -y jupyter_enterprise_gateway
+	pip install dist/jupyter_enterprise_gateway-*.tar.gz && \
+		jupyter enterprisegateway --help
 
 bdist: lint
 	make $(WHEEL_FILE)
 
 $(WHEEL_FILE): $(WHEEL_FILES)
-	$(SA) $(ENV) && python setup.py bdist_wheel $(POST_SDIST) \
+	python setup.py bdist_wheel $(POST_SDIST) \
```
blink1073, line 75: Invoking `setup.py` directly is deprecated.  This would be `pip install build && python -m build --wheel .` (use `--sdist` below)

blink1073 (commented):
LGTM other than one suggestion

Makefile
kevin-bates: Thanks @blink1073!  My latest commit updates these to use `build` and `twine`.  I also fixed up the targets to utilize target dependencies correctly.

setup.py
```diff
@@ -29,6 +29,7 @@
 Jupyter Notebooks to share resources across distributed clusters such as
 Apache Spark, Kubernetes and others..
 """,
+    long_description_content_type="text/plain",
```
blink1073, line 32:
```suggestion
    long_description_content_type="text/markdown",
```

Makefile
blink1073: Thanks!

blink1073 (approved):
LGTM pending the one suggestion

setup.py
kevin-bates: `setup.py` currently uses [an _in-line_ long description](https://github.com/jupyter-server/enterprise_gateway/blob/00d7376b932eacd347b3c32c863691bfbad53b86/setup.py#L27-L30), so I used `text/plain` as the content type.  Seems like most PyPi projects point at their `README.md` file.  Is that what you had thought?  I'm good with using it as our long description - it winds up looking really clean - like server's looks: https://pypi.org/project/jupyter-server/

docs/source/contributors/devinstall.md
```diff
@@ -28,53 +28,28 @@ Enterprise Gateway's build environment is centered around `make` and the corresp
...
 clean-images                   Remove docker images (includes kernel-based images)
 clean-kernel-images            Remove kernel-based images
 clean                          Make a clean source tree
-dev                            Make a server in jupyter_websocket mode
 dist                           Make source, binary, kernelspecs and helm chart distributions to dist folder
 docker-images                  Build docker images (includes kernel-based images)
 docs                           Make HTML documentation
-env                            Make a dev environment
 helm-chart                     Make helm chart distribution
-install                        Make a conda env with dist/jupyter_enterprise_gateway-*.whl and dist/jupyter_enterprise_gateway-*.tar.gz installed
```
lresende, line 41: Are we removing `install`? Or just hiding it?

setup.py
blink1073: Ah, fair, thanks~

blink1073: I think it is fine as-is, as long as it points back to the repo on PyPI

Makefile
```diff
@@ -50,50 +43,44 @@ clean: ## Make a clean source tree
...
 
 docs: ## Make HTML documentation
-	$(SA) $(ENV) && make -C docs requirements html
+	make -C docs requirements html
 
 kernelspecs:  kernelspecs_all kernelspecs_yarn kernelspecs_conductor kernelspecs_kubernetes kernelspecs_docker kernel_image_files ## Create archives with sample kernelspecs
 kernelspecs_all kernelspecs_yarn kernelspecs_conductor kernelspecs_kubernetes kernelspecs_docker kernel_image_files:
 	make VERSION=$(VERSION) TAG=$(TAG) SPARK_VERSION=$(SPARK_VERSION) -C  etc $@
 
-install: ## Make a conda env with dist/jupyter_enterprise_gateway-*.whl and dist/jupyter_enterprise_gateway-*.tar.gz installed
```
lresende, line 72: Could we still leave the `install` at least as an undocumented task?

kevin-bates: Not sure why - it has no value other than to test installs.   From the PR description:
> While doing this, I found that the former make install target was more like a test-install target, so I named it as such and split the two forms of installation it was performing into separate targets (so we can make use of one from our CI actions).

I didn't realize this, but it was doing two installations, one via the wheel, the other via the tar file, and **uninstalling** both after each step.  Nothing more than a test target.

docs/source/contributors/devinstall.md
kevin-bates: Per the PR description, it has been renamed to `test-install`, since it does nothing more than that.

setup.py
kevin-bates: > as long as it points back to the repo on PyPI

Confirmed.

Makefile
lresende: I was under the impression that, once we run a  `make install` we build and install the package on the current env, no?

kevin-bates: > I was under the impression that, once we run a make install we build and install the package on the current env, no?

See the deleted code immediately below. It...

0. deletes the "install" conda env
1. creates the "install" conda env
2. activates the env
3. installs the wheel file
4. runs EG --help
5. uninstalls EG
6. removes the "install" conda env
then repeats 1-6 using the tar file.

That feels like an "install test" to me, so I renamed the target and broke each loop into separate targets, and eliminated the creation/removal of the "install" conda env.

I suppose we can add a `make install` target to invoke `pip install`, but the `make install` target you've been using doesn't wind up installing anything permanent.  I tend to copy the `whl` file to a cluster or build it into an image, so I rarely install EG where I build it.  But when I need to, I use use `pip install`.

> And thus, without that, I will need to run make build, and look for the wheel to install it?

The distributions are always in `dist` and the wheel is the only `whl` extension. You'd also need to deal with the kernelspec tars in that same directory.

PR #1083, opened 2022-05-03
Use flit build backend

blink1073:
Use a simpler, standards-compliant backend.


pyproject.toml
```diff
@@ -1,6 +1,71 @@
...
+  "Programming Language :: Python :: 3.8",
+  "Programming Language :: Python :: 3.9",
+  "Programming Language :: Python :: 3.10"
+]
+requires-python = ">=3.7"
+dependencies = [
+  "docker>=3.5.0",
+  "future",
+  "jinja2>=3.1",
+  "jupyter_client~=6.1",
```
kevin-bates, line 27: I'd like to start documenting "caps" when they are necessary (and since we're here...):
```suggestion
  "jupyter_client~=6.1",  # Remove cap once EG supports kernel provisioners
```

pyproject.toml
```diff
@@ -1,6 +1,71 @@
...
+  "jupyter_core>=4.6.0",
+  "kubernetes>=4.0.0",
+  "jupyter_server>=1.2",
+  "paramiko>=2.1.2",
+  "pexpect>=4.2.0",
+  "pycryptodomex>=3.9.7",
+  "pyzmq>=17.0.0",
+  "requests~=2.7",
```
kevin-bates, line 35: In this case, I think this "cap" can be removed since `requests` hasn't had a major version since 2013.  As with others, we can simply take our chances.
```suggestion
  "requests>=2.7",
```

pyproject.toml
```diff
@@ -1,6 +1,71 @@
...
+  "yarn-api-client>=1.0"
+]
+dynamic = ["version"]
+
+[[project.authors]]
+name = "Jupyter Development Team"
+email = "jupyter@googlegroups.com"
+
+[project.readme]
+text = "A lightweight, multi-tenant, scalable and secure gateway that enables\nJupyter Notebooks to share resources across distributed clusters such as\nApache Spark, Kubernetes and others.."
```
kevin-bates, line 48: Sorry, we should either add a dot or remove a dot here.  I'm thinking remove (also adding a comma):
```suggestion
text = "A lightweight, multi-tenant, scalable, and secure gateway that enables\nJupyter Notebooks to share resources across distributed clusters such as\nApache Spark, Kubernetes and others."
```

kevin-bates (commented):
Thanks @blink1073 - these changes look good.  I had some unrelated comments that essentially take advantage of this opportunity. If you feel these would be better served in a separate PR, I'm okay with that as well.

blink1073:
All green!

kevin-bates (approved):
LGTM - thanks Steve!

PR #1084, opened 2022-05-06
fix(k8s): handle status of deleted Pod correctly

dnwe:
Whilst deleting a kernel namespace returns a v1.Status, deleting a Pod
will return a v1.Pod and its status field is a v1.PodStatus, not a
string.

The existing code currently raises an exception as it attempts to do an
'in' check against the V1PodStatus type rather than a string as it
expected. This is logged as:
> Error occurred deleting pod: argument of type 'V1PodStatus' is not iterable


kevin-bates (approved):
These changes look great @dnwe - thank you for cleaning up this area.

PR #1093, opened 2022-05-19
Enable CodeQL analysis in CI

akchinSTC:
This workflow file will enable codeql checks for the master/main branch as well as any subsequent PRs. 
- Will only analyze python (.py) files to start. (May have support for other file types e.g. Dockerfile, scala, shell etc. added in future)
- Uses the default `security-and-quality` test suite, others can be added in the future. List of available queries [here](https://github.com/github/codeql/tree/main/python/ql/src/codeql-suites)


kevin-bates (approved):
LGTM - Thanks @akchinSTC - this is a good step forward for the project!

Issue #1086, opened 2022-05-10
Persisted sessions are not restored at startup, only if requested by kernel ID, leaking pods?


dnwe:
## Description

After changes were made under https://github.com/jupyter/enterprise_gateway/pull/737#issuecomment-567598751 persisted sessions were no longer loaded at startup, but only "on demand" if an `/api/kernels/<kernel-id>` came in that matched one of the persisted sessions.

In the scenario where a single enterprise-gateway replica is running in a Kube cluster and using KubernetesProcessProxy, this can lead to driver+executor Pods being leaked and remaining as idle processes that are never cleaned up.

## Reproduce

1. Deploy enterprise-gateway into a kube cluster and spinup some kernels
2. `kubectl delete --grace-period=0` the pod (or submit a `kill -9` manually) so that it is restarted without stopping any kernels
3. Observe time passing without querying those kernels
4. Note that the idle driver+executor pods are still present and seemingly lost.

## Expected behavior

```patch
--- a/enterprise_gateway/enterprisegatewayapp.py
+++ b/enterprise_gateway/enterprisegatewayapp.py
@@ -141,8 +141,7 @@ class EnterpriseGatewayApp(EnterpriseGat
         )
 
         # Attempt to start persisted sessions
-        # Commented as part of https://github.com/jupyter/enterprise_gateway/pull/737#issuecomment-567598751
-        # self.kernel_session_manager.start_sessions()
+        self.kernel_session_manager.start_sessions()
 
         self.contents_manager = None  # Gateways don't use contents manager
 ```

## Context

- Operating System and version: Kubernetes v1.21.12
- Browser and version: N/A
- Jupyter Server version: 2.6.0

kevin-bates:
Hi @dnwe - thanks for opening this issue.

I suppose this makes sense because, without the query, there's nothing that is going to "hydrate" a `KernelManager` to "connect" to outstanding kernels.  Of course, the same "leak" can happen if there are no further uses of orphaned kernels (i.e., the users have walked away), which, I would view as outside the scope here.  

I'm curious if you see this issue when multiple replicas are configured since that is the design center here (Active/Active)? If nothing is requesting a kernel's output following the termination of one of the EG nodes, then I'm not sure what the expected behavior should be.   How are your kernels being terminated if users just stop operating against them?  Are you relying completely on culling, and that's the difference here?  Are you using a different client than Jupyter Lab, since I'm fairly certain it will send period inquiries for the kernels it knows about?

[...]

PR #1095, opened 2022-05-25
Availability modes

kevin-bates:
#1086 references an issue whereby the loading of persisted kernel sessions at EG's startup was commented out when the changes for #737 were merged.  PR #737 essentially enabled the ability to, so to speak, have multiple instances of EG running simultaneously emulating an _active-active_ availability. The previous code, on the other hand, emulated more of an _active-passive_ behavior where only a single EG instance is running but introducing a higher degree of _resiliency_, as pointed out in #1086.  Some users have found that functionality helpful and we should try to accommodate that use case as well.

This pull request introduces a configurable option named `availability_mode` that can hold one of three values: `None` (default), `active-active`, and `active-passive`.  Both non-none values require that kernel session persistence also be enabled.  Since 'active-active' was essentially the default behavior (when kernel session persistence was enabled), we will automatically set the `availability_mode` to `active-active` whenever kernel session persistence is enabled and availability mode is not - thereby providing a form of backward compatibility.

Users desiring a single-instanced EG that is capable of restarting following an unexpected failure can now use the availability mode of 'active-passive'.

These modes (including kernel session persistence) can be enabled via a configuration file, command line, or environment variables as noted in the documentation or when running `jupyter enterprisegateway --help-all`.

As noted in the companion documentation, this functionality should be considered **experimental!**

Resolves: #1086


enterprise_gateway/enterprisegatewayapp.py
```diff
@@ -133,9 +134,26 @@ def init_configurables(self):
...
+            if self.kernel_session_manager.enable_persistence is False:
+                raise TraitError(
+                    f"Availability mode is configured as '{self.availability_mode}', "
+                    f"yet kernel session persistence has not been enabled.  Configure "
+                    f"KernelSessionManager.enable_persistence and restart Enterprise Gateway."
+                )
+
+            # If we're using active-passive availability, attempt to start persisted sessions
+            if self.availability_mode == "active-passive":
+                self.kernel_session_manager.start_sessions()
```
rahul26goyal, line 156: @kevin-bates : I went over the previous comments on this decision, but I am still  do not understand completely on "why we wouldn't want to  load all the sessions from persistence at server start" irrespective of the availability_mode?

docs/source/operators/config-availability.md
```diff
@@ -0,0 +1,106 @@
+# Availability modes
+
+Enterprise Gateway can be optionally configured in one of two "availability modes": _active-active_ and _active-passive_. Both modes require that kernel session persistence also be enabled via `KernelSessionManager.enable_persistence=True`.
+
+```{warning}
+**Availability modes and kernel session persistence should be considered experimental!**
+Known issues include:
+1. Culling configurations do not account for different nodes and therefore could result in the premature culling of kernels.
+2. Each "node switch" requires a manual reconnect to the kernel.
+
```
rahul26goyal, line 10: Are  the above issues only with "active-active" mode and not with "active-passive" mode of EG?

enterprise_gateway/mixins.py
```diff
@@ -680,6 +681,22 @@ def dynamic_config_interval_changed(self, event):
...
+        values=["active-active", "active-passive"],
+        config=True,
+        help="""Specifies the type of availability.  Values must be one of "active-active" or "active-passive".
+                Configuration of this this option requires that KernelSessionManager.enable_persistence is True.
+                (EG_AVAILABILITY_MODE env var)""",
+    )
+
+    @default("availability_mode")
+    def availability_mode_env_default(self):
+        return os.getenv(self.availability_mode_env)
```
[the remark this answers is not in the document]
kevin-bates: Good point.  Although the default value is `None` and `None` will be returned from `os.getenv()` if the env is not set, we should not assume the default will always be `None`.

enterprise_gateway/enterprisegatewayapp.py
kevin-bates: Yeah, I've had similar discussions with myself. :smile:  (It's good to have someone else to talk with about this!)
  
I think the primary issue here is **affinity** and if we load the sessions in _active-active_ then all EG nodes will have a `KernelManager` _thinking_ they are managing the kernel.  In _active-active_, a "second" node will only manage a "previously managed" kernel when the "previously-managing node" has gone down, so there's still only one node managing the kernel (because we always require "node affinity".

Perhaps the terms _active-active_ and _active-passive_ to describe these modes are not quite correct.  As @dnwe pointed out in the issue, they use _active-passive_ as more of a form of _resilency_ than HA (I suppose more for DR).  Perhaps we could spin _active-active_ to _HA_ and _active-passive_ to _DR_?

Thoughts?

docs/source/operators/config-availability.md
kevin-bates: I think reconnecting is necessary for both forms.  Even with "active-active", because we still expect/advise affinity with the managed kernel, you shouldn't run into an issue where the original EG is still running and can cull the kernel pre-maturely.  Only if the affinity is not configured could the kernel be culled prematurely from the previous node.  Since this is a general warning callout, I'm not sure we need to clarify, but we should probably take a closer look at this.

enterprise_gateway/enterprisegatewayapp.py
rahul26goyal: my thoughts on the naming the availability modes: 
active-passive  -> single_instance
"active-active" -> multi_instance

kevin-bates: I've updated the values to use the _instance_ references.  Note that I also added code to auto-enable kernel session persistence if not set when availability mode is set.  It felt a little overbearing to require the persistence setting when it's required to use "availability".  So, rather than throw an exception, we'll log an informational message.

rahul26goyal: @kevin-bates : I was going over some other service documentation where I found 2 new terms used to describe the similar availability scenarios:
1. active-passive -> `standalone`
2. active-active -> `replication`

enterprise_gateway/enterprisegatewayapp.py
```diff
@@ -133,9 +133,28 @@ def init_configurables(self):
...
+            # auto-enable persistence
+            if self.availability_mode is not None:
+                self.kernel_session_manager.enable_persistence = True
+                self.log.info(
+                    f"Availability mode is set to '{self.availability_mode}' yet kernel session "
+                    "persistence is not enabled.  Enabling kernel session persistence."
+                )
+
+        # If we're using single-instance availability, attempt to start persisted sessions
+        if self.availability_mode == "single-instance":
```
rahul26goyal, line 156: can we define  constants / static variables for these `availability_mode` so that it can be used across modules / files.

enterprise_gateway/services/kernels/remotemanager.py
```diff
@@ -162,7 +162,9 @@ def check_kernel_id(self, kernel_id):
                 self.parent.kernel_session_manager.delete_session(kernel_id)
                 raise web.HTTPError(404, "Kernel does not exist: %s" % kernel_id)
 
-    def _refresh_kernel(self, kernel_id):
+    def _refresh_kernel(self, kernel_id) -> bool:
+        if not self.parent.availability_mode or self.parent.availability_mode == "single-instance":
```
rahul26goyal, line 166: the thought here is, incase of `s_i` mode, the kernels are already hydrated when the EG server starts..so there is not need to check the persistence for kernel ?

enterprise_gateway/enterprisegatewayapp.py
kevin-bates: Hmm - I think I like these names over "single-instance" and "multi-instance", especially if there's precedent.

@lresende - you just approved this PR.  Are you okay with going with the names "Standalone" and "Replication"?

enterprise_gateway/services/kernels/remotemanager.py
kevin-bates: Correct. The multi-kernel manager should be aware of all active kernels in this case.

rahul26goyal (approved):
LGTM. Approving it.

kevin-bates:
I still need to apply the final name changes with "Standalone" and "Replication" so let's not merge yet.

kevin-bates:
Need to rework the docs now that #1101 has been merged.

PR #1097, opened 2022-05-26
Add support for Tenant ID

kevin-bates:
Although Enterprise Gateway states it supports multi-tenancy prior to this change, any "tenant" listing active kernels via `/api/kernels` will see kernels corresponding to every tenant.  As a result, the shutdown of one tenant would terminate the kernels of all tenants.

This pull request introduces `tenant_id` as a means for associating kernels to a given tenant and thereby adds support for multi-tenancy in a minimally viable way.

The [`GatewayClient` object that resides within Jupyter Server](https://github.com/jupyter-server/jupyter_server/blob/f25fd33ae068e55039905013f1552ff5ea051e37/jupyter_server/gateway/gateway_client.py#L13) will add support for the configuration of a tenant_id.  When configured, the `body` associated with a kernel's start request will include the tenant_id.  In addition, the _list kernels_ call will include a query parameter specifying the same tenant_id.

EG will use the tenant_id in the start request to manage a list of corresponding kernels.  When the client requests the list of active kernels, it will use the kernel ids associated with the given tenant-id to filter the results.

If no tenant-id is specified in the start request, the `UNIVERSAL_TENANT_ID` will be used.  This results in common functionality for both new and legacy applications.  Applications not configuring a tenant-id will result in the same behavior seen today (which is really only viable for "single-tenant" installations).

Although I haven't created a PR for the `GatewayClient` work (waiting for this merge), its changes can be found [here](https://github.com/jupyter-server/jupyter_server/compare/main...kevin-bates:add-gateway-tenant-support).


rahul26goyal:
hi @kevin-bates
Can multiple users share a single notebook and work on it simultaneously if connected to the same jupyterlab?

kevin-bates:
I'm not sure how real time collaboration (rtc) is implemented at that level but I suspect the "sharing" is at the notebook file level and each user has their own kernel. 
If they shared kernels then they'd need to share an EG and a tenant ID as well. Is that where your question is headed?

lresende:
We already have things like KERNEL_USERNAME, which provides a sort of a human-readable `tenant_id`. What is the drawbacks of using something like that instead of a UUID `tenant_id`?

kevin-bates:
> We already have things like KERNEL_USERNAME, which provides a sort of a human-readable tenant_id. What is the drawbacks of using something like that instead of a UUID tenant_id?

`KERNEL_USERNAME` is not unique enough - both when the jupyter server is multi-user, but also in order to distinguish between "alice" from two different tenants.

The UUID is not displayed anywhere, only (optionally) configured at the client (jupyter_server).  If we went with anything like a "name", then that opens the door for requiring a registry, etc. so that uniqueness can be detected.

enterprise_gateway/services/kernels/handlers.py
```diff
@@ -97,7 +104,10 @@ async def post(self):
             # so do a temporary partial (ugh)
             orig_start = self.kernel_manager.start_kernel
             self.kernel_manager.start_kernel = partial(
-                self.kernel_manager.start_kernel, env=env, kernel_headers=kernel_headers
+                self.kernel_manager.start_kernel,
+                env=env,
+                kernel_headers=kernel_headers,
+                tenant_id=tenant_id,
```
rahul26goyal, line 110: its already set in `env["KERNEL_TENANT_ID"]`. why do we pass again as kw args?

enterprise_gateway/services/kernels/handlers.py
```diff
@@ -45,35 +47,40 @@ async def post(self):
...
-                env_whitelist = model["env"].keys()
-            env.update(
-                {
-                    key: value
-                    for key, value in model["env"].items()
-                    if key.startswith("KERNEL_") or key in env_whitelist
-                }
-            )
+        if model is not None:
+            tenant_id = model.get("tenant_id", UNIVERSAL_TENANT_ID)
```
rahul26goyal, line 54: is `tenant_id` going to be an `env` variable which will be set on jupyterlab / notebook server?  If yes, may be this should be passed in the "env" part of the payload?

docs/source/contributors/system-architecture.md
```diff
@@ -491,3 +491,23 @@ Theoretically speaking, enabling a kernel for use in other frameworks amounts to
...
+
+Likewise, when calling the `/api/kernels` endpoint to get the list of active kernels, tenant-aware applications should add a `tenant_id` query parameter in order to get appropriate managed kernel information.
+
+```
+GET /api/kernels?tenant_id=f730794d-d175-40fa-b819-2a67d5308210
+```
+
+Kernel start or list requests that do not include a `tenant_id` will have their kernels associated with the `UNIVERSAL_TENANT_ID` which merely acts a catch-all and allows common code usage relative to existing clients.
+
+Enterprise Gateway will add the environment variable `KERNEL_TENANT_ID` to the kernel's environment so that this value is available to the kernel's launch logic and the kernel itself. It should be noted that if the original request also included a `KERNEL_TENANT_ID` in the body's `env` stanza, it will be overwritten with the value corresponding to `tenant_id` (or `UNIVERSAL_TENANT_ID` if `tenant_id` was not provided).
```
rahul26goyal, line 511: how are we currently passing "KERNEL_TENANT_ID" to the remote kernel?

enterprise_gateway/services/kernels/remotemanager.py
```diff
@@ -181,16 +238,19 @@ async def start_kernel(self, *args, **kwargs):
                 kernel_name=kwargs["kernel_name"], username=username
             )
         )
-
         # Check max kernel limits
         self._enforce_kernel_limits(username)
 
         RemoteMappingKernelManager.pending_requests.increment(username)
+        tenant_id = kwargs.pop(
```
rahul26goyal, line 245: can we read this from `kwargs['env']` ?

enterprise_gateway/services/sessions/kernelsessionmanager.py
```diff
@@ -93,6 +93,7 @@ def create_session(self, kernel_id, **kwargs):
         # Compose the kernel_session entry
         kernel_session = dict()
         kernel_session["kernel_id"] = kernel_id
+        kernel_session["tenant_id"] = self.kernel_manager.tenant_kernels.get_tenant_id(kernel_id)
```
rahul26goyal, line 96: can we read from `kwargs.get("env", {}).get("KERNEL_TENANT_ID")` ?

enterprise_gateway/services/kernels/remotemanager.py
```diff
@@ -286,8 +346,20 @@ def remove_kernel(self, kernel_id):
         super().remove_kernel(kernel_id)
         self.parent.kernel_session_manager.delete_session(kernel_id)
 
+    def list_kernels(self, tenant_id: Optional[str] = UNIVERSAL_TENANT_ID):
+        """Returns a list of kernel_id's of kernels running."""
+        kernels = []
+        kernel_ids = self.tenant_kernels.get_kernel_ids(tenant_id)
```
rahul26goyal, line 352: correct me, by doing this new way of `listing kernels`, we are introducing 2 different source of truth for running kernels list: 
1. Comes from the `MultiKernelManager::_kernels` which has been the single source of truth until now.
2. The new class ` TenantToKernels` which maintains the mapping with `tenant_id` -> `kernel_id` and vice versa. 

The challenge will be to always keep these 2 in sync.

Instead, we can iterate over each kernels and check if it matches the given `tenant_id`.

enterprise_gateway/services/kernels/remotemanager.py
```diff
@@ -286,8 +346,20 @@ def remove_kernel(self, kernel_id):
         super().remove_kernel(kernel_id)
         self.parent.kernel_session_manager.delete_session(kernel_id)
 
+    def list_kernels(self, tenant_id: Optional[str] = UNIVERSAL_TENANT_ID):
+        """Returns a list of kernel_id's of kernels running."""
```
rahul26goyal, line 350: incase, `tenant_id ==  UNIVERSAL_TENANT_ID`, should we not return all the kernels irrespective of `tenant_id`  the individual kernels are associated with.
Without this, 
how will I ever be able to list all kernels running on an EG?

enterprise_gateway/services/kernels/handlers.py
```diff
@@ -117,10 +127,16 @@ async def get(self):
         tornado.web.HTTPError
             403 Forbidden if kernel listing is disabled
         """
-        if not self.settings.get("eg_list_kernels"):
+
+        tenant_id_filter = self.request.query_arguments.get("tenant_id") or UNIVERSAL_TENANT_ID
+        if isinstance(tenant_id_filter, list):
+            tenant_id_filter = tenant_id_filter[0].decode("utf-8")
+        if not self.settings.get("eg_list_kernels") and tenant_id_filter == UNIVERSAL_TENANT_ID:
```
rahul26goyal, line 134: why do we have `and tenant_id_filter == UNIVERSAL_TENANT_ID:` check to this?

enterprise_gateway/services/kernels/remotemanager.py
```diff
@@ -19,6 +20,8 @@
 from ..processproxies.processproxy import LocalProcessProxy, RemoteProcessProxy
 from ..sessions.kernelsessionmanager import KernelSessionManager
 
+UNIVERSAL_TENANT_ID: str = "27182818-2845-9045-2353-602874713527"
```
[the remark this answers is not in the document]
kevin-bates: The UUID should be a static value because it will be persisted with the kernel session information.  So, when persistence is enabled, all EG servers should have the same UNIVERSAL_TENANT_ID for backwards compatibility.

enterprise_gateway/services/kernels/handlers.py
kevin-bates: It will be a configurable in `GatewayClient` which _can_ be set using `JUPYTER_GATEWAY_TENANT_ID` but that's a separate thing.  We can certainly require `JUPYTER_GATEWAY_TENANT_ID` reside in the `env` stanza - independent of how it's configured.  This will also avoid an API change (per se) in the start request.

So, if we say tenant_id is conveyed via the env portion of the body, we _could_ use `KERNEL_TENANT_ID` as the lookup key.  We'd still need to handle the case where no `env` is used (b/c) and set `KERNEL_TENANT_ID` to `UNIVERSAL_TENANT_ID` however.

I'm good making this change assuming you're okay with the above.

docs/source/contributors/system-architecture.md
kevin-bates: We're not right now.  It's only available to the kernel launch framework (process proxy and EG-side of things).  We'd need to formalize it as a parameter in each of the launchers if we really want/need it on the kernel side.

enterprise_gateway/services/kernels/remotemanager.py
kevin-bates: Yes.  Per above we'll use that as our form of conveyance.

enterprise_gateway/services/kernels/remotemanager.py
kevin-bates: > we can iterate over each kernels and check if it matches the given tenant_id.

Not sure what you mean here.  We do not store the tenant_id in the `KernelManager`. Only the `RemoteMappingKernelManager` knows about tenant_id.

enterprise_gateway/services/kernels/remotemanager.py
kevin-bates: Hmm - this is a good point.  You're referring to something like an admin app and not a request from a specific gateway client.  

Perhaps we have another UUID constant that only `list_kernels` checks and, when that tenant_id is present, it ignores tenant_id values and lists all kernels?

enterprise_gateway/services/kernels/handlers.py
kevin-bates: Today, EG requires that "list kernels" functionality be enabled explicitly - default it off.  This would merely allow those configurations that use their own tenant_ids to not have to set that configurable.  It's purely for convenience and I'd be okay not including this little "easter egg".

docs/source/contributors/system-architecture.md
kevin-bates: But, yes, I see I need to clarify this document.  It's only available to launch framework.

enterprise_gateway/services/kernels/remotemanager.py
kevin-bates: After a sidebar discussion, I'll push tenant_id onto `RemoteKernelManager` and we'll continue to leverage `self._kernels` as the single source of truth and discard the management of the mapping lists.

enterprise_gateway/services/kernels/remotemanager.py
kevin-bates: After a sidebar discussion, we will drop the defaulting of tenant_id to UNIVERSAL_TENANT_ID and display all kernels when nothing is specified - as is the behavior today.

enterprise_gateway/services/kernels/handlers.py
kevin-bates: Per discussion, the tenant_id will be conveyed in the `env` body via `JUPYTER_GATEWAY_TENANT_ID` and _converted_ to `KERNEL_TENANT_ID` _within_ EG.  Requests not including the tenant_id in the `env` will have `KERNEL_TENANT_ID` set to `UNIVERSAL_TENANT_ID`.

enterprise_gateway/services/sessions/kernelsessionmanager.py
kevin-bates: We're able to just use `km.tenant_id` since it will have that value set at this point.

kevin-bates:
@lresende, @rahul26goyal - are you available to review this PR?

kevin-bates:
I've moved this to draft.  After talking with @lresende, we felt it might be best to hold off on this until we have more firm requirements - particularly with how the tenant identities are configured and managed.

We will revisit down the road as necessary.

PR #1099, opened 2022-05-30
Add support to define terminationGracePeriodSeconds in helm chart

CiprianAnton:
The `EnterpriseGatewayApp` knows to respond to `SIGTERM` signal, but there is no support in helm chart to define `terminationGracePeriodSeconds`.

In case there are many pods owned by the EnterpriseGatewayApp, the default 30s terminationGracePeriodSeconds may not be enough to shutdown all the kernels.


etc/kubernetes/helm/enterprise-gateway/templates/deployment.yaml
```diff
@@ -31,6 +31,9 @@ spec:
     spec:
       # Created by this chart.
       serviceAccountName: {{ .Values.deployment.serviceAccountName }}
+  {{- if .Values.deployment.terminationGracePeriodSeconds }}
+      terminationGracePeriodSeconds: {{ .Values.deployment.terminationGracePeriodSeconds }}
+  {{- end }}
```
kevin-bates, line 36: If 30 seconds is the default, I think this might be more readable adding a `terminationGracePeriodSeconds` entry to the `values.yaml` file of `30` and updating this to:
```suggestion
      terminationGracePeriodSeconds: {{ .Values.deployment.terminationGracePeriodSeconds }}
```
That way operators can get a better idea of what knobs can be configured w/o looking through all the files.

kevin-bates (commented):
This is great - thank you for the contribution @CiprianAnton!

Besides the simplification comment, we should also update the appropriate locations within the [operators guide](https://github.com/jupyter-server/enterprise_gateway/blob/main/docs/source/operators/deploy-kubernetes.md) - namely the various `deployment` stanza examples (search for `replicas:`) as well as the [Helm Configuration Parameters table](https://github.com/jupyter-server/enterprise_gateway/blob/main/docs/source/operators/deploy-kubernetes.md#helm-configuration-parameters).

Thank you!

kevin-bates (approved):
LGTM - thank you @CiprianAnton!

PR #1103, opened 2022-06-04
Use conda clean --all instead of -tipsy

kevin-bates:
Looks like the `-s` option used in `conda clean` in various Dockerfile scripts has been deprecated and it's recommended that `conda clean --all` be used over `conda clean -tipsy`, so this pull request updates those locations.  We were also calling `conda clean` twice within the demo-base Dockerfile so the first of the two has been removed.

This is part of the release preparation process as this was causing `make docker-images` to fail during the execution of the release script.


akchinSTC (approved):
LGTM

PR #1105, opened 2022-06-07
[Feature] Support for configure magic on Spark Python Kubernetes Kernels (WIP)

rahul26goyal:
# Feature Description
The changes proposed in this PR are to add support for a well known magic `%%configure -f {}` which allows Notebook users to change the spark properties at runtime without having to create / update any kernel spec file.  This would allow users to change spark driver, executor resources (like cores, memory), enable / disable spark configuration etc. 

**Example**: The below snipped can be copied into a notebook cell to update the various spark properties associated with the current kernel.
```
%%configure -f 
{
  "driverMemory": "3G",
  "driverCores" : "2",
  "executorMemory" : "3G",
  "executorCores" : "2",
  "numExecutors" : 5,
  "conf" : {
      "spark.kubernetes.driver.label.test": "test-label"
  }
}
```

# Implementation Details
The below are the changes made at the high level:

1. I have introduced a new API on JEG `POST api/configure/<kernel_id>` which accepts a payload similar to create kernel API. This API currently support updating the `["KERNEL_EXTRA_SPARK_OPTS", "KERNEL_LAUNCH_TIMEOUT"]` env variables.
2. The above API tries to restart the same Kernel with the updated configuration. This is done because we want to keep the `kernel_id` same and want to give a smooth end user experience.
3. Once the old kernel goes away and a replacement comes up, we also need to refresh the ZMQ sockets to establish the connection with the new kernel so that existing active websocket connection from notebook / jupyterlab UI clients can continue to work. There hooks introduced to handle the same.
4. Further, in order to complete the usual Jupyter Kernel messaging handshake, we fire the missing  zmq messages from JEG to the websocket clients. Example: In order to mark the completion on the current cell, we need to send the `exec_reply` message and to mark the kernel idle, we need to kernel `status=idle` messages etc . These messages are pre-generated on the kernel and sent to JEG while making the API call to refresh the kernel. 

I will update more details about the changes and add some diagrams. 

# Testing
- Basic sanity testing done.  

# Note
Opening this PR for some early feedback and discussion on the changes.

### DIFF
--- a/enterprise_gateway/services/kernels/handlers.py
+++ b/enterprise_gateway/services/kernels/handlers.py
@@ -1,6 +1,7 @@
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
 """Tornado handlers for kernel CRUD and communication."""
+from datetime import datetime, timezone
 import json
 import os
 from functools import partial
@@ -8,7 +9,12 @@ from functools import partial
 import jupyter_server.services.kernels.handlers as jupyter_server_handlers
 import tornado
 from jupyter_client.jsonutil import date_default
+from jupyter_server.base.handlers import APIHandler
 from tornado import web
+try:
+    from jupyter_client.jsonutil import json_default
+except ImportError:
+    from jupyter_client.jsonutil import date_default as json_default
 
 from ...mixins import CORSMixin, JSONErrorsMixin, TokenAuthorizationMixin
 
@@ -146,11 +152,155 @@ class KernelHandler(
         self.finish(json.dumps(model, default=date_default))
 
 
-default_handlers = []
+class ConfigureMagicHandler(CORSMixin, JSONErrorsMixin, APIHandler):
+    @web.authenticated
+    async def post(self, kernel_id):
+        self.log.info(f"Update request received for kernel: {kernel_id}")
+        km = self.kernel_manager
+        km.check_kernel_id(kernel_id)
+        payload = self.get_json_body()
+        self.log.debug(f"Request payload: {payload}")
+        if payload is None:
+            self.log.info("Empty payload in the request body.")
+            self.finish(json.dumps({"message": "Empty payload received. No operation performed on the Kernel."},
+                                   default=date_default))
+            return
+        if type(payload) != dict:
+            self.log.info("Payload is not in acceptable format.")
+            raise web.HTTPError(400, u'Invalid JSON payload received.')
+        if payload.get("env", None) is None:  # We only allow env field for now.
+            self.log.info("Payload is missing the required env field.")
+            raise web.HTTPError(400, u'Missing required field `env` in payload.')
+        kernel = km.get_kernel(kernel_id)
+        if kernel.restarting:  # handle duplicate request.
+            self.log.info("An existing restart request is still in progress. Skipping this request.")
+            raise web.HTTPError(400, u'Duplicate Kernel update request received for Id: %s.' % kernel_id)
+        try:
+            # update Kernel metadata
+            kernel.set_user_extra_overrides(payload)
+            await km.restart_kernel(kernel_id)
+            kernel.fire_kernel_event_callbacks(event="kernel_refresh", zmq_messages=payload.get("zmq_messages", {}))
+        except web.HTTPError as he:
+            self.log.exception(f"HTTPError exception occurred in refreshing kernel: {kernel_id}: {he}")
+            await km.shutdown_kernel(kernel_id)
+            kernel.fire_kernel_event_callbacks(event="kernel_refresh_failure", zmq_messages=payload.get("zmq_messages", {}))
+            raise he
+        except Exception as e:
+            self.log.exception(f"An exception occurred in updating kernel : {kernel_id}: {e}")
+            await km.shutdown_kernel(kernel_id)
+            kernel.fire_kernel_event_callbacks(event="kernel_refresh_failure", zmq_messages=payload.get("zmq_messages", {}))
+            raise web.HTTPError(500, u'Error occurred while refreshing Kernel: %s.' % kernel_id, reason="{}".format(e))
+        else:
+            response_body = {
+                "message": f"Successfully refreshed kernel with id: {kernel_id}"
+            }
+            self.finish(json.dumps(response_body, default=date_default))
+            return
+
+
+class RemoteZMQChannelsHandler(TokenAuthorizationMixin,
+                               CORSMixin,
+                               JSONErrorsMixin,
+                               jupyter_server_handlers.ZMQChannelsHandler):
+
+    def open(self, kernel_id):
+        self.log.info(f"Websocket open request received for kernel: {kernel_id}")
+        super(RemoteZMQChannelsHandler, self).open(kernel_id)
+        km = self.kernel_manager
+        km.add_kernel_event_callbacks(kernel_id, self.on_kernel_refresh, "kernel_refresh")
+        km.add_kernel_event_callbacks(kernel_id, self.on_kernel_refresh_failure, "kernel_refresh_failure")
+
+    def on_kernel_refresh(self, **kwargs):
+        self.log.info("Refreshing the client websocket to kernel connection.")
+        self.refresh_zmq_sockets()
+        zmq_messages = kwargs.get("zmq_messages", {})
+        if "stream_reply" in zmq_messages:
+            self.log.debug("Sending stream_reply success message.")
+            success_message = zmq_messages.get("stream_reply")
+            success_message["content"] = {
+                "name": "stdout",
+                'text': "The Kernel is successfully refreshed."
+            }
+            self._send_ws_message(success_message)
+        if "exec_reply" in zmq_messages:
+            self.log.debug("Sending exec_reply message.")
+            self._send_ws_message(zmq_messages.get("exec_reply"))
+        if "idle_reply" in zmq_messages:
+            self.log.debug("Sending idle_reply message.")
+            self._send_ws_message(zmq_messages.get("idle_reply"))
+        self._send_status_message('kernel_refreshed')  # In the future, UI clients might start to consume this.
+
+    def on_kernel_refresh_failure(self, **kwargs):
+        self.log.error("Kernel %s refresh failed!", self.kernel_id)
+        zmq_messages = kwargs.get("zmq_messages", {})
+        if "error_reply" in zmq_messages:
+            self.log.debug("Sending stream_reply error message.")
+            error_message = zmq_messages.get("error_reply")
+            error_message["content"] = {
+                "ename": "KernelRefreshFailed",
+                'evalue': "The Kernel refresh operation failed.",
+                'traceback': ["The Kernel refresh operation failed."]
+            }
+            self._send_ws_message(error_message)
+        if "exec_reply" in zmq_messages:
+            self.log.debug("Sending exec_reply message.")
+            exec_reply = zmq_messages.get("exec_reply").copy()
+            if "metadata" in exec_reply:
+                exec_reply["metadata"]["status"] = "error"
+            exec_reply["content"]["status"] = "error"
+            exec_reply["content"]["ename"] = "KernelRefreshFailed."
+            exec_reply["content"]["evalue"] = "The Kernel refresh operation failed."
+            exec_reply["content"]["traceback"] = ["The Kernel refresh operation failed."]
+            self._send_ws_message(exec_reply)
+        if "idle_reply" in zmq_messages:
+            self.log.info("Sending idle reply message.")
+            self._send_ws_message(zmq_messages.get("idle_reply"))
+        self.log.debug("sending kernel dead message.")
+        self._send_status_message('dead')
+
+    def refresh_zmq_sockets(self):
+        self.close_existing_streams()
+        kernel = self.kernel_manager.get_kernel(self.kernel_id)
+        self.session.key = kernel.session.key  # refresh the session key
+        self.log.debug("Creating new ZMQ Socket streams.")
+        self.create_stream()
+        for channel, stream in self.channels.items():
+            self.log.debug(f"Updating channel: {channel}")
+            stream.on_recv_stream(self._on_zmq_reply)
+
+    def close_existing_streams(self):
+        self.log.debug(f"Closing existing channels for kernel: {self.kernel_id}")
+        for channel, stream in self.channels.items():
+            if stream is not None and not stream.closed():
+                self.log.debug(f"Close channel : {channel}")
+                stream.on_recv(None)
+                stream.close()
+        self.channels = {}
+
+    def _send_ws_message(self, kernel_msg):
+        self.log.debug(f"Sending websocket message: {kernel_msg}")
+        if "header" in kernel_msg and type(kernel_msg["header"] == dict):
+            kernel_msg["header"]["date"] = datetime.utcnow().replace(tzinfo=timezone.utc)
+        self.write_message(json.dumps(kernel_msg, default=json_default))
+
+    def on_close(self):
+        self.log.info(f"Websocket close request received for kernel: {self.kernel_id}")
+        super(RemoteZMQChannelsHandler, self).on_close()
+        self.kernel_manager.remove_kernel_event_callbacks(self.kernel_id, self.on_kernel_refresh, "kernel_refresh")
+        self.kernel_manager.remove_kernel_event_callbacks(self.kernel_id, self.on_kernel_refresh_failure,
+                                                          "kernel_refresh_failure")
+
+
+_kernel_id_regex = r"(?P<kernel_id>\w+-\w+-\w+-\w+-\w+)"
+default_handlers = [
+    (r"/api/configure/%s" % _kernel_id_regex, ConfigureMagicHandler)
+]
 for path, cls in jupyter_server_handlers.default_handlers:
     if cls.__name__ in globals():
         # Use the same named class from here if it exists
         default_handlers.append((path, globals()[cls.__name__]))
+    elif cls.__name__ == jupyter_server_handlers.ZMQChannelsHandler.__name__:  # TODO: check for override.
+        default_handlers.append((path, RemoteZMQChannelsHandler))
     else:
         # Gen a new type with CORS and token auth
         bases = (TokenAuthorizationMixin, CORSMixin, JSONErrorsMixin, cls)
--- a/enterprise_gateway/services/kernels/remotemanager.py
+++ b/enterprise_gateway/services/kernels/remotemanager.py
@@ -15,6 +15,8 @@ from traitlets import directional_link
 from traitlets import log as traitlets_log
 
 from enterprise_gateway.mixins import EnterpriseGatewayConfigMixin
+from jupyter_client.multikernelmanager import kernel_method
+from traitlets import Dict
 
 from ..processproxies.processproxy import LocalProcessProxy, RemoteProcessProxy
 from ..sessions.kernelsessionmanager import KernelSessionManager
@@ -369,6 +371,14 @@ class RemoteMappingKernelManager(AsyncMappingKernelManager):
 
         return new_kernel_id(kernel_id_fn=super().new_kernel_id, log=self.log, **kwargs)
 
+    @kernel_method
+    def add_kernel_event_callbacks(self, kernel_id, callback, event='kernel_refresh'):
+        """ Add kernel related events."""
+
+    @kernel_method
+    def remove_kernel_event_callbacks(self, kernel_id, callback, event='kernel_refresh'):
+        """ Remove event.."""
+
 
 class RemoteKernelManager(EnterpriseGatewayConfigMixin, AsyncIOLoopKernelManager):
     """
@@ -378,6 +388,11 @@ class RemoteKernelManager(EnterpriseGatewayConfigMixin, AsyncIOLoopKernelManager
     returned - upon which methods of poll(), wait(), send_signal(), and kill() can be called.
     """
 
+    event_callbacks = Dict()
+
+    def _event_callbacks_default(self):
+        return dict(kernel_refresh=[], kernel_refresh_failure=[])  # define new default when adding new event.
+
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
         self.process_proxy = None
@@ -385,7 +400,8 @@ class RemoteKernelManager(EnterpriseGatewayConfigMixin, AsyncIOLoopKernelManager
         self.public_key = None
         self.sigint_value = None
         self.kernel_id = None
-        self.user_overrides = {}
+        self.user_overrides = {}  # this is populated via create kernel request.
+        self.user_update_overrides = {}  # this is populated via update kernel request.
         self.kernel_launch_timeout = default_kernel_launch_timeout
         self.restarting = False  # need to track whether we're in a restart situation or not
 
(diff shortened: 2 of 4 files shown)


enterprise_gateway/services/kernels/handlers.py
```diff
@@ -146,11 +153,178 @@ def get(self, kernel_id):
...
+        if payload.get("env", None) is None:  # We only allow env field for now.
+            self.log.info("Payload is missing the required env field.")
+            raise web.HTTPError(400, "Missing required field `env` in payload.")
+        kernel = km.get_kernel(kernel_id)
+        if kernel.restarting:  # handle duplicate request.
+            self.log.info(
+                "An existing restart request is still in progress. Skipping this request."
+            )
+            raise web.HTTPError(
+                400, "Duplicate Kernel update request received for Id: %s." % kernel_id
```
kevin-bates, line 185:
```suggestion
                400, "Duplicate configure kernel request received for kernel: %s." % kernel_id
```

kevin-bates, line 185:
```suggestion
                400, "Duplicate configure kernel request received for kernel: %s." % kernel_id
```

enterprise_gateway/services/kernels/handlers.py
```diff
@@ -146,11 +153,178 @@ def get(self, kernel_id):
...
+            raise he
+        except Exception as e:
+            self.log.exception(f"An exception occurred in updating kernel : {kernel_id}: {e}")
+            await km.shutdown_kernel(kernel_id)
+            kernel.fire_kernel_event_callbacks(
+                event="kernel_refresh_failure", zmq_messages=payload.get("zmq_messages", {})
+            )
+            raise web.HTTPError(
+                500,
+                "Error occurred while refreshing Kernel: %s." % kernel_id,
```
kevin-bates, line 211:
```suggestion
                "Error occurred while refreshing kernel: %s." % kernel_id,
```

kevin-bates, line 211:
```suggestion
                "Error occurred while refreshing kernel: %s." % kernel_id,
```

enterprise_gateway/services/kernels/handlers.py
```diff
@@ -146,11 +153,178 @@ def get(self, kernel_id):
...
+            self.log.exception(
+                f"HTTPError exception occurred in refreshing kernel: {kernel_id}: {he}"
+            )
+            await km.shutdown_kernel(kernel_id)
+            kernel.fire_kernel_event_callbacks(
+                event="kernel_refresh_failure", zmq_messages=payload.get("zmq_messages", {})
+            )
+            raise he
+        except Exception as e:
+            self.log.exception(f"An exception occurred in updating kernel : {kernel_id}: {e}")
```
kevin-bates, line 204:
```suggestion
            self.log.exception(f"An exception occurred in re-configuring kernel: {kernel_id}: {e}")
```

kevin-bates, line 204:
```suggestion
            self.log.exception(f"An exception occurred in re-configuring kernel: {kernel_id}: {e}")
```

enterprise_gateway/services/kernels/handlers.py
```diff
@@ -146,11 +153,178 @@ def get(self, kernel_id):
...
+            self.finish(
+                json.dumps(
+                    {"message": "Empty payload received. No operation performed on the Kernel."},
+                    default=date_default,
+                )
+            )
+            return
+        if type(payload) != dict:
+            self.log.info("Payload is not in acceptable format.")
+            raise web.HTTPError(400, "Invalid JSON payload received.")
```
kevin-bates, line 175: Same goes for these `info` messages that precede exceptions.  I believe EG will be logging these exceptions.

Could you please append something like `f" for kernel: {kernel_id}"` to the end?  As EG becomes more multi-tenant, its important we try to include some unique identifier in messages wherever possible.

kevin-bates, line 175: Same goes for these `info` messages that precede exceptions.  I believe EG will be logging these exceptions.

Could you please append something like `f" for kernel: {kernel_id}"` to the end?  As EG becomes more multi-tenant, its important we try to include some unique identifier in messages wherever possible.

enterprise_gateway/services/kernels/handlers.py
```diff
@@ -146,11 +153,178 @@ def get(self, kernel_id):
...
+            response_body = {"message": f"Successfully refreshed kernel with id: {kernel_id}"}
+            self.finish(json.dumps(response_body, default=date_default))
+            return
+
+
+class RemoteZMQChannelsHandler(
+    TokenAuthorizationMixin, CORSMixin, JSONErrorsMixin, jupyter_server_handlers.ZMQChannelsHandler
+):
+    def open(self, kernel_id):
+        self.log.info(f"Websocket open request received for kernel: {kernel_id}")
```
kevin-bates, line 224: info -> debug

kevin-bates, line 224: info -> debug

enterprise_gateway/services/kernels/handlers.py
```diff
@@ -146,11 +153,178 @@ def get(self, kernel_id):
...
+    def on_kernel_refresh(self, **kwargs):
+        self.log.info("Refreshing the client websocket to kernel connection.")
+        self.refresh_zmq_sockets()
+        zmq_messages = kwargs.get("zmq_messages", {})
+        if "stream_reply" in zmq_messages:
+            self.log.debug("Sending stream_reply success message.")
+            success_message = zmq_messages.get("stream_reply")
+            success_message["content"] = {
+                "name": "stdout",
+                "text": "The Kernel is successfully refreshed.",
```
kevin-bates, line 241: This is throughout.  Please lowercase `kernel` unless part of a name (e.g., `KernelManager`).

kevin-bates, line 241: This is throughout.  Please lowercase `kernel` unless part of a name (e.g., `KernelManager`).

enterprise_gateway/services/kernels/handlers.py
```diff
@@ -146,11 +153,178 @@ def get(self, kernel_id):
...
+        self.kernel_manager.remove_kernel_event_callbacks(
+            self.kernel_id, self.on_kernel_refresh, "kernel_refresh"
+        )
+        self.kernel_manager.remove_kernel_event_callbacks(
+            self.kernel_id, self.on_kernel_refresh_failure, "kernel_refresh_failure"
+        )
+
+
+_kernel_id_regex = r"(?P<kernel_id>\w+-\w+-\w+-\w+-\w+)"
+default_handlers = [(r"/api/configure/%s" % _kernel_id_regex, ConfigureMagicHandler)]
```
kevin-bates, line 319: This feels like it should be `r"/api/kernels/configure/%s"`.  Is there an issue with placing it there?  I don't see that conflicting with the patterns of the default handlers.

kevin-bates, line 319: This feels like it should be `r"/api/kernels/configure/%s"`.  Is there an issue with placing it there?  I don't see that conflicting with the patterns of the default handlers.

enterprise_gateway/services/kernels/handlers.py
```diff
@@ -146,11 +153,178 @@ def get(self, kernel_id):
...
 for path, cls in jupyter_server_handlers.default_handlers:
     if cls.__name__ in globals():
         # Use the same named class from here if it exists
         default_handlers.append((path, globals()[cls.__name__]))
+    elif (
+        cls.__name__ == jupyter_server_handlers.ZMQChannelsHandler.__name__
+    ):  # TODO: check for override.
+        default_handlers.append((path, RemoteZMQChannelsHandler))
```
kevin-bates, line 327: Isn't this meant to _replace_ `ZMQChannelsHandler`?  I guess I don't understand why `ZMQChannelsHandler` isn't satisfied by the first condition - but I'm not that familiar with `globals()` (sorry).

kevin-bates, line 327: Isn't this meant to _replace_ `ZMQChannelsHandler`?  I guess I don't understand why `ZMQChannelsHandler` isn't satisfied by the first condition - but I'm not that familiar with `globals()` (sorry).

enterprise_gateway/services/kernels/remotemanager.py
```diff
@@ -378,14 +387,22 @@ class RemoteKernelManager(EnterpriseGatewayConfigMixin, AsyncIOLoopKernelManager
...
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
         self.process_proxy = None
         self.response_address = None
         self.public_key = None
         self.sigint_value = None
         self.kernel_id = None
-        self.user_overrides = {}
+        self.user_overrides = {}  # this is populated via create kernel request.
+        self.user_update_overrides = {}  # this is populated via update kernel request.
```
kevin-bates, line 405: Was this feature originally named 'update' or something?  I think `configure_kernel_overrides` or `configure_magic_overrides` is more indicative of where these are coming from.

kevin-bates, line 405: Was this feature originally named 'update' or something?  I think `configure_kernel_overrides` or `configure_magic_overrides` is more indicative of where these are coming from.

enterprise_gateway/services/kernels/remotemanager.py
```diff
@@ -462,15 +479,24 @@ def _capture_user_overrides(self, **kwargs):
...
+        if (
+            not self.user_overrides
+        ):  # user_overrides preserve the original envs with which the kernel is started.
+            self.user_overrides.update(
+                {
+                    key: value
+                    for key, value in env.items()
+                    if key.startswith("KERNEL_")
+                    or key in self.parent.parent.env_process_whitelist
+                    or key in self.parent.parent.env_whitelist
```
kevin-bates, line 492: These should revert to not include `parent.parent` as these are "massaged" in `__init__()` via `self._link_dependent_props()`:
```suggestion
                    or key in self.env_process_whitelist
                    or key in self.env_whitelist
```

kevin-bates, line 492: These should revert to not include `parent.parent` as these are "massaged" in `__init__()` via `self._link_dependent_props()`:
```suggestion
                    or key in self.env_process_whitelist
                    or key in self.env_whitelist
```

enterprise_gateway/services/kernels/remotemanager.py
```diff
@@ -507,7 +533,6 @@ async def _launch_kernel(self, kernel_cmd, **kwargs):
 
         # Apply user_overrides to enable defaulting behavior from kernelspec.env stanza.  Note that we do this
         # BEFORE setting KERNEL_GATEWAY and removing {EG,KG}_AUTH_TOKEN so those operations cannot be overridden.
```
kevin-bates, line 535: This comment nolonger makes sense with the update statement removed.  Perhaps moving this following the setting of `KERNEL_GATEWAY`?
```suggestion
        # Remove {EG,KG}_AUTH_TOKEN so these envs cannot be overridden.
```

kevin-bates, line 535: This comment nolonger makes sense with the update statement removed.  Perhaps moving this following the setting of `KERNEL_GATEWAY`?
```suggestion
        # Remove {EG,KG}_AUTH_TOKEN so these envs cannot be overridden.
```

enterprise_gateway/services/kernels/remotemanager.py
```diff
@@ -521,7 +546,11 @@ async def _launch_kernel(self, kernel_cmd, **kwargs):
                 self.kernel_spec.display_name, kernel_cmd
             )
         )
-
+        self.log.debug(
+            "Launching kernel: {} with command: {}".format(
+                self.kernel_spec.display_name, kernel_cmd
+            )
+        )
```
kevin-bates, line 553: This is a duplicate of the lines immediately before it.

kevin-bates, line 553: This is a duplicate of the lines immediately before it.

enterprise_gateway/services/kernels/remotemanager.py
```diff
@@ -725,3 +754,100 @@ def mapping_kernel_manager(self):
...
+    def add_kernel_event_callbacks(self, callback_func, event="kernel_refresh"):
+        """register a callback to fire on a particular event
+
+        :param callback_func:
+        :param event:
+            - 'kernel_refresh' (default): kernel has received an update request and has successfully restarted.
+        :return:
+        """
+        try:
+            self.log.info(
```
kevin-bates, line 805: info -> debug

kevin-bates, line 805: info -> debug

enterprise_gateway/services/kernels/remotemanager.py
```diff
@@ -725,3 +754,100 @@ def mapping_kernel_manager(self):
...
+
+    def remove_kernel_event_callbacks(self, callback_func, event="kernel_refresh"):
+        """Deregister a callback from this kernel event.
+
+        :param callback_func: the callback to be removed if exists.
+        :param event:  'kernel_refresh'
+        :return: nothing.
+        """
+
+        self.log.info(
```
kevin-bates, line 825: info -> debug

kevin-bates, line 825: info -> debug

enterprise_gateway/services/kernels/remotemanager.py
```diff
@@ -725,3 +754,100 @@ def mapping_kernel_manager(self):
...
+                "Failed to remove callback for event: {}: callback: {}".format(
+                    event, callback_func.__name__
+                ),
+                exc_info=True,
+            )
+
+    def fire_kernel_event_callbacks(self, **kwargs):
+        """fire the callbacks for a particular kernel event"""
+        event = kwargs.get("event")
+        self.log.info(f"fire_kernel_event_callbacks: called for event: {event}")
```
kevin-bates, line 841: info -> debug

kevin-bates, line 841: info -> debug

enterprise_gateway/services/kernels/remotemanager.py
```diff
@@ -725,3 +754,100 @@ def mapping_kernel_manager(self):
...
+        for callback in self.event_callbacks[event]:
+            try:
+                self.log.info(f"triggering callback to {callback.__name__}")
```
kevin-bates, line 844: info -> debug

kevin-bates, line 844: info -> debug

enterprise_gateway/services/processproxies/container.py
```diff
@@ -175,6 +175,7 @@ def kill(self):
         result = None
 
         if self.container_name:  # We only have something to terminate if we have a name
+            self.log.info(f"Killing Kernel with name: {self.container_name}")
```
kevin-bates, line 178: info -> debug
```suggestion
            self.log.debug(f"Terminating kernel with container name: {self.container_name}")
```
or replace with kernel_id...
```
self.log.debug(f"Terminating kernel: {self.kernel_id}")
```
although using the container name here is probably more appropriate and will contain the kernel_id.

kevin-bates, line 178: info -> debug
```suggestion
            self.log.debug(f"Terminating kernel with container name: {self.container_name}")
```
or replace with kernel_id...
```
self.log.debug(f"Terminating kernel: {self.kernel_id}")
```
although using the container name here is probably more appropriate and will contain the kernel_id.

etc/kernel-launchers/python/scripts/configure_magic.py
```diff
@@ -0,0 +1,296 @@
...
+logger.setLevel(logging.DEBUG)
+logger.propagate = True
+
+RESERVED_SPARK_CONFIGS = [
+    "spark.kubernetes.container.image",
+    "spark.kubernetes.driver.container.image",
+    "spark.kubernetes.executor.container.image",
+    "spark.kubernetes.namespace",
+    "spark.kubernetes.driver.label.component",
+    "spark.kubernetes.executor.label.component" "spark.kubernetes.driver.label.kernel_id",
```
kevin-bates, line 26:
```suggestion
    "spark.kubernetes.executor.label.component", "spark.kubernetes.driver.label.kernel_id",
```
```suggestion
    "spark.kubernetes.executor.label.component" "spark.kubernetes.driver.label.kernel_id",
```

kevin-bates, line 26:
```suggestion
    "spark.kubernetes.executor.label.component", "spark.kubernetes.driver.label.kernel_id",
```
```suggestion
    "spark.kubernetes.executor.label.component" "spark.kubernetes.driver.label.kernel_id",
```

etc/kernel-launchers/python/scripts/configure_magic.py
```diff
@@ -0,0 +1,296 @@
...
+        MAX_LAUNCH_TIMEOUT
+    )
+    INVALID_PAYLOAD_ERROR = "{} with error: {}"
+
+    def __init__(self, shell=None, **kwargs):
+        logger.info("New Initializing ConfigureMagic...")
+        super().__init__(shell=None, **kwargs)
+        self.shell = shell
+        self.kernel_id = os.environ.get("KERNEL_ID", None)
+        self.endpoint_ip = os.environ.get("ENDPOINT_IP", "").split(":")[0]
```
kevin-bates, line 73: What sets the `ENDPOINT_ID` env?

kevin-bates, line 73: What sets the `ENDPOINT_ID` env?

etc/kernel-launchers/python/scripts/configure_magic.py
```diff
@@ -0,0 +1,296 @@
...
+        self.endpoint_port = 9547
+        self.protocol = "https"
```
kevin-bates, line 75: Should these also be configurable?

kevin-bates, line 75: Should these also be configurable?

etc/kernel-launchers/python/scripts/configure_magic.py
```diff
@@ -0,0 +1,296 @@
...
+            # Flush output before sending the request.
+            sys.stdout.flush()
+            sys.stderr.flush()
+            time.sleep(0.005)  # small delay
+            response = requests.post(
+                self.update_kernel_url,
+                data=json.dumps(payload_dict, default=str),
+                headers=self.headers,
+                verify=False,
+            )
```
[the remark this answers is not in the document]
kevin-bates: We should address this alert somehow.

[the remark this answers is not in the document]
kevin-bates: We should address this alert somehow.

enterprise_gateway/services/kernels/remotemanager.py
```diff
@@ -725,3 +754,100 @@ def mapping_kernel_manager(self):
...
+            if override_key in user_requested_env_overrides:
+                self.log.info("Key exist in extra overrides..")
+                if override_key == "KERNEL_LAUNCH_TIMEOUT":
+                    allowed_env_overrides[override_key] = str(
+                        user_requested_env_overrides.get(override_key)
+                    )
+                else:
+                    allowed_env_overrides[override_key] = self.user_overrides.get(
+                        override_key, ""
+                    ) + user_requested_env_overrides.get(override_key)
```
kevin-bates, line 775: Is this block assuming it's only for `KERNEL_EXTRA_SPARK_OPTS`?  it appears to be _additive_ rather than a _replacement_ (like the launch timeout).  WIll that be problematic or does Spark take the last duplicated value it finds?

kevin-bates, line 775: Is this block assuming it's only for `KERNEL_EXTRA_SPARK_OPTS`?  it appears to be _additive_ rather than a _replacement_ (like the launch timeout).  WIll that be problematic or does Spark take the last duplicated value it finds?

enterprise_gateway/services/kernels/remotemanager.py
```diff
@@ -725,3 +754,100 @@ def mapping_kernel_manager(self):
...
+        return allowed_env_overrides
+
+    def set_user_extra_overrides(self, update_payload):
+        # TODO need to read this list from env variable
+        whitelisted_override_keys = ["KERNEL_EXTRA_SPARK_OPTS", "KERNEL_LAUNCH_TIMEOUT"]
```
kevin-bates, line 780: We're trying to move away from certain socially-sensitive terms, so I'd rather we not introduce other instances.  Perhaps something like:
```suggestion
        allowed_override_keys = ["KERNEL_EXTRA_SPARK_OPTS", "KERNEL_LAUNCH_TIMEOUT"]
```

kevin-bates, line 780: We're trying to move away from certain socially-sensitive terms, so I'd rather we not introduce other instances.  Perhaps something like:
```suggestion
        allowed_override_keys = ["KERNEL_EXTRA_SPARK_OPTS", "KERNEL_LAUNCH_TIMEOUT"]
```

kevin-bates (commented):
This is really interesting @rahul26goyal - thank you.  I'm not sure we should hold the 3.0 release for this as this seems additive.  Most of the comments are little things.

One thing I'm not sure about is that this only applies to Python kernels, and only those configured for Spark (at the moment, although it could be more general).

I'd also like to look into incorporating the `RemoteZMQChannelsHandler` in a general way.  I haven't given it a real close look, but do you think that could be useful even outside this particular "configure/restart" feature?

etc/kernel-launchers/python/scripts/configure_magic.py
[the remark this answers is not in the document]
rahul26goyal: sure...we can discuss the approach to resolve this.

rahul26goyal:
@kevin-bates : 
I am help is deciding the right terminology for the operation we are performing on the kernel using this new `configure`   API:
1. do we call it " refreshing kernel" or "re-configuring kernel" ?
2. do we change the api to `api/refresh/<kernel_id>`  and call this "refreshing kerne" operation?

we need use this term in both logs and response messages.
pls give this some thought

enterprise_gateway/services/kernels/handlers.py
rahul26goyal: my bad. will change is all places.

enterprise_gateway/services/kernels/handlers.py
rahul26goyal: sure..I can make the change.
I am also having a second thought to call this kernel Refresh API:
`r"/api/kernels/refresh/%s"` ?

enterprise_gateway/services/kernels/handlers.py
rahul26goyal: we need to discuss this further.
what I am trying do here is replace the  `ZMQChannelsHandler` with `RemoteZMQChannelsHandler` for handling the `channels` requests. 

I tried to re-use the same class name on EG but was facing some issue where websocket connection was failing.

enterprise_gateway/services/kernels/remotemanager.py
rahul26goyal: yes @kevin-bates . Originally when I had started to implement it, I was extending the `api/kernels` API handler which would have made it an "kernel update" operation. Thats why there is a lot of reference to the word "update". I will clean this up.

enterprise_gateway/services/kernels/remotemanager.py
rahul26goyal: yes..ll remove

enterprise_gateway/services/kernels/remotemanager.py
rahul26goyal: its an additive operations because:
1. The `KERNEL_EXTRA_SPARK_OPTS` which are passed when the kernel starts come from the ENV variable set on the jupyter notebook/lab server and the notebook users might not know what  configs are set. 
2. With `%%configure ` magic, users is either trying to override those default or add new properties. 
Thus, its an additive operation on top on the original `KERNEL_EXTRA_SPARK_OPTS` which was passed while creating the kernel.

**does Spark take the last duplicated value it finds?**
Yes.

kevin-bates:
> @kevin-bates : I am help is deciding the right terminology for the operation we are performing on the kernel using this new `configure` API:
> 
> 1. do we call it " refreshing kernel" or "re-configuring kernel" ?
> 2. do we change the api to `api/refresh/<kernel_id>`  and call this "refreshing kerne" operation?
> 
> we need use this term in both logs and response messages. pls give this some thought

I guess _refresh_ seems a little easier to understand than _reconfigure_.  Does this imply the magic name would change to `%%refresh` and does that conflict with existing magics?  I think having the terminology match the magic name would be helpful.

I would also like to see the endpoint be _under_ `api/kernels` rather than a sibling to `api/kernels`.  Do you agree?  If not, could you please help me understand why not?  Is adding an endpoint under `api/kernels` violating some kind of convention?

kevin-bates:
Hi @rahul26goyal - what is the status of this PR since it's been about 6 weeks since its last update and it seems there are a few things still to work out?

PR #1106, opened 2022-06-08
Enabling specifying annotations for service accounts in helm chart

afaytelson:


lresende (approved):
LGTM, @kevin-bates what is your take on the chart version? particular for the one in dev mode (main branch)? Do we have the release script bumping the version when we perform the release?

etc/kubernetes/helm/enterprise-gateway/values.yaml
```diff
@@ -124,7 +125,7 @@ kip:
   serviceAccountName: 'kernel-image-puller-sa'
   podSecurityPolicy:
     create: true
-    annotatons: {}
+    annotations: {}
```
kevin-bates, line 128: Nice catch - thank you!

kevin-bates (commented):
The changes look good to me, although we should probably talk about the chart `version`.

In https://github.com/jupyter-server/enterprise_gateway/pull/1106#pullrequestreview-1000366708 @lresende asks:
>  @kevin-bates what is your take on the chart version? particular for the one in dev mode (main branch)? Do we have the release script bumping the version when we perform the release?

My immediate response is, yes, the release script handles this as of #1102.  See: 
https://github.com/jupyter-server/enterprise_gateway/blob/22cd118e8c642ee5373ac922751f8e5afc9d27ca/release.sh#L204-L208

But my question is whether we really need to increment this version string on _every_ change _within a release_ or is just tracking it to the release sufficient?  If the former then it would need to ALWAYS be a manual operation and I would argue that we move to a version string that is completely detached from the EG version - although (IMHO) that would be unfortunate.  I would prefer that we only update this string on releases and ensure it's a syntactically correct semVer (which was the intention of #1102).

afaytelson:
@kevin-bates To reiterate your point - I can change back the Chart version to 3.0.0 and we can use the release script for a pre-release. Then once the Chart is ready for a release then it should be updated manually.

lresende:
@afaytelson Could we leave the version here matching the release version? Thank you.

lresende (approved):

kevin-bates:
>  Then once the Chart is ready for a release then it should be updated manually.

So at that point we'll remove the automation performed in `release.sh` and it will diverge from the EG release number?  

Are charts viewed as their own deliverable packages (separate from EG)?  Why would you increment the Chart version _within_ a release when the Chart should not be delivered _until_ the release is complete?

Issue #1098, opened 2022-05-29
Setting up local development environment (pip, conda, pipenv, etc) 


bloomsa:
Help us improve the Jupyter Enterprise Gateway project by reporting issues
or asking questions.

## Description
Hi all, I'm working on setting up my IDE and local environment with all of the required packages so I can eventually debug EG with breakpoints. (for code exploration purposes) I had some issues getting this set up and I'm wondering if recent changes to the codebase have left this process behind a bit. 

In #1079 the steps to build a conda environment for the project were removed without a replacement process. Some sort of environment, whether it be conda, pipenv, venv, or just the local python environment, containing the required packages for the project is needed in order to run the app and debug. [requirements.yml ](https://github.com/jupyter-server/enterprise_gateway/blob/5b6970daf19e3b07dfd9efb47f7714639182600c/requirements.yml) is still around with a listing of the required packages, but as far as I can tell, only conda knows how to use this. 

For my personal setup, I decided to use pipenv for my python environment. As prerequisite to that, I built the conda env based on `requirements.yml` 

> `conda env create --file requirements.yml --name eg`

I then activated the conda env so that I could generate a requirements.txt by using the following command

> `pip list --format=freeze > requirements.txt` 

From here, I deactivated the conda env and was able to run `pipenv install -r requirements.txt` to generate a Pipfile, Pipfile.lock, and the venv with all required packages. At this point, I'm all setup to run/debug the code in PyCharm.

Let me know if I'm missing something that would have made this easier. If not, I would be happy to contribute and document my process more formally than in this Issue
## Screenshots / Logs

If applicable, add screenshots and/or logs to help explain your problem.
To generate better logs, please run the gateway with `--debug` command line parameter.

## Environment

- Enterprise Gateway Version 2.6.0 (latest)
- Platform: local (macOS laptop)
- PyCharm IDE for code exploration

## Reference
[Process used to generate requirements.txt from conda env](https://stackoverflow.com/a/57845418)

kevin-bates:
Hi @bloomsa.  Yeah, we should probably restore the doc for how to build the conda env and could even restore the make target. What I think should remain removed though is the forced usage of an env on every other makefile target. 
I'm afk until next Wednesday so can't deal with this sooner. If you'd like to create a pr to restore those portions that would be great.

bloomsa:
That makes sense, you got it 👍

PR #1107, opened 2022-06-09
reintroduce conda env setup and docs

bloomsa:
Per #1098, I am reintroducing some of the conda env pieces to assist with locally running EG and with debugging using an IDE. I specifically brought the targets for making and activating the conda environment but did not add that back to all of the other steps as it remains unnecessary for them. 

I also updated the Contributor docs to reflect the additional step that can be used to aid local development

Resolves: #1098

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -32,9 +32,6 @@ jobs:
       - name: Install Python dependencies
         run: |
           pip install ".[test]"
-      - name: Build Jupyter Enterprise Gateway conda env
-        run: |
-          SA="source $CONDA_HOME/bin/activate" make env
       - name: Build and install Jupyter Enterprise Gateway
         uses: nick-invision/retry@v1.0.0
         with:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -32,6 +32,9 @@ jobs:
       - name: Install Python dependencies
         run: |
           pip install ".[test]"
+      - name: Build Jupyter Enterprise Gateway conda env
+        run: |
+          SA="source $CONDA_HOME/bin/activate" make env
       - name: Build and install Jupyter Enterprise Gateway
         uses: nick-invision/retry@v1.0.0
         with:
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,8 @@
     push-kernel-images push-enterprise-gateway push-kernel-py push-kernel-spark-py push-kernel-r push-kernel-spark-r \
     push-kernel-scala push-kernel-tf-py push-kernel-tf-gpu-py push-kernel-image-puller publish helm-chart
 
+SA?=source activate
+ENV:=enterprise-gateway-dev
 SHELL:=/bin/bash
 
 VERSION?=3.0.0.dev0
@@ -37,6 +39,14 @@ help:
 # http://marmelab.com/blog/2016/02/29/auto-documented-makefile.html
 	@grep -E '^[a-zA-Z0-9_-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
 
+build:
+env: ## Make a dev environment
+	-conda env create --file requirements.yml --name $(ENV)
+	-conda env config vars set PYTHONPATH=$(PWD) --name $(ENV)
+
+activate: ## Print instructions to activate the virtualenv (default: enterprise-gateway-dev)
+	@echo "Run \`$(SA) $(ENV)\` to activate the environment."
+
 clean: ## Make a clean source tree
 	-rm -rf dist
 	-rm -rf build
@@ -52,6 +62,9 @@ clean: ## Make a clean source tree
 
 lint: ## Check code style
 	pre-commit run --all-files
+
+nuke: ## Make clean + remove conda env
+	-conda env remove -n $(ENV) -y
 
 run-dev: test-install-wheel ## Make a server in jupyter_websocket mode
 	python enterprise_gateway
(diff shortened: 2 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -32,6 +32,9 @@ jobs:
       - name: Install Python dependencies
         run: |
           pip install ".[test]"
+      - name: Build Jupyter Enterprise Gateway conda env
+        run: |
+          SA="source $CONDA_HOME/bin/activate" make env
```
kevin-bates, line 37: I don't think we need to build an env as part of the CI since none of the targets use it explicitly.

Makefile
```diff
@@ -37,6 +39,14 @@ help:
 # http://marmelab.com/blog/2016/02/29/auto-documented-makefile.html
 	@grep -E '^[a-zA-Z0-9_-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
 
+build:
+env: ## Make a dev environment
+	-conda env create --file requirements.yml --name $(ENV)
+	-conda env config vars set PYTHONPATH=$(PWD) --name $(ENV)
+
+activate: ## Print instructions to activate the virtualenv (default: enterprise-gateway-dev)
+	@echo "Run \`$(SA) $(ENV)\` to activate the environment."
```
kevin-bates, line 48: Not sure how useful this will be, but there may be some that find it helpful.

Makefile
```diff
@@ -53,6 +63,9 @@ clean: ## Make a clean source tree
 lint: ## Check code style
 	pre-commit run --all-files
 
+nuke: ## Make clean + remove conda env
```
kevin-bates, line 66: This target name has kind of always bothered me.  Would you mind renaming this to `remove-env`?

docs/source/contributors/devinstall.md
```diff
@@ -42,6 +44,7 @@ itest-yarn                     Run integration tests (optionally) against docker
 kernel-images                  Build kernel-based docker images
 kernelspecs                    Create archives with sample kernelspecs
 lint                           Check code style
+nuke                           Make clean + remove conda env
```
kevin-bates, line 47: See the previous comment about rename.

kevin-bates (commented):
This looks good @bloomsa - thank you.  I just had a couple of comments.  Thanks for you help here.

Makefile
bloomsa: I found it helpful myself as I was getting started with this project since I don't have a lot of familiarity with different python environment tools

kevin-bates: Sounds good - perfect feedback.

docs/source/contributors/devinstall.md
```diff
@@ -42,6 +44,7 @@ itest-yarn                     Run integration tests (optionally) against docker
 kernel-images                  Build kernel-based docker images
 kernelspecs                    Create archives with sample kernelspecs
 lint                           Check code style
+remove-ev                      Make clean + remove conda env
```
kevin-bates, line 47: Shoot, just a typo...
```suggestion
remove-env                     Make clean + remove conda env
```

docs/source/contributors/devinstall.md
```diff
@@ -28,12 +28,14 @@ Enterprise Gateway's build environment is centered around `make` and the corresp
 Entering `make` with no parameters yields the following:
 
 ```
+ activate                       Print instructions to activate the virtualenv (default: enterprise-gateway-dev)
```
kevin-bates, line 31:
```suggestion
activate                       Print instructions to activate the virtualenv (default: enterprise-gateway-dev)
```
Not sure if this was introducing a space here or not, so I suggest removing it.

kevin-bates (approved):
Looks good @bloomsa - thank you!

blink1073:
It might be that it is installing from `docs/requirement.yml` from the top level directory.  Perhaps it needs to be changed to `.` instead of `..`

kiersten-stokes:
Trying `.` does get things further along, but then I get the below error where conda is having trouble resolving the project requirement for `jinja2>=3.1` (from `pyproject.toml`) with the doc requirement for `jinja2<3.1`.

Related, I see that `jinja2<3.1` was required for the docs in order to prevent an `ImportError` (https://github.com/sphinx-doc/sphinx/issues/10291). Looks like this is no longer an issue as of `sphinx==4.0.2` (and `5.0.1` is the latest). I tested removing the jinja cap from `doc-requirements.txt` and creating a new env (with `.` in `requirements.yml`) and everything proceeded as expected!

Also, it may be worth pointing out that `requirements.yml` lists the `jinja2` requirement as `>=2.10` and `pyproject.toml` requires `jinaj2>=3.1`. This isn't a conflict, obviously, but feels relevant to mention.

```
The conflict is caused by:
    The user requested jinja2<3.1
    jupyter-enterprise-gateway 3.0.0.dev0 depends on jinja2>=3.1

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict


Pip subprocess error:
ERROR: Cannot install -r /Users/kierstenstokes/repos/enterprise_gateway/condaenv.urq965k2.requirements.txt (line 1) and jinja2<3.1 because these package versions have conflicting dependencies.
ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/topics/dependency-resolution/#dealing-with-dependency-conflicts

failed

CondaEnvException: Pip failed
```

kevin-bates:
Thanks @blink1073 and @kiersten-stokes. I'm afk for the weekend but, assuming I'm following things, sounds like the jinja dependency requires updating (along with the relative directory adjustment).  Kiersten, if that works for you would you mind opening a pull request?
Thank you both!

bloomsa:
Sorry about this!! 😢 I must not have fully cleaned my env up before trying to rebuild it.. Thanks for fixing it up @kiersten-stokes

PR #1108, opened 2022-06-10
Update dependency requirements to support `make env` target

kiersten-stokes:
When attempting to run the newly re-added `make env` target, environment creation was failing citing "`Directory '..' is not installable. Neither 'setup.py' nor 'pyproject.toml' found.`". The directory in `requirements.yml` is updated here (to `.`) to point to the correct location for doc dependency installation.

After this change was made, another error was raised (below) where conda was unable to resolve the project requirement for `jinja2>=3.1` (from `pyproject.toml`) with the doc requirement for `jinja2<3.1`.
```
The conflict is caused by:
    The user requested jinja2<3.1
    jupyter-enterprise-gateway 3.0.0.dev0 depends on jinja2>=3.1
```

The jinja cap in `doc-requirements.txt` (in place due to https://github.com/sphinx-doc/sphinx/issues/10291) is removed here as the cited issue no longer occurs as of `sphinx==4.0.2`. The `make env` target now succeeds.

I also updated the `jinja` dependency to `>=3.1` in `requirements.yml` to bring it in line with the version in  `pyproject.toml` (although I'm not sure if there are reasons for this difference that would make this ill-advised; please advise if so 🙂 ).

### DIFF
--- a/docs/doc-requirements.txt
+++ b/docs/doc-requirements.txt
@@ -1,5 +1,3 @@
-# https://github.com/sphinx-doc/sphinx/issues/10291
-jinja2<3.1
 # https://github.com/miyakogi/m2r/issues/66
 mistune<1
 myst-parser
--- a/requirements.yml
+++ b/requirements.yml
@@ -31,5 +31,5 @@ dependencies:
   - flake8
 
   - pip:
-      - ..
+      - .
       - -r docs/doc-requirements.txt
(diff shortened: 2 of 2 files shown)


bloomsa:
Thanks! I tested this on my machine too as was able to run the `make env` command as well. Since the conda env was able to be build this time around, I was also able to validate that `make remove-env` still works as intended as well

bloomsa (approved):

rahul26goyal (approved):
LGTM! thanks @kiersten-stokes for fixing it.

kevin-bates (approved):
Thank you @kiersten-stokes.

PR #1110, opened 2022-06-13
Add clean-env task to remove conda environment

lresende:


Makefile
```diff
@@ -60,10 +60,12 @@ clean: ## Make a clean source tree
 	-make -C docs clean
 	-make -C etc clean
 
+clean-env: clean remove-env  ## Make clean + remove conda env
```
kevin-bates, line 63: I prefer we not introduce another "env" target when this could be done with a simple command already. I think the times we need to clean up both the env and the build in a single command will be rare and `clean-env` and `remove-env` are too similar - which leads to confusion.

Makefile
```diff
@@ -60,10 +60,12 @@ clean: ## Make a clean source tree
...
+
 lint: ## Check code style
 	pre-commit run --all-files
 
-remove-env: ## Make clean + remove conda env
+remove-env: ## Remove conda env
```
kevin-bates, line 68: Good catch - this update makes sense.

kevin-bates (commented):
Thanks @lresende.

kevin-bates:
> @kevin-bates what is your preference? use clean-env which aligns with the clean-xxx to remove parts of the build/environment? Or use remove-env?

I suppose I'd prefer to rename `clean-xxx` to `remove-xxx` since its more descriptive, but because that's **lots** of water already under the bridge, we should rename `remove-env` to `clean-env` and have it only remove the env (just like `clean-xxx` only removes `xxx`).  I guess "clean-xxx" gives us leeway to do other things that "remove-xxx" does not.

kevin-bates (commented):
Thanks Luciano.  I just remembered that this change affects the docs and the `make` output in https://github.com/jupyter-server/enterprise_gateway/blob/main/docs/source/contributors/devinstall.md requires an update as well.

PR #1111, opened 2022-06-13
Install pre-commit as part of the lint make task

lresende:
The build wasn't working in clean environments
unless you had manually installed pre-commit dependency.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
 
-.PHONY: help build clean remove-env dev dev-http docs install bdist sdist test release check_dists \
+.PHONY: help clean remove-env dev dev-http docs install bdist sdist test release check_dists \
     clean-images clean-enterprise-gateway clean-demo-base clean-kernel-images clean-enterprise-gateway \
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -39,7 +39,8 @@ help:
 # http://marmelab.com/blog/2016/02/29/auto-documented-makefile.html
 	@grep -E '^[a-zA-Z0-9_-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
 
-build:
+build: dist
```
kevin-bates, line 42: Good catch.  Not sure if we should just remove the `build` target or make it an alias for `dist`.  Do you find yourself attempting `make build` naturally?  If so, should we document `build` in the help message that is produced when typing `make`?

Makefile
```diff
@@ -61,6 +62,7 @@ clean: ## Make a clean source tree
 	-make -C etc clean
 
 lint: ## Check code style
+	pip install pre-commit
```
kevin-bates, line 65: Could we make this silent since it may only do something once?
```suggestion
	@pip install -q pre-commit
```

kevin-bates (commented):
Thanks @lresende - I left a couple of comments/questions.

kevin-bates (approved):
LGTM - thanks @lresende

Issue #1112, opened 2022-06-13
Interrupts not working on PySpark Kernel


SamRanjha:
## Description
Jupyter Notebook allows user to interrupt the cell execution by sending the interrupt command. But when we are using PySpark Kernel and we interrupt a cell execution, even though the kernel becomes idle on the UI but in the background the job keeps on executing. So, since the UI is idle we can send new commands and the kernel goes to busy but in the driver the old job keep on executing and the execution of the new job starts only after the first job execution is finished.
We do receive signum : 2 in logs for the "launch_ipykernel.py" but the execution continues.
Thus, the interrupt feature is broken for PySpark Kernels.

## Reproduce
1. Start a long running job on Jupyter Notebook.
2. Interrupt the job.
<img width="1536" alt="Screenshot 2022-06-13 at 9 00 01 PM" src="https://user-images.githubusercontent.com/30014834/173390800-f248bb05-6304-4a0e-b19b-7c10fc1c5362.png">

<img width="1178" alt="Screenshot 2022-06-13 at 9 03 53 PM" src="https://user-images.githubusercontent.com/30014834/173390599-82e1f8a2-1ca6-4840-9cc7-67fa47c2c691.png">

3. The kernel becomes idle but the job keeps on executing.
4. The execution gets completed in the driver logs.
<img width="1433" alt="Screenshot 2022-06-13 at 9 06 38 PM" src="https://user-images.githubusercontent.com/30014834/173391081-314a9965-06c3-435f-a51f-d349aaae8c33.png">
5. Until the previous execution is completed, we don't receive output for the next executions.
Eg - This job took more than three minutes which usually get executed in seconds.
<img width="1175" alt="Screenshot 2022-06-13 at 9 10 40 PM" src="https://user-images.githubusercontent.com/30014834/173392144-15a0315b-2a8c-424c-83a9-f166483fa610.png">

## Expected behavior
The job should have stopped execution once the interrupt signal was received.

## Context

- Operating System and version: Kubernetes v1.21.12

- Browser and version: Firefox, 91.10.0esr (64-bit)
- Jupyter Notebook version:  6.4.11

<details><summary>Command Line Output</summary>
<pre>
JEG 
[I 220613 15:29:50 web:2243] 200 GET /api/swagger.json (192.168.163.57) 0.82ms
[I 220613 15:29:50 web:2243] 200 GET /api/swagger.json (192.168.133.171) 0.76ms
[...]

PR #1114, opened 2022-06-14
Changes for handling interrupts for PySpark Kubernetes Kernel

SamRanjha:
This PR fixes the issue #1112 

Testing:
1. Started a PySpark kernel with the changes.
2. Started kernels and executed code followed by interrupting the same. Verified job gets killed in Spark Logs and also we can start executing other cells. 
3. Changed the spark.ui.port, added the env variable to driver and verified interrupt works.
4. Ran jobs in threads and verified all running jobs were killed.

### DIFF
--- a/etc/kernel-launchers/python/scripts/launch_ipykernel.py
+++ b/etc/kernel-launchers/python/scripts/launch_ipykernel.py
@@ -7,6 +7,7 @@ import random
 import socket
 import tempfile
 import uuid
+import requests
 from multiprocessing import Process
 from threading import Thread
 
@@ -35,6 +36,9 @@ logger = logging.getLogger("launch_ipykernel")
 logger.setLevel(log_level)
 
 DEFAULT_KERNEL_CLASS_NAME = "ipykernel.ipkernel.IPythonKernel"
+# If one needs to modify the Spark UI Port then the user
+# also needs to be add "spark.kubernetes.driverEnv.SPARK_UI_PORT" to the spark config
+SPARK_UI_PORT_KEY = 'SPARK_UI_PORT'
 
 
 class ExceptionThread(Thread):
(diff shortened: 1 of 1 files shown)


etc/kernel-launchers/python/scripts/launch_ipykernel.py
```diff
@@ -360,7 +364,64 @@ def get_server_request(sock):
     return request_info
 
 
-def server_listener(sock, parent_pid):
+def get_application_id(ui_port):
+    application_id = ""
+    url = f"http://localhost:{ui_port}/api/v1/applications/"
```
lresende, line 369: Should this be actually based on the Spark master URL rather than assuming always localhost? Maybe instead of using an env var for the port, we should just propagate the master URL?

SamRanjha: When you say "propagate the master URL" you mean share it with the server listener process? I tried sharing ui_port by setting it during spark context initialisation and sharing with manager.value but that didn't work.

So do you have any alternative for this?

kevin-bates, line 369: Are `http` and `localhost` always going to be correct?

etc/kernel-launchers/python/scripts/launch_ipykernel.py
```diff
@@ -360,7 +364,64 @@ def get_server_request(sock):
...
+    response = requests.get(url)
+    if response.status_code == 200 and response.json():
+        application_id = response.json()[0].get("id")
```
kevin-bates, line 372: How do we know the _first_ id is _our_ applicationId?

etc/kernel-launchers/python/scripts/launch_ipykernel.py
```diff
@@ -360,7 +364,64 @@ def get_server_request(sock):
...
+
+    return application_id
+
+
+def get_active_jobs_id(ui_port, application_id):
+    active_job_ids = []
+    url = "http://localhost:{}/api/v1/applications/{}/jobs?status=running".format(
+        ui_port, application_id
+    )
```
kevin-bates, line 381: Can we use an f-string here?

etc/kernel-launchers/python/scripts/launch_ipykernel.py
```diff
@@ -360,7 +364,64 @@ def get_server_request(sock):
...
+    return active_job_ids
+
+
+def kill_job(ui_port, jobid):
+    url = f"http://localhost:{ui_port}/jobs/job/kill/?id={jobid}"
+    response = requests.get(url)
+    if response.status_code == 200:
+        logger.info(f"Killed job with id: {jobid}")
+    else:
+        logger.warning(f"Failed to kill job with id: {jobid}")
```
kevin-bates, line 397: I think it would be a good idea to also include the applcationId (and kernel_id) in these messages so "paper trails" can be established...
Perhaps something like:
"Killed job '{job_id}' of application '{application_id}' and kernel '{kernel_id}'"

kevin-bates (commented):
It's great to hear this approach works - thank you.  That said, it feels a little fragile to me.  It would ideal to leverage the content in the spark context and I wonder if you've explored things like a signal handler that traps a user signal (`SIGUSR` or equivalent) where the sighandler (in the parent) could then access the spark context and do the right thing.

etc/kernel-launchers/python/scripts/launch_ipykernel.py
lresende: What I am saying is, instead of providing a env var with just the master port, use one for the whole spark master unless that is available somewhere else.

PR #1121, opened 2022-06-23
Make EnterpriseGatewayApp.shutdown more robust, since it's a cleanup procedure

CiprianAnton:
In our logs, it happened to get a `raise web.HTTPError(404, u'Kernel does not exist: %s' % kernel_id)` exception in enterprisegatewayapp.shutdown, thus the cleanup being compromised.

Since this function is more a cleanup function, it shouldn't fail and cleanup as much as possible instead. Otherwise, kernels pods may leak.

The "Shut down complete" log is useful to see if termination grace period is enough and EG is not being killed.


kevin-bates (approved):
Good idea @CiprianAnton - thank you.

rahul26goyal:
thanks you @CiprianAnton for fixing this.

Issue #1127, opened 2022-07-08
support dynamic load new kernel spec


wuyueandrew:
### Problem

for now, we should restart eg after install a new kernel spec and it is not convenient. cuz other users cannot connect to gateway 
due to restarting eg.

### Proposed Solution

only support installing new kernel specs. elder kernel  in used are not support to escape running problem.

kevin-bates:
Hi @wuyueandrew - it is not typically necessary to restart EG (or any jupyter server) to use additional kernelspecs added after the server has started.  When kernelspecs are requested via, for example, `curl -X GET -i http://localhost:8888/api/kernelspecs`, the server will hit the filesystem to locate any new kernel specifications, and return all that it finds.

However, if this is a Kubernetes deployment, we are using the `KernelSpecManager.whitelist` configurable (set via env `EG_KERNEL_WHITELIST`) which _would_ require the need to restart and redeploy EG.

If this is not a containerized env, then you could get the dynamic behavior you need by either _not_ setting `KernelSpecManager.whitelist` and getting the default (dynamic) behavior, or by configuring it within the `jupyter_enterprise_gateway_config.py` configuration file and setting `EnterpriseGatewayApp.c.EnterpriseGatewayConfigMixin.dynamic_config_interval` to a non-zero (and positive) value.  In this case, updates to the config file are then detected and reflected in the subsequent kernelspec fetch request.

If you're using a containerized environment, then, yes, we should introduce a way to NOT [set the whitelist](https://github.com/jupyter-server/enterprise_gateway/blob/c878033231f90974f174e5f51fd68189c2b85de1/etc/kubernetes/helm/enterprise-gateway/values.yaml#L80-L89), so that users can have the ability to add kernelspecs dynamically.  Since the configuration file would live in the container, accessing would require mounts and an adjustment to the [`start-enterprise-gateway.sh`](https://github.com/jupyter-server/enterprise_gateway/blob/c878033231f90974f174e5f51fd68189c2b85de1/etc/docker/enterprise-gateway/start-enterprise-gateway.sh#L29-L38) script to NOT configure the whitelist via command-line arguments since CLI configurables are not privy to dynamic config detection.

If K8s, perhaps we could allow for the entry in values.yaml to be:
```
whitelist:
  - *
```
[...]

PR #1131, opened 2022-07-12
Enable ability to not filter kernels in container deployments

kevin-bates:
As discussed in #1127, there is currently no way to easily enable dynamic kernelspec additions to containerized deployments because the deployment scripts always specific kernelspec entries.  Since dynamic kernelspec additions are default behaviors (when kernelspecs are not explicitly listed in the configuration), we should enable similar support in containerized deployments.  This pull request enables that functionality.

The `start-enterprise-gateway.sh` that is the `CMD` in the enterprise-gateway image has been modified to detect if the `EG_KERNEL_WHITELIST` env has a value of `null`.  When that value is detected, the `--KernelSpecManager.whitelist` command-line option is omitted, thereby exposing all kernelspecs configured into the image.  It is presumed that deployments wishing to dynamically add kernelspec definitions have mounted the `/usr/local/share/jupyter/kernels` directory (or otherwise made its update available).  Updates made to that directory will be automatically picked up when kernelspecs are refreshed.

A value of `null` was used for `EG_KERNEL_WHITELIST` to make this determination because that is the value Helm will set when the `whitelist:` entry of the `values.yaml` has no entries.  This allows operators to easily remove (or comment out) all existing kernelspec entries to enable this functionality.

Resolves: #1127

### DIFF
--- a/etc/docker/enterprise-gateway/start-enterprise-gateway.sh
+++ b/etc/docker/enterprise-gateway/start-enterprise-gateway.sh
@@ -30,12 +30,20 @@ EG_KERNEL_WHITELIST=${EG_KERNEL_WHITELIST:-"'r_docker','python_docker','python_t
 export EG_KERNEL_WHITELIST=`echo ${EG_KERNEL_WHITELIST} | sed 's/[][]//g'` # sed is used to strip off surrounding brackets as they should no longer be included.
 export EG_DEFAULT_KERNEL_NAME=${EG_DEFAULT_KERNEL_NAME:-python_docker}
 
+# Determine whether the kernels-allowed list should be added to the start command.
+# This is conveyed via a 'null' value for the env - which indicates no kernel names
+# were used in the helm chart.
+allowed_kernels_option="--KernelSpecManager.whitelist=[${EG_KERNEL_WHITELIST}]"
+if [ "${EG_KERNEL_WHITELIST}" == 'null' ]; then
+	allowed_kernels_option=""
+fi
+
 
 echo "Starting Jupyter Enterprise Gateway..."
 
 exec jupyter enterprisegateway \
 	--log-level=${EG_LOG_LEVEL} \
-	--KernelSpecManager.whitelist=[${EG_KERNEL_WHITELIST}] \
+	${allowed_kernels_option} \
 	--RemoteMappingKernelManager.cull_idle_timeout=${EG_CULL_IDLE_TIMEOUT} \
 	--RemoteMappingKernelManager.cull_interval=${EG_CULL_INTERVAL} \
 	--RemoteMappingKernelManager.cull_connected=${EG_CULL_CONNECTED} \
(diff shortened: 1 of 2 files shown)


bloomsa (commented):
thanks for doing this! this directly improves my deployment of EG. 👍

etc/docker/enterprise-gateway/start-enterprise-gateway.sh
```diff
@@ -30,12 +30,20 @@ EG_KERNEL_WHITELIST=${EG_KERNEL_WHITELIST:-"'r_docker','python_docker','python_t
...
+	allowed_kernels_option=""
+fi
+
 
 echo "Starting Jupyter Enterprise Gateway..."
 
 exec jupyter enterprisegateway \
 	--log-level=${EG_LOG_LEVEL} \
-	--KernelSpecManager.whitelist=[${EG_KERNEL_WHITELIST}] \
+	${allowed_kernels_option} \
```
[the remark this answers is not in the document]
kevin-bates: Hi Sam.  I joined `${allowed_kernels_option}` with the log level option as it seems less risky to future changes (properly escaping `\`, etc.).

etc/docker/enterprise-gateway/start-enterprise-gateway.sh
```diff
@@ -30,12 +30,19 @@ EG_KERNEL_WHITELIST=${EG_KERNEL_WHITELIST:-"'r_docker','python_docker','python_t
 export EG_KERNEL_WHITELIST=`echo ${EG_KERNEL_WHITELIST} | sed 's/[][]//g'` # sed is used to strip off surrounding brackets as they should no longer be included.
 export EG_DEFAULT_KERNEL_NAME=${EG_DEFAULT_KERNEL_NAME:-python_docker}
 
+# Determine whether the kernels-allowed list should be added to the start command.
+# This is conveyed via a 'null' value for the env - which indicates no kernel names
+# were used in the helm chart.
+allowed_kernels_option="--KernelSpecManager.whitelist=[${EG_KERNEL_WHITELIST}]"
+if [ "${EG_KERNEL_WHITELIST}" == 'null' ]; then
+	allowed_kernels_option=""
+fi
```
bloomsa, line 39: I think it makes more sense to have `allowed_kernels_option` default to "". Then change the if to check if  "${EG_KERNEL_WHITELIST}" != 'null' and set the flag inside the if.

bloomsa (commented):
one change requested for readability

etc/docker/enterprise-gateway/start-enterprise-gateway.sh
kevin-bates: Good idea.  I also switched the default value to `null` since I found that the `docker-compose.yaml` script includes the Docker-based kernel specs.

bloomsa (approved):

PR #1135, opened 2022-07-25
Add typing hints to session managers and handler

### LINKED ISSUE #850: Add typing hints to session managers and handler
Part of the larger [umbrella issue](https://github.com/jupyter/enterprise_gateway/issues/847) this issue will track adding typing hints to the session managers and handlers located in `enterprise_gateway/services/sessions`.

kiersten-stokes:
Fixes #850 

Most of these changes are pretty straightforward. The only uncertainty I have is the type hint for the `kernel_manager` `__init__` param for the session managers. I have the type as `MappingKernelManager` right now. The docstring of the base session manager is listed as `SeedingMappingKernelManager`, but the `RemoteMappingKernelManager` is what I see during debug. I split the difference with the common base class `MappingKernelManager`.

### DIFF
--- a/enterprise_gateway/services/sessions/kernelsessionmanager.py
+++ b/enterprise_gateway/services/sessions/kernelsessionmanager.py
@@ -142,10 +142,11 @@ class KernelSessionManager(LoggingConfigurable):
         finally:
             kernels_lock.release()
 
-    def start_session(self, kernel_id: str) -> bool:
+    def start_session(self, kernel_id: str) -> Optional[bool]:
         kernel_session = self._sessions.get(kernel_id, None)
         if kernel_session is not None:
             return self._start_session(kernel_session)
+        return None
 
     def start_sessions(self) -> None:
         """
--- a/enterprise_gateway/services/sessions/kernelsessionmanager.py
+++ b/enterprise_gateway/services/sessions/kernelsessionmanager.py
@@ -15,8 +15,6 @@ from requests.auth import HTTPBasicAuth, HTTPDigestAuth
 from traitlets import Bool, CaselessStrEnum, Unicode, default
 from traitlets.config.configurable import LoggingConfigurable
 
-from enterprise_gateway.services.kernels.remotemanager import RemoteMappingKernelManager
-
 kernels_lock = threading.Lock()
 
 # These will be located under the `persistence_root` and exist
@@ -73,7 +71,7 @@ class KernelSessionManager(LoggingConfigurable):
     def persistence_root_default(self) -> str:
         return os.getenv(self.persistence_root_env, "/")
 
-    def __init__(self, kernel_manager: RemoteMappingKernelManager, **kwargs):
+    def __init__(self, kernel_manager: RemoteMappingKernelManager, **kwargs):  # noqa: F821
         super().__init__(**kwargs)
         self.kernel_manager = kernel_manager
         self._sessions = dict()
@@ -337,7 +335,7 @@ class FileKernelSessionManager(KernelSessionManager):
     def persistence_root_default(self) -> str:
         return os.getenv(self.persistence_root_env, jupyter_data_dir())
 
-    def __init__(self, kernel_manager: RemoteMappingKernelManager, **kwargs):
+    def __init__(self, kernel_manager: RemoteMappingKernelManager, **kwargs):  # noqa: F821
         super().__init__(kernel_manager, **kwargs)
         if self.enable_persistence:
             self.log.info(f"Kernel session persistence location: {self._get_sessions_loc()}")
@@ -449,7 +447,7 @@ class WebhookKernelSessionManager(KernelSessionManager):
     def auth_type_default(self) -> str | None:
         return os.getenv(self.auth_type_env, None)
 
-    def __init__(self, kernel_manager: RemoteMappingKernelManager, **kwargs):
+    def __init__(self, kernel_manager: RemoteMappingKernelManager, **kwargs):  # noqa: F821
         super().__init__(kernel_manager, **kwargs)
         if self.enable_persistence:
             self.log.info("Webhook kernel session persistence activated")
--- a/enterprise_gateway/services/sessions/kernelsessionmanager.py
+++ b/enterprise_gateway/services/sessions/kernelsessionmanager.py
@@ -6,9 +6,11 @@ import getpass
 import json
 import os
 import threading
+from typing import List, Optional
 
 import requests
 from jupyter_core.paths import jupyter_data_dir
+from jupyter_server.services.kernels.kernelmanager import MappingKernelManager
 from requests.auth import HTTPBasicAuth, HTTPDigestAuth
 from traitlets import Bool, CaselessStrEnum, Unicode, default
 from traitlets.config.configurable import LoggingConfigurable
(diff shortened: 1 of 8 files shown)


kevin-bates:
Hi @kiersten-stokes - these changes look good.  The `SeedingMappingKernelManager` was the `MappingKernelManager` introduced by [Jupyter _Kernel_ Gateway](https://github.com/jupyter-server/kernel_gateway/blob/c72803ab1ea4bb75847917e0a7c682acbe700fbe/kernel_gateway/services/kernels/manager.py#L11).  Since we use `RemoteMappingKernelManager` exclusively within EG, I think it would be best to update all `MappingKernelManager` type references to `RemoteMappingKernelManager` as well as the help-string refs of `SeedingMappingKernelManager` (if you could please).  Thank you.

enterprise_gateway/services/sessions/kernelsessionmanager.py
```diff
@@ -6,13 +6,16 @@
...
 import threading
+from typing import List, Optional
 
 import requests
 from jupyter_core.paths import jupyter_data_dir
 from requests.auth import HTTPBasicAuth, HTTPDigestAuth
 from traitlets import Bool, CaselessStrEnum, Unicode, default
 from traitlets.config.configurable import LoggingConfigurable
 
+from enterprise_gateway.services.kernels.remotemanager import RemoteMappingKernelManager
```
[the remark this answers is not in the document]
kevin-bates: Could we give the `from __future__ import annotations` trick a try here?  @bloomsa introduced that in the process-proxy hints PR: https://github.com/jupyter-server/enterprise_gateway/pull/1129#issuecomment-1185719856

akchinSTC (approved):
LGTM

lresende:
@kiersten-stokes are we still waiting on an update to @kevin-bates comments above or is this ready to merge?

enterprise_gateway/services/sessions/kernelsessionmanager.py
```diff
@@ -66,16 +68,16 @@ def session_persistence_default(self):
     )
 
     @default("persistence_root")
-    def persistence_root_default(self):
+    def persistence_root_default(self) -> str:
         return os.getenv(self.persistence_root_env, "/")
 
-    def __init__(self, kernel_manager, **kwargs):
+    def __init__(self, kernel_manager: RemoteMappingKernelManager, **kwargs):  # noqa: F821
```
kevin-bates, line 74: Is there a reason why the same [approach taken in #1129](https://github.com/jupyter-server/enterprise_gateway/pull/1129/files#diff-300bcd9b5be9a895a110de4b9fa865ef6c13aeadc1e1e8e3246c22839e8e64bbR4-R37) can't be applied here?  This would avoid the `noqa` statement and allow for the import which seems odd to not have.

This [SO hit seems to state similar](https://stackoverflow.com/questions/53605806/mypy-flake8-is-there-any-way-to-surpress-warning-of-f821-undefined-name) although its not the accepted answer (albeit more popular).

Hmm, if I try to apply the same set of changes as in #1129, I see that the build fails on `make test-install-wheel` with:
```
  File "/opt/miniconda3/envs/eg-dev/bin/jupyter-enterprisegateway", line 5, in <module>
    from enterprise_gateway.enterprisegatewayapp import launch_instance
  File "/opt/miniconda3/envs/eg-dev/lib/python3.9/site-packages/enterprise_gateway/enterprisegatewayapp.py", line 31, in <module>
    from .services.kernels.remotemanager import RemoteMappingKernelManager
  File "/opt/miniconda3/envs/eg-dev/lib/python3.9/site-packages/enterprise_gateway/services/kernels/remotemanager.py", line 19, in <module>
  File "/opt/miniconda3/envs/eg-dev/lib/python3.9/site-packages/enterprise_gateway/services/processproxies/processproxy.py", line 51, in <module>
  File "/opt/miniconda3/envs/eg-dev/lib/python3.9/site-packages/enterprise_gateway/services/sessions/kernelsessionmanager.py", line 18, in <module>
    from ..kernels.remotemanager import RemoteMappingKernelManager
ImportError: cannot import name 'RemoteMappingKernelManager' from partially initialized module 'enterprise_gateway.services.kernels.remotemanager' (most likely due to a circular import) (/opt/miniconda3/envs/eg-dev/lib/python3.9/site-packages/enterprise_gateway/services/kernels/remotemanager.py)
make: *** [test-install-wheel] Error 1
```
and I wonder if the difference here is that in #1129, the _runtime_ circular reference is not detected because process-proxies are loaded dynamically.

@kiersten-stokes - did you first attempt the #1129 approach but then settled on `noqa` due to this?

Hmm, interesting.  On closer inspection of #1129, I [do see the use of `noqa` without the actual import statement](https://github.com/jupyter-server/enterprise_gateway/pull/1129/files#diff-b63cd0218a1f0491e50a03a2a676575b152556cd098ed428b80d24c8dd8c6553R400) occurring while the others (that don't use `noqa`) reside in truly remote process proxies - which lends credence to this theory since `process-proxies.py` will be loaded for local kernels.  As a result, I think we should try to determine if remote process proxies might be problematic when their respective modules get loaded and may need to circle back on #1129.   (I will take a look.)

In the meantime, since we do have precedence of using `noqa` for this issue, I'm going to approve and merge this PR since my suspicion is that this is what we need to do.

kevin-bates (approved):
These changes look good.  I suspect we might have an issue with #1129 per https://github.com/jupyter-server/enterprise_gateway/pull/1135#discussion_r931468908, but let's move forward with this PR.

enterprise_gateway/services/sessions/kernelsessionmanager.py
kiersten-stokes: Yeah, this is the exact process I went through! Attempted the #1129 approach and then got the error above during build. It was only after that that I noticed `noqa` was being used in certain 1129 modules. Though I hadn't quite gotten to the bottom of why yet, but I like the build vs runtime hypothesis

kevin-bates: > I think we should try to determine if remote process proxies might be problematic when their respective modules get loaded and may need to circle back on https://github.com/jupyter-server/enterprise_gateway/pull/1129. (I will take a look.)

I realized that our `DistributedProcessProxy` and `YarnClusterProcessProxy` classes are used in the integration tests - which appear clean in #1129.  To eliminate any questions I decided to build the `elyra/enterprise-gateway-demo` image (confirming that it does indeed contain the code in question) and the integration tests pass.

I find it odd that the `noqa` is not necessary for the remote process proxies and it must be the deferred load that changes this particular set of behaviors.  Oh well - this is definitely the path of least resistance so "let's go!".

Thanks for the contribution @kiersten-stokes!

bloomsa: Hi, I know the change was merged but I saw the other type hinting PR mentioned here that I worked on so I thought I'd give my 2¢. I ran into issues like this on some of the processProxy classes as you mentioned. I _think_ it came down to bugs or misleading behavior in flake, the IDEs, or other linting tools. For example, Flake wants to import the class being referenced as a type hint (even if  `from __future__ import annotations` is used, even though that import means that type hints resolve to simply Strings rather than actual class references) [This proposal](https://peps.python.org/pep-0649/) goes into detail about it but if you skim it you'll see that "from __future__ import annotations, in which annotations are converted back into strings and must be reparsed and executed by eval() to be used."

I'm wondering if any update to flake or other linting tools used in the build process would clean this up? I haven't done enough research into those to know for sure.
@kevin-bates @kiersten-stokes

Issue #848, opened 2020-08-18
Add typing hints to kernel management and handlers


kevin-bates:
Part of the larger [umbrella issue](https://github.com/jupyter/enterprise_gateway/issues/847) this issue will track adding typing hints to the manager and handlers located in `enterprise_gateway/services/kernels`.

rahul26goyal:
@kevin-bates : I can work on these if someone could guide with it.
I hav no prior experienc with JEG but wiling to learn and contribute :)

kevin-bates:
Hi @rahul26goyal - thank you for your interest in contributing to EG!  I will be happy to help you with this issue.

If you have any questions on a given type or where to locate things, just ping this issue.  I think you'll find most of the uses of this stuff occurs in the [jupyter_client](https://github.com/jupyter/jupyter_client) repo.  However, any process-proxy references will be in the [processproxies](https://github.com/jupyter/enterprise_gateway/tree/master/enterprise_gateway/services/processproxies) sub-package in EG.

PR #1140, opened 2022-07-29
Add type hints to kernels service

kevin-bates:
This PR adds type hints to the methods within the `services/kernels` modules. 

One item I would like to discuss is the return value on `async def` methods.  I added a return value of `Awaitable[str]` on the `start_kernel` method, but the other async methods (here and elsewhere in EG to the best of my knowledge) do not use the `Awaitable` type and I'm curious what others think.

The other thing I noticed is that my IDE (PyCharm) is hi-lighting `kernel_id` as an _expected_ type of `uuid.pyi` where I've added `str` yet none of the dependent repositories (`jupyter_server` and `jupyter_client`) specify `uuid.pyi` as a type hint (few even specify `str`).  I suspect this particular annotation is coming from the _installed_ package area via the `py.typed` file (although it is zero-length), but I'm not sure.  Does anyone know what might be happening there?  I suspect its some kind of `mypy` thing that `jupyter_server` and `jupyter_client` are built with.

Fixes: #848


lresende (approved):

kevin-bates:
> One item I would like to discuss is the return value on async def methods.

Regarding my question regarding how to type-hint the return type of `async def` methods, it looks like `-> Awaitable[`_`return-type`_`]:` should really just be `-> `_`return-type`_`:` per https://mypy.readthedocs.io/en/stable/more_types.html#typing-async-await, so I'm going to make another commit that updates [this occurrence](https://github.com/jupyter-server/enterprise_gateway/pull/1140/files#diff-4506972f3901d94f98686c3af28ccfb7792fa8d1f105622025a9a16a8d0a8299R183-R185) to `str` and addresses the CodeQL issue that also exists.

PR #1143, opened 2022-08-04
Use local kernel images if they exist

kevin-bates:
Prior to these changes, the Kernel Image Puller would unconditionally pull images on its first pass when the policy was `IfNotPresent` despite the fact that applicable images are locally present.  This prevented the ability to use locally-defined images - which is useful in dev environments, but also in environments in which you wish "pre-seed" images.

There are two areas of change.
1. The KIP has been updated to check for image existence even when the image has yet to be pulled (i.e., the first-time scenario).  If the image exists, it considers it pulled and adds it to KIP's pulled-images set.  KIP will continue to monitor the images, just as it does today.
2. The helm chart has been modified to tolerate a missing `Values.kip.defaultContainerRegistry` value in which case the corresponding env `KIP_DEFAULT_CONTAINER_REGISTRY` will not be set when KIP is started.  
    a. Related to this, KIP now defaults the non-existence of `KIP_DEFAULT_CONTAINER_REGISTRY` to the empty string `""`, which then prevents a registry-identifier from being prefixed to the image name if the `kernel.json` doesn't already include one.  

Change 2 just makes it easier to convey that no registry should be used.  However, it should be noted (probably when using the Docker runtime) that `docker.io` is still the default registry when the empty string is in play.  I believe this is a function of the `DockerClient` used to pull images.

In my testing, even when the default registry was `docker.io` local images named `elyra/kernel-py:dev` would still be used.  If those images did not exist or were removed while KIP was running, KIP will pull `docker.io/elyra/kernel-py:dev`.  However, when `quay.io` was configured as `KIP_DEFAULT_CONTAINER_REGISTRY`, local images will be pulled from `quay.io` unless the local image included the `quay.io` prefix (e.g., `quay.io/elyra/kernel-py:dev`).

### DIFF
--- a/etc/docker/kernel-image-puller/kernel_image_puller.py
+++ b/etc/docker/kernel-image-puller/kernel_image_puller.py
@@ -46,7 +46,7 @@ class KernelImagePuller:
         self.num_pullers = int(os.getenv("KIP_NUM_PULLERS", "2"))
         self.num_retries = int(os.getenv("KIP_NUM_RETRIES", "3"))
         self.policy = os.getenv("KIP_PULL_POLICY", KernelImagePuller.POLICY_IF_NOT_PRESENT)
-        self.default_container_registry = os.getenv("KIP_DEFAULT_CONTAINER_REGISTRY", "docker.io")
+        self.default_container_registry = os.getenv("KIP_DEFAULT_CONTAINER_REGISTRY", "")
         self.runtime_endpoint = os.getenv(
             "KIP_CRI_ENDPOINT", "unix:///run/containerd/containerd.sock"
         )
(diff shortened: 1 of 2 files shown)


kiersten-stokes (approved):
Having the option to use local images makes a lot of sense to me. Confirmed that locally built images skip pull! 

```
[D 2022-08-08 19:28:01,866 kernel_image_puller.t1] Waiting for new image to pull
[I 2022-08-08 19:28:01,866 kernel_image_puller.t1] Task received to pull image: elyra/kernel-tf-py:dev
[D 2022-08-08 19:28:01,886 kernel_image_puller.t1] Checked existence of image 'elyra/kernel-tf-py:dev' in 0.020 secs. exists = True
[I 2022-08-08 19:28:01,886 kernel_image_puller.t1] Image 'elyra/kernel-tf-py:dev' has not been pulled but exists, and policy is 'IfNotPresent'. Skipping pull.
[D 2022-08-08 19:28:01,886 kernel_image_puller.t1] Waiting for new image to pull
[D 2022-08-08 19:28:01,890 kernel_image_puller.t2] Checked existence of image 'elyra/kernel-scala:dev' in 0.024 secs. exists = True
[I 2022-08-08 19:28:01,890 kernel_image_puller.t2] Image 'elyra/kernel-scala:dev' has not been pulled but exists, and policy is 'IfNotPresent'. Skipping pull.
```

kevin-bates:
Thanks for the review @kiersten-stokes!

PR #1146, opened 2022-08-10
Address delayed shutdown behavior in k8s kernels

kevin-bates:
This pull request attempts to address the delay seen when shutting down k8s-managed kernels.  I believe the issue is because the pod (and namespace if not BYO namespace) is not terminated when the kernel process has stopped.  Previously, the pod/namespace is not deleted until the cleanup phase, but these changes move the container resource termination to when the listener is shut down.  This allows the polling to check that the "kernel" has terminated to leverage the container status, which, previously, because it had yet to occur, was causing the fallback termination behavior to occur.

I suspect this was not a problem in older versions of kubernetes because (just a hunch) when the container's process terminated the container terminated.  However, I suspect k8s has "grown" enough where interaction with the resource manager is necessary and has introduced some delay.

There are several areas of race condition with starting/stopping pods from other applications, so I think this improves on those, but may not necessarily be the final solution.

### DIFF
--- a/enterprise_gateway/services/processproxies/container.py
+++ b/enterprise_gateway/services/processproxies/container.py
@@ -150,7 +150,6 @@ class ContainerProcessProxy(RemoteProcessProxy):
         # See https://github.com/jupyter-server/enterprise_gateway/issues/827
         if container_status in self.get_initial_states():
             result = None
-
         return result
 
     def send_signal(self, signum: int) -> bool | None:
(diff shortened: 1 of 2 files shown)


kiersten-stokes (approved):
LGTM! Cut termination time roughly in half in my tests

lresende (approved):

kiersten-stokes (commented):
One non-blocking NIT that I just came across: I think [this comment](https://github.com/jupyter-server/enterprise_gateway/blob/b2d74ed7d161a2af63c9461bc0b6b9bf8501eba4/enterprise_gateway/services/processproxies/k8s.py#L104-L106) explaining the functionality of `terminate_container_resources` may want a small update as a result of the changes in this PR

kevin-bates:
Good point @kiersten-stokes.  I'm also going to [remove the call to `get_container_status()` within the namespace deletion block ](https://github.com/jupyter-server/enterprise_gateway/pull/1146/files#diff-8c5248ba4616baed286b229ad44a404f730c6302735f7375aec921d0a1ba191cR157-R160)because it really reflects the state of the pod and not the namespace and was unconditionally invoked previously.  As a result, we should hold off on this merge until I test this and push the commit.

kiersten-stokes (approved):

PR #1147, opened 2022-08-13
Ensure kernel session retrieve from resp is correct type

chiawchen:
In `load_sessions()`, we need to traverse an iterable list `kernel_sessions` to get `kernel_session` . However if we retrieve the response with `response.content`, it actually give us back byte string, and leads to iterate on byte char instead of a dict object. Therefore, this PR address the improper used `content` to `json()` and rm the `json.loads()` part in `_load_session_from_response`

[Test Plan]
Having a customized WebhookKernelSessionManager deploy on Kubernetes and works properly with these changes.


kevin-bates (approved):
These changes look good (and appropriate) - great catch @chiawchen, and I'm glad to see you're finding this contribution helpful.

I'm also adding @amazingarman who contributed this feature just to make sure we're all on the same page.

Issue #978, opened 2021-07-09
Running notebook programmatically, nbclient doesnt use the kernel from the kernelspec


achandak123:
## Description
As a followup of [this](https://github.com/jupyter/enterprise_gateway/issues/972) issue, i created a custom python kubernetes based kernel with boto3 package installed in it. When i run the jupyterlab server against this Enterprise gateway (EG) and chose the custom kernel,  i am able to use boto3.

However, when i try running nbClient like this
```
client = NotebookClient(nb=newNb, kernel_manager_class=GatewayKernelManager)
```
I see that EG is launching Python3 kernel, instead of my custom kernel.

This is because of the [this](https://github.com/jupyter-server/jupyter_server/blob/e3b2dada0ae305d4512476ff664b612a87c5bc50/jupyter_server/gateway/managers.py#L398) code, i think.

Was wondering how can i override it to use the custom Kernel, when i call it from nbClient itself?

## Screenshots / Logs
When kernel is launched using the jupyterlab server, it picks up the custom kernel
```
[D 2021-07-08 23:21:04.120 EnterpriseGatewayApp] Launching kernel: 'custom Kernel' with command: ['/opt/conda/bin/python3.7', '/usr/local/share/jupyter/kernels/python_kubernetes/scripts/launch_kubernetes.py', '--RemoteProcessProxy.kernel-id', '1635c0a0-5b8a-4f58-b204-c1b5cab0d8d0', '--RemoteProcessProxy.port-range', '0..0', '--RemoteProcessProxy.response-address', '10.1.61.18:37037']
```

However, when i am running using nbClient, although the **kernelspec** in the ipynb file says "custom Kernel", it picks Python 3 as the kernel.

```
[D 2021-07-09 01:11:32.445 EnterpriseGatewayApp] Launching kernel: 'Python 3' with command: ['/opt/conda/bin/python', '-m', 'ipykernel_launcher', '-f', '/home/jovyan/.local/share/jupyter/runtime/kernel-cacd2a85-d5a9-418f-86db-cede2b136617.json']
[D 2021-07-09 01:18:46.316 EnterpriseGatewayApp] Launching kernel: 'Python 3' with command: ['/opt/conda/bin/python', '-m', 'ipykernel_launcher', '-f', '/home/jovyan/.local/share/jupyter/runtime/kernel-f1bb0ba9-9957-4d54-b213-d2987cbe6ee5.json']
[...]

PR #1159, opened 2022-09-21
Update kernel manager library docs

chiawchen:
fix https://github.com/jupyter-server/enterprise_gateway/issues/978 
We need to pass the `kernel_name` as argument explicitly since setting it in the `NotebookClient` constructor time doesn't really propagate down to `GatewayKernelManager`

### DIFF
--- a/docs/source/developers/kernel-manager.md
+++ b/docs/source/developers/kernel-manager.md
@@ -21,8 +21,7 @@ with open("my_notebook.ipynb") as fp:
 gw_client = GatewayClient.instance()
 gw_client.url = "http://my-gateway-server.com:8888"
 
-client = NotebookClient(nb=test_notebook,
-                        kernel_manager_class=GatewayKernelManager)
+client = NotebookClient(nb=test_notebook, kernel_manager_class=GatewayKernelManager)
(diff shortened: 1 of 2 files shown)


docs/source/developers/kernel-manager.md
```diff
@@ -22,9 +22,8 @@ gw_client = GatewayClient.instance()
 gw_client.url = "http://my-gateway-server.com:8888"
 
 client = NotebookClient(nb=test_notebook,
-                        kernel_manager_class=GatewayKernelManager,
-                        kernel_name='my_remote_kernel')
-client.execute()
+                        kernel_manager_class=GatewayKernelManager)
```
kevin-bates, line 25: This is a total nit and nothing you introduced, but, for the sake of consistency, I think it would be good to either join the parameter lines on the constructor here or split the parameters on the `NotebookClient` constructor in the kernel-library.md.  That way, users reading these pages will be able to easily determine that the constructors have the same set of parameters, just different manager class references.  To that point, it may be more clear to join the parameter lines into one.

kevin-bates (approved):
Thank you @chiawchen!

Issue #1160, opened 2022-09-23
user impersonation not setting $HOME or ~ location


blair-anson:
Not sure if this is a bug, or a lack of understanding on my part.

I have user impersonation setup, so these commands all work as expected in the notebook...
```
!whoami
user1
```

```
!pwd
/home/user1
```

```
!id
uid=1001(user1) gid=1001(user1)
```
However these commands still show jovyan as being the user instead of user1...

```
!echo $HOME
/home/jovyan
```

```
!ls ~
group  passwd  work    <<--- this is an ls of /home/jovyan
```  

---
### Configuration ###

**values.yaml**
```
mirrorWorkingDirs: true
```
(so EG_MIRROR_WORKING_DIRS is set)

**JupyterLab Spawner env variables**
```
env['KERNEL_USERNAME'] = self.user.name
env['JUPYTER_GATEWAY_URL'] = "https://xxxxxxxxx:8888"
env['JUPYTER_GATEWAY_REQUEST_TIMEOUT'] = '120'
env['KERNEL_WORKING_DIR'] = f'/home/{self.user.name}'
env['KERNEL_USER_HOME'] = f'/home/{self.user.name}'
user = pwd.getpwnam(self.user.name)
env['KERNEL_UID'] = str(user.pw_uid)
env['KERNEL_GID'] = str(user.pw_gid)
env['USER'] = self.user.nameenv['HOME'] = f'/home/{self.user.name}'
env['KG_HTTP_USER'] = self.user.name

```

kevin-bates:
hi @blair-anson.  We don't currently _flow_ the user's home directory, only the _working_ directory - which all appears to be working.  I don't think we can assume that just because `KERNEL_WORKING_DIR` has a value, that value is the user's HOME directory, so I think we'd need another vehicle to convey this information and notice that you created `KERNEL_USER_HOME`.

Although this would only apply to containerized kernels (docker, k8s), I wonder if we could honor `KERNEL_USER_HOME` such that the container launcher scripts do something like `env['HOME'] = os.getenv('KERNEL_USER_HOME', '/home/jovyan')`?

[...]

PR #1164, opened 2022-09-27
Flow configured envs from client to kernel pod

kevin-bates:
While looking into #1160, it became apparent that the flow of env variables from the client to the kernel pods is not working for `KubernetesProcessProxy`.  Only a fixed set of `KERNEL_`-prefixed variables were transferred to the pod, along with system-defined values for the kernel launcher parameters.  This pull request ensures that envs configured to flow to the kernel pod do flow.

Previously, because k8s relies on its own envs, we unconditionally included all EG-configured envs as well, even though those also never found their way into the kernel pod.  This PR only copies EG envs prefixed with `KUBECONFIG` or `KUBERNETES_SERVICE` rather than the entire EG environment.


kevin-bates:
@lresende, @rahul26goyal - could I please get a review from either of you?

@blair-anson - if you'd like to provide a review (or anyone else) - that would be much appreciated!

Thank you.

kiersten-stokes (approved):
This LGTM! Confirmed that envs prefixed with `KUBECONFIG` or `KUBERNETES_SERVICE` are flowed to the pod (and others are not, unless specified separately in the YAML) by running some notebooks that print out assigned env values. Overwrite of YAML values with current env values is also working as expected.

PR #1166, opened 2022-09-29
CI Cleanup

blink1073:
- Use the link-check action from jupyterlab/maintainer-tools
- Add Python 3.11 to test matrix
- Add minimum version check job - tests a subset since the 2 failing tests gave no meaningful output
- Bump the hatchling version to a post-1.0 version


.github/workflows/build.yml
```diff
@@ -13,7 +13,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        python-version: ["3.7", "3.8", "3.9", "3.10"]
+        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11.0-beta - 3.11.0"]
```
kevin-bates, line 16: I'm curious about the 3.11 version identifier.   I would have expected one or the other but not both separated by " - ".  Is this saying "Prefer 3.11.0, but if not found, use 3.11.0-beta"?

.github/workflows/build.yml
```diff
@@ -89,6 +85,20 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: jupyterlab/maintainer-tools/.github/actions/base-setup@v1
+        with:
+          python_version: "3.7"
+      - uses: jupyterlab/maintainer-tools/.github/actions/install-minimums@v1
+      - name: Run the unit tests
+        run: |
+          pytest -vv -W default enterprise_gateway/tests/test_enterprise_gateway.py || pytest -vv -W default --lf
```
kevin-bates, line 100: From the description...
> Add minimum version check job - tests a subset since the 2 failing tests gave no meaningful output

is the inclusion of `enterprise_gateway/tests/test_enterprise_gateway.py` something we should revisit down the road at some point?

kevin-bates (approved):
This is great @blink1073 - thanks once again!

I did have a couple questions, mostly my own curiosity, that I'm hoping you can answer.  Regardless of those answers, I'm happy moving forward with this.  Thank you.

.github/workflows/build.yml
blink1073: "[hyphen ranges](https://github.com/npm/node-semver#hyphen-ranges-xyz---abc) to download and set up the latest available version of Python (includes both pre-release and stable versions)".  https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#using-the-python-version-input

.github/workflows/build.yml
blink1073: Potentially yes, I was lazy and didn't want to debug the two failing tests.  :smile:

blink1073: Thanks!

PR #1171, opened 2022-10-17
Maintenance cleanup

blink1073:
- Use `Flake8-pyproject` to avoid needing a `.flake8` file
- Use `mdformat` instead of `prettier`
- Add pyproject validation
- Rename `readthedocs` config file to recommended name
- Add manual pre-commit check in CI
- Add rollup check that can be used for branch protection and auto-merge
- Other minor workflow cleanup

PR #1172, opened 2022-10-18
Add mdformat plugins for GitHub support

hukkin:
Hello!

I noticed that https://github.com/jupyter-server/enterprise_gateway/pull/1171 mangled some Markdown tables and frontmatters. This happens because mdformat "core" only formats CommonMark syntax and these extensions are not part of it. This PR fixes the issue.


kevin-bates (approved):
Hi @hukkin - thank you for the contribution.  These changes look good.  I suspect we'll be dropping the website files shortly, but its good to have these plugins in place - thank you.

PR #1180, opened 2022-10-25
Fix version portion of download command to refer to tag name

kevin-bates:
While troubleshooting an issue, I discovered the example `helm install` command was not correct in that the first occurrence of `[VERSION]` in the example should be prefixed with `v` because it represents the tag name.  This pull request fixes the example.


kiersten-stokes (approved):
Good catch!

PR #1187, opened 2022-11-01
Flow kernel-class-name to kernel container if specified in kernel specification

### LINKED ISSUE #1184: Kernelspec option --kernel-class-name does not flow to kernel containers
While working on the [Remote Provisioners](https://github.com/gateway-experiments/remote_provisioners), I found that the `--kernel-class-name` option does not flow from the server to the kernel in containerized environments.  Both the container launcher scripts and the bootstrap script in the kernel containers need to know this option is set.  This only affects python-based containers.

kevin-bates:
This pull request updates the container-based launchers (kubernetes, docker) and the `bootstrap-kernel.sh` script used in the kernel images to recognize and flow `--kernel-class-name` values to the python launcher.

When set, python-based kernel containers will have a `KERNEL_CLASS_NAME` env value corresponding to the configured kernel class that is set in the `kernel.json` via `--kernel-class-name`.

Since our kernelspecs don't set this value, a good way to test is to exec into the EG pod/container and add 
```
    "--kernel-class-name",
    "ipykernel.ipkernel.IPythonKernel"
```
to the `argv` stanza, then confirm the environment values from within the launcher kernel.

Resolves #1184

### DIFF
--- a/etc/kernel-launchers/bootstrap/bootstrap-kernel.sh
+++ b/etc/kernel-launchers/bootstrap/bootstrap-kernel.sh
@@ -5,17 +5,27 @@ RESPONSE_ADDRESS=${RESPONSE_ADDRESS:-${EG_RESPONSE_ADDRESS}}
 PUBLIC_KEY=${PUBLIC_KEY:-${EG_PUBLIC_KEY}}
 KERNEL_LAUNCHERS_DIR=${KERNEL_LAUNCHERS_DIR:-/usr/local/bin/kernel-launchers}
 KERNEL_SPARK_CONTEXT_INIT_MODE=${KERNEL_SPARK_CONTEXT_INIT_MODE:-none}
+KERNEL_CLASS_NAME=${KERNEL_CLASS_NAME}
 
 echo $0 env: `env`
 
 launch_python_kernel() {
-    # Launch the python kernel launcher - which embeds the IPython kernel and listens for interrupts
-    # and shutdown requests from Enterprise Gateway.
+  # Launch the python kernel launcher - which embeds the IPython kernel and listens for interrupts
+  # and shutdown requests from Enterprise Gateway.
 
-    export JPY_PARENT_PID=$$  # Force reset of parent pid since we're detached
+  export JPY_PARENT_PID=$$  # Force reset of parent pid since we're detached
+
+  if [ -z "${KERNEL_CLASS_NAME}" ]
+  then
+    kernel_class_option = ""
+  else
+    kernel_class_option = "--kernel_class_name ${KERNEL_CLASS_NAME}"
+  fi
 
 	set -x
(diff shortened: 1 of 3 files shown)


kiersten-stokes (approved):
Confirmed that class name is flowing as expected when kernel `argv` is updated!

Screen Shot 2022-11-02 at 11 43 06 AM

etc/kernel-launchers/bootstrap/bootstrap-kernel.sh
```diff
@@ -5,17 +5,27 @@ RESPONSE_ADDRESS=${RESPONSE_ADDRESS:-${EG_RESPONSE_ADDRESS}}
...
-    # and shutdown requests from Enterprise Gateway.
+  # Launch the python kernel launcher - which embeds the IPython kernel and listens for interrupts
+  # and shutdown requests from Enterprise Gateway.
+
+  export JPY_PARENT_PID=$$  # Force reset of parent pid since we're detached
 
-    export JPY_PARENT_PID=$$  # Force reset of parent pid since we're detached
+  if [ -z "${KERNEL_CLASS_NAME}" ]
+  then
+    kernel_class_option = ""
```
manoj-jeswani, line 20: Hi @kevin-bates and @kiersten-stokes, on launching the kernel pod (via enterprise-gateway) with `bootstrap-kernel.sh` as the entrypoint for the pod, though the kernel process is getting launched successfully but following error is visible in the pod logs:
`bootstrap-kernel.sh: line 20: kernel_class_option: command not found`

This would have to do with the space present around `=` during value assignment to `kernel_class_option` variable. 
Same goes for line number 22:
`kernel_class_option = "--kernel_class_name ${KERNEL_CLASS_NAME}"`

Can you please verify and confirm the same and I will go ahead and raise a pull-request with the fix. 
Thanks :)

kevin-bates: Hi @manoj-jeswani - thank you for alerting us to this - you are absolutely correct!  I apologize - my criteria for successfully "testing" this was flawed indeed.

Yes, please raise a PR and we'll get this moved forward.

kevin-bates: @manoj-jeswani - I've re-opened #1184.  Please add a "Fixes #1184" or "Resolves #1184" entry in your PR.  Thank you.

manoj-jeswani: Hi @kevin-bates - thank you for your response and confirming on the issue.

After fixing the above-mentioned issue, while testing the end to end flow of `--kernel-class-name` option (by updating `argv` stanza in EG pod/container), I found that kernel pod was failing to launch due to following error:
`launch_ipykernel.py: error: unrecognized arguments: --kernel_class_name`

This had to do with the issue with the value `kernel_class_name` that was being used in line number 22
`kernel_class_option = "--kernel_class_name ${KERNEL_CLASS_NAME}"`

I have fixed this as well and tested the end to end flow of `--kernel-class-name` option. Working fine now.

Please find [here](https://github.com/jupyter-server/enterprise_gateway/pull/1260) the corresponding pull request.
Please let me know in case of any feedback or suggestions. 
Thanks :)

PR #1194, opened 2022-11-09
multiarch image build support

leoyifeng:
1. change JAVA_HOME to be generic
2. use continuumio/miniconda3 as base image for demo-base to remove platform specific package
3. change make file to build multi-arch image using buildx

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -117,10 +117,6 @@ helm-lint: helm-clean
 helm-clean: # Remove any .DS_Store files that might wind up in the package
 	$(shell find etc/kubernetes/helm -type f -name '.DS_Store' -exec rm -f {} \;)
 
-buildx-builder: # start a buildx builder for multi-arc build
-	@echo "starting buildx builder"
-	docker buildx create --use
-
 $(HELM_CHART): $(HELM_CHART_FILES)
 	make helm-lint
 	helm package $(HELM_CHART_DIR) -d dist
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,7 @@
 SA?=source activate
 ENV:=enterprise-gateway-dev
 SHELL:=/bin/bash
+MULTIARCH_BUILD?=
 
 VERSION?=3.1.0.dev0
 SPARK_VERSION?=3.2.1
@@ -116,6 +117,10 @@ helm-lint: helm-clean
 helm-clean: # Remove any .DS_Store files that might wind up in the package
 	$(shell find etc/kubernetes/helm -type f -name '.DS_Store' -exec rm -f {} \;)
 
+buildx-builder: # start a buildx builder for multi-arc build
+	@echo "starting buildx builder"
+	docker buildx create --use
+
 $(HELM_CHART): $(HELM_CHART_FILES)
 	make helm-lint
 	helm package $(HELM_CHART_DIR) -d dist
@@ -151,10 +156,10 @@ kernel-images: ## Build kernel-based docker images
 docker-images: demo-base enterprise-gateway-demo kernel-images enterprise-gateway kernel-py kernel-spark-py kernel-r kernel-spark-r kernel-scala kernel-tf-py kernel-tf-gpu-py kernel-image-puller
 
 enterprise-gateway-demo kernel-images enterprise-gateway kernel-py kernel-spark-py kernel-r kernel-spark-r kernel-scala kernel-tf-py kernel-tf-gpu-py kernel-image-puller:
-	make WHEEL_FILE=$(WHEEL_FILE) VERSION=$(VERSION) NO_CACHE=$(NO_CACHE) TAG=$(TAG) SPARK_VERSION=$(SPARK_VERSION) -C etc $@
+	make WHEEL_FILE=$(WHEEL_FILE) VERSION=$(VERSION) NO_CACHE=$(NO_CACHE) TAG=$(TAG) SPARK_VERSION=$(SPARK_VERSION) MULTIARCH_BUILD=$(MULTIARCH_BUILD) -C etc $@
 
 demo-base:
-	make WHEEL_FILE=$(WHEEL_FILE) VERSION=$(VERSION) NO_CACHE=$(NO_CACHE) TAG=$(SPARK_VERSION) SPARK_VERSION=$(SPARK_VERSION) -C etc $@
+	make WHEEL_FILE=$(WHEEL_FILE) VERSION=$(VERSION) NO_CACHE=$(NO_CACHE) TAG=$(SPARK_VERSION) SPARK_VERSION=$(SPARK_VERSION) MULTIARCH_BUILD=$(MULTIARCH_BUILD) -C etc $@
 
 # Here for doc purposes
 clean-images: clean-demo-base ## Remove docker images (includes kernel-based images)
--- a/etc/Makefile
+++ b/etc/Makefile
@@ -154,7 +154,6 @@ ifdef MULTIARCH_BUILD
 	@echo "cleanup buildx builder $1"
 	-@(docker buildx rm $1)
 	(docker buildx create --use --name $1)
-	@echo "running build on builder instance $(BUILDER_NAME)"
 	(cd ../build/docker/$1; docker buildx build ${NO_CACHE} --platform linux/arm64,linux/amd64 --build-arg HUB_ORG=${HUB_ORG} --build-arg TAG=${TAG} --build-arg SPARK_VERSION=${SPARK_VERSION} -t $(HUB_ORG)/$1:$(TAG) . --push)
 	@echo "remove builder instance $1"
 	-(docker buildx rm $1)
--- a/etc/Makefile
+++ b/etc/Makefile
@@ -104,7 +104,6 @@ kernel_image_files: ../build/kernel_image_files
 #
 
 KERNEL_IMAGES := kernel-py kernel-spark-py kernel-r kernel-spark-r kernel-scala kernel-tf-py kernel-tf-gpu-py
-# DOCKER_IMAGES := demo-base enterprise-gateway-demo enterprise-gateway kernel-image-puller $(KERNEL_IMAGES)
 DOCKER_IMAGES := demo-base enterprise-gateway-demo enterprise-gateway kernel-image-puller $(KERNEL_IMAGES)
 PUSHED_IMAGES := demo-base enterprise-gateway-demo enterprise-gateway kernel-image-puller $(KERNEL_IMAGES)
 
--- a/etc/Makefile
+++ b/etc/Makefile
@@ -154,10 +154,9 @@ ifdef MULTIARCH_BUILD
 	@echo "cleanup buildx builder $1"
 	-@(docker buildx rm $1)
 	(docker buildx create --use --name $1)
-	@$(eval BUILDER_NAME:=`docker buildx create --use`)
 	@echo "running build on builder instance $(BUILDER_NAME)"
 	(cd ../build/docker/$1; docker buildx build ${NO_CACHE} --platform linux/arm64,linux/amd64 --build-arg HUB_ORG=${HUB_ORG} --build-arg TAG=${TAG} --build-arg SPARK_VERSION=${SPARK_VERSION} -t $(HUB_ORG)/$1:$(TAG) . --push)
-	@echo "remove builder instance $(BUILDER_NAME)"
+	@echo "remove builder instance $1"
 	-(docker buildx rm $1)
 else
 	(cd ../build/docker/$1; docker build ${NO_CACHE} --build-arg HUB_ORG=${HUB_ORG} --build-arg TAG=${TAG} --build-arg SPARK_VERSION=${SPARK_VERSION} -t $(HUB_ORG)/$1:$(TAG) .)
--- a/etc/Makefile
+++ b/etc/Makefile
@@ -150,8 +150,7 @@ $1: ../.image-$1
 	@mkdir -p ../build/docker/$1
 	@cp -r docker/$1/* $$(FILES_$1) ../build/docker/$1
 ifdef MULTIARCH_BUILD
-	@echo "starting buildx builder"
-	@echo "cleanup buildx builder $1"
+	@echo "starting buildx builder for $1"
 	-@(docker buildx rm $1)
 	(docker buildx create --use --name $1)
 	(cd ../build/docker/$1; docker buildx build ${NO_CACHE} --platform linux/arm64,linux/amd64 --build-arg HUB_ORG=${HUB_ORG} --build-arg TAG=${TAG} --build-arg SPARK_VERSION=${SPARK_VERSION} -t $(HUB_ORG)/$1:$(TAG) . --push)
--- a/etc/Makefile
+++ b/etc/Makefile
@@ -152,7 +152,7 @@ $1: ../.image-$1
 ifdef MULTIARCH_BUILD
 	@echo "starting buildx builder"
 	(docker buildx create --use)
-	(cd ../build/docker/$1; docker buildx build ${NO_CACHE} --build-arg HUB_ORG=${HUB_ORG} --build-arg TAG=${TAG} --build-arg SPARK_VERSION=${SPARK_VERSION} -t $(HUB_ORG)/$1:$(TAG) . --push)
+	(cd ../build/docker/$1; docker buildx build ${NO_CACHE} --platform linux/arm64,linux/amd64 --build-arg HUB_ORG=${HUB_ORG} --build-arg TAG=${TAG} --build-arg SPARK_VERSION=${SPARK_VERSION} -t $(HUB_ORG)/$1:$(TAG) . --push)
 else
 	(cd ../build/docker/$1; docker build ${NO_CACHE} --build-arg HUB_ORG=${HUB_ORG} --build-arg TAG=${TAG} --build-arg SPARK_VERSION=${SPARK_VERSION} -t $(HUB_ORG)/$1:$(TAG) .)
 endif
--- a/etc/Makefile
+++ b/etc/Makefile
@@ -149,7 +149,13 @@ $1: ../.image-$1
 	@make clean-$1 TARGETS_$1
 	@mkdir -p ../build/docker/$1
 	@cp -r docker/$1/* $$(FILES_$1) ../build/docker/$1
+ifdef MULTIARCH_BUILD
+	@echo "starting buildx builder"
+	(docker buildx create --use)
+	(cd ../build/docker/$1; docker buildx build ${NO_CACHE} --build-arg HUB_ORG=${HUB_ORG} --build-arg TAG=${TAG} --build-arg SPARK_VERSION=${SPARK_VERSION} -t $(HUB_ORG)/$1:$(TAG) . --push)
+else
 	(cd ../build/docker/$1; docker build ${NO_CACHE} --build-arg HUB_ORG=${HUB_ORG} --build-arg TAG=${TAG} --build-arg SPARK_VERSION=${SPARK_VERSION} -t $(HUB_ORG)/$1:$(TAG) .)
+endif
 	@touch ../.image-$1
 	@-docker images $(HUB_ORG)/$1:$(TAG)
 endef
--- a/etc/Makefile
+++ b/etc/Makefile
@@ -104,6 +104,7 @@ kernel_image_files: ../build/kernel_image_files
 #
 
 KERNEL_IMAGES := kernel-py kernel-spark-py kernel-r kernel-spark-r kernel-scala kernel-tf-py kernel-tf-gpu-py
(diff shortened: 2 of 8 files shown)


lresende:
@leoyifeng assuming today we usually do `make docker-images publish-docker-images` how this pr changes this user experience? With this we would need to do `MULTIARCH=true make docker-images` ?

etc/docker/demo-base/Dockerfile
```diff
@@ -1,4 +1,4 @@
-ARG BASE_CONTAINER=ubuntu:bionic
+ARG BASE_CONTAINER=continuumio/miniconda3
```
lresende, line 1: I am assuming you are doing this to simplify the removal of miniconda install which is platform-specific (`Miniconda3-py37_4.8.2-Linux-x86_64.sh`) ?

leoyifeng: > @leoyifeng assuming today we usually do `make docker-images publish-docker-images` how does this pr change this user experience? With this, we would need to do `MULTIARCH=true make docker-images` assuming the multi-arch pushes directly to the docker repository.

yes, with my current implementation, user has to specify the MULTIARCH parameter to trigger the buildx build. Do we want to build multi arch images by default going forward ?

leoyifeng: > I am assuming you are doing this to simplify the removal of miniconda install which is platform-specific (`Miniconda3-py37_4.8.2-Linux-x86_64.sh`) ?

yes

kevin-bates:
I like this - thank @leoyifeng!  

A couple quick questions:

1. I'm not familiar with multi-arch builds but in the examples I see, there is a comma-separated list of platforms specified via `--platform`.  Are all platforms built if that option is not specified?

2. I'd rather not push the images until they are validated and this code would (unconditionally) attempt to push `:dev`-tagged images.  Is the `--push` required for this?

> Do we want to build multi arch images by default going forward?

If they are indeed this painless, then I'm thinking yes.  My understanding is that these are all in the same image and the Docker Engine determines which arch to use via the built-in manifest in conjunction with the runtime architecture.  Do these images tend to be significantly larger than "single-arch" images?  If not, I'd be cool with enabling this by default.

lresende:
I guess based on what @kevin-bates mentioned about the ability to test, this way looks good then. 

So, just validating the user experience (in this case someone that is building the images or a release)

- Build regularly so it can have the images locally and test
- When validated, he has to build it again with 'MULTIARCH=true'? 


@kevin-bates as for your question, as far as I know, the only way to publish multi-arch is to push it directly to the docker repository.

Makefile
```diff
@@ -11,6 +11,7 @@
 SA?=source activate
 ENV:=enterprise-gateway-dev
 SHELL:=/bin/bash
+MULTIARCH_BUILD?=
```
lresende, line 14: Does undefined resolves to false, and thus you don't need to initialize this to false?

lresende (approved):
LGTM, Thank you @leoyifeng

Makefile
leoyifeng: yes

kevin-bates:
> So, just validating the user experience (in this case someone that is building the images or a release)
>- Build regularly so it can have the images locally and test
>- When validated, he has to build it again with 'MULTIARCH=true'?

Are you saying that I cannot locally validate a multi-arch image w/o pushing to some repository first?

>@kevin-bates as for your question, as far as I know, the only way to publish multi-arch is to push it directly to the docker repository.

I thought the act of _pushing_ the image was the same as "publishing".

That said, if I remove the `--push` from the `buildx` statement, the image build completes, but `docker images` does not display the image.  However, if I leave things as-is (with `--push` present) my build times out at that step due to authorization issues.  Even though I can push to `dockerhub` from the CLI, what else is necessary to have the `--push` step work?

@lresende - have you tried running a build of applicable images with this PR?

leoyifeng:
> I like this - thank you @leoyifeng!
> 
> A couple of quick questions:
> 
> 1. I'm not familiar with multi-arch builds but in the examples I see, there is a comma-separated list of platforms specified via `--platform`.  Are all platforms built if that option is not specified?
> 2. I'd rather not push the images until they are validated and this code would (unconditionally) attempt to push `:dev`-tagged images.  Is the `--push` required for this?
> 
> > Do we want to build multi-arch images by default going forward?
> 
> If they are indeed this painless, then I'm thinking yes. My understanding is that these are all in the same image and the Docker Engine determines which arch to use via the built-in manifest in conjunction with the runtime architecture. Do these images tend to be significantly larger than "single-arch" images? If not, I'd be cool with enabling this by default.

for the --platform, good point, I think I miss the option for my PR, you need to specify which option to build with, otherwise my understanding is it's going to build with the platform for the machine. The other part is the base image need to support the platform. I'm testing with linux/amd64,linux/arm64

leoyifeng:
> Do these images tend to be significantly larger than "single-arch" images?

@kevin-bates yes, it will be larger since it actually contains multiple singe image for different platforms

kevin-bates:
yeah, I think I better understand this a bit.  It seems that `buildx` is more of a _context_.  Since I don't see the images, I ran `docker buildx ls` to list the current builds and got this...
```
$ docker buildx ls
NAME/NODE              DRIVER/ENDPOINT  STATUS  PLATFORMS
beautiful_khayyam      docker-container         
  beautiful_khayyam0   rancher-desktop  running linux/amd64, linux/amd64/v2, linux/amd64/v3, linux/arm64, linux/386
distracted_haibt       docker-container         
  distracted_haibt0    rancher-desktop  running linux/amd64, linux/amd64/v2, linux/amd64/v3, linux/arm64, linux/386
epic_antonelli         docker-container         
  epic_antonelli0      rancher-desktop  running linux/amd64, linux/amd64/v2, linux/amd64/v3, linux/arm64, linux/386
hungry_poincare        docker-container         
  hungry_poincare0     rancher-desktop  running linux/amd64, linux/amd64/v2, linux/amd64/v3, linux/arm64, linux/386
interesting_feynman *  docker-container         
  interesting_feynman0 rancher-desktop  running linux/amd64, linux/amd64/v2, linux/amd64/v3, linux/arm64, linux/386
```
As a result, I'm not sure `--platform` is necessary unless these do add significant bloat to the image.  Some of these were previous builds terminated with ctrl-c (like when the auth stuff was timing out).  This seems like it could be problematic.  Is it necessary to `stop` or `rm` the builder instance after each image is built?

Is it the act of "pushing" the image that closes the buildx instance? (I'm guess this is what @lresende may have been getting at).

lresende:
I think we are sort all saying the same, with the caveat that, my understanding is that you can only run the docker image locally that matches your platform, so in this case, if you have a mac with M1 processor, you can't actually build and test a x64 image. But I will let @leoyifeng confirm.

kevin-bates:
> My understanding is that you can only run the docker image locally that matches your platform, so in this case, if you have a mac with M1 processor, you can't actually build and test an x64 image.

The point of the multi-arch builds is to _combine_ the layers of multiple architectures in the _same image_.  So _building_ and _testing_ are two different things.  We should be able to _build_ x64 images on an M1 platform, but, yes, you can't _run_ an x64 image on an M1 platform.

I (now) have multiple concerns:
1. Using the previous platform example, how would one test the M1 portion of a multi-arch image on an M1 platform WITHOUT having to first PUSH that image to some repository?
2. I have accumulated 9 buildx contexts in the course of trying things out.  The builds now fail due to resource (disk) issues, so there's nothing in place to _recover_ from failures and remove the context (via `docker buildx rm <context>`).  I think we will need that as part of this PR.

Since it appears that `--push` IS necessary (I suppose so that the context goes away), I think this kind of target needs to be reserved for actual release builds.  What is concerning about this is whether or not we can trust that a single-arch build (that is validated) behaves identically to that same platform in a multi-arch build.  

I'm not sure if I'm not understanding how this all works, but given that I've already blown out my docker build environment due to resource exhaustion, I'm hoping there are steps that can be taken to make this more natural for developers.

Makefile
```diff
@@ -116,6 +117,10 @@ helm-lint: helm-clean
 helm-clean: # Remove any .DS_Store files that might wind up in the package
 	$(shell find etc/kubernetes/helm -type f -name '.DS_Store' -exec rm -f {} \;)
 
+buildx-builder: # start a buildx builder for multi-arc build
+	@echo "starting buildx builder"
+	docker buildx create --use
```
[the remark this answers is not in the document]
leoyifeng: I was think to add a quick make target for convenience, but agree we don't actually need it.

leoyifeng:
> 1. Using the previous platform example, how would one test the M1 portion of a multi-arch image on an M1 platform WITHOUT having to first PUSH that image to some repository?

if you don't set the MULTIARCH_BUILD, make docker-images will build the image for the platform that machine support, and can verify locally 

> 2. I have accumulated 9 buildx contexts in the course of trying things out.  The builds now fail due to resource (disk) issues, so there's nothing in place to _recover_ from failures and remove the context (via `docker buildx rm <context>`).  I think we will need that as part of this PR.

yes, this is a valid concern, I think we should remove the buildx instance once the build finished. I'll update the PR to include that.

> I'm not sure --platform is necessary unless these do add significant bloat to the image.

I think --platform is necessary, otherwise, buildx will just build the image that works for the build machine.

etc/Makefile
```diff
@@ -149,7 +149,13 @@ $1: ../.image-$1
 	@make clean-$1 TARGETS_$1
 	@mkdir -p ../build/docker/$1
 	@cp -r docker/$1/* $$(FILES_$1) ../build/docker/$1
+ifdef MULTIARCH_BUILD
+	@echo "starting buildx builder"
+	(docker buildx create --use)
+	(cd ../build/docker/$1; docker buildx build ${NO_CACHE} --platform linux/arm64,linux/amd64 --build-arg HUB_ORG=${HUB_ORG} --build-arg TAG=${TAG} --build-arg SPARK_VERSION=${SPARK_VERSION} -t $(HUB_ORG)/$1:$(TAG) . --push)
```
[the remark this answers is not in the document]
leoyifeng: @kevin-bates thx for the suggestion. I was trying to use a random name and remember and remove, but just use the target name is great !!! PR updated.

etc/Makefile
```diff
@@ -149,9 +149,19 @@ $1: ../.image-$1
 	@make clean-$1 TARGETS_$1
 	@mkdir -p ../build/docker/$1
 	@cp -r docker/$1/* $$(FILES_$1) ../build/docker/$1
+ifdef MULTIARCH_BUILD
+	@echo "starting buildx builder"
+	@echo "cleanup buildx builder $1"
```
kevin-bates, line 154: Let's combine these since we're hiding the cleanup anyway...
```suggestion
	@echo "starting buildx builder for $1"
```

kevin-bates (commented):
@leoyifeng - thanks for the updates. This is looking good.

I'd like to make a pass across the images using `MULTIARCH_BUILD=y` which will push `dev` tags and then deploy in my local k8s.

I think I then need to confirm that "regular images" (those produced w/o MULTIARCH_BUILD) build and behave correctly so developers can use these locally w/o having to push anywhere.

This may take a couple of days.  I hope that is okay.

kevin-bates:
Found this interesting regarding local images: https://github.com/docker/buildx/issues/166

I think having the manifest that is produced when multiple platforms are specified is a good thing.  I am wondering, however, if the "default build" should take a single platform name defaulting to the more common arch (no idea what that is) but allowing others to build a different arch as their "default build".  This should produce a local image of that architecture.  I don't think its necessary to do this in this PR, but perhaps something to keep in mind.

kevin-bates:
FWIW - I've pushed multi-arch images for `:dev` to the elyra org in docker hub.  If anyone uses `arm64`, it would be great if you could check them out.  The helm charts in the `main` branch should reference `:dev` (as should those in this PR's branch).  I'll checkout `amd64` tomorrow, then will build "default local" images (single-arch) and check those out.

etc/docker/kernel-spark-py/Dockerfile
```diff
@@ -23,7 +23,9 @@ RUN apt-get update && \
     libssl-dev \
     && rm -rf /var/lib/apt/lists/*
 
-ENV JAVA_HOME /usr/lib/jvm/java-1.8.0-openjdk-amd64
+RUN mkdir -p /usr/lib/jvm/java
```
kevin-bates, line 26: I'm seeing the following issues with this image and `kernel-spark-r` (from their pod logs):
```
+ CMD=("$SPARK_HOME/bin/spark-submit" --conf "spark.driver.bindAddress=$SPARK_DRIVER_BIND_ADDRESS" --deploy-mode client "$@")
+ exec /usr/bin/tini -g -- /opt/spark/bin/spark-submit --conf spark.driver.bindAddress=10.42.0.247 --deploy-mode client --properties-file /opt/spark/conf/spark.properties --class org.apache.spark.deploy.PythonRunner local:///usr/local/bin/kernel-launchers/python/scripts/launch_ipykernel.py --RemoteProcessProxy.kernel-id 2b027ae7-309f-4421-9fba-14c284e9d328 --RemoteProcessProxy.port-range 0..0 --RemoteProcessProxy.response-address 10.42.0.235:8877 --RemoteProcessProxy.public-key MIGfMA0GCSqGSIb3DQEBAQUAA4GNADCBiQKBgQCa/SIFcGS1qe/LiszCrcpg+IEmS9P0cml/7HBZZuqqARQFXSo74r6MzUCCFDgao5x9fESoVcGZWWraNY8kvGcRe5IJo3Y7rsQ/PVosQloxNcfQtUHZn4F6kEHycr8QwoM0TJW0H9TKttAT/WZ7u0UAIy8YrDoOCe4XYN5Cl79keQIDAQAB --RemoteProcessProxy.spark-context-initialization-mode lazy
/opt/spark/bin/spark-class: line 71: /usr/lib/jvm/java/bin/java: No such file or directory
/opt/spark/bin/spark-class: line 96: CMD: bad array subscript
```

Looking within the images I see:
```
jovyan@c7778786e3ee:/usr/lib/jvm$ ls -l
total 8
drwxr-xr-x 1 root root 4096 Nov 10 23:25 java
lrwxrwxrwx 1 root root   20 Oct 25 17:26 java-1.8.0-openjdk-amd64 -> java-8-openjdk-amd64
drwxr-xr-x 7 root root 4096 Nov 10 23:25 java-8-openjdk-amd64
jovyan@c7778786e3ee:/usr/lib/jvm$ ls java
java-8-openjdk-amd64
jovyan@c7778786e3ee:/usr/lib/jvm$ ls java/java-8-openjdk-amd64
ASSEMBLY_EXCEPTION  bin  docs  include  jre  lib  man  src.zip  THIRD_PARTY_README
```
and wonder if this `mkdir` should be something like:
```suggestion
RUN ln -s /usr/lib/jvm/java-8-openjdk-amd64 /usr/lib/jvm/java
```

kevin-bates: ok, I think I see what this statement is about: `RUN ln -s $(readlink -f /usr/bin/javac | sed "s:/bin/javac::") ${JAVA_HOME}`

It essentially provides a means of linking the arch-specific directory to a non-arch name and not need to know the architecture type involved in that.  As a result, would simply _removing_ the `mkdir` statement be sufficient?

kevin-bates: I've confirmed that only removing the `mkdir` statement is sufficient.
```suggestion
```

etc/docker/kernel-spark-r/Dockerfile
```diff
@@ -18,7 +18,9 @@ RUN apt-get update && apt-get install -y \
     openjdk-8-jdk \
     libssl-dev
 
-ENV JAVA_HOME /usr/lib/jvm/java-1.8.0-openjdk-amd64
+RUN mkdir -p /usr/lib/jvm/java
```
kevin-bates, line 21: Per https://github.com/jupyter-server/enterprise_gateway/pull/1194/files#r1020484617
```suggestion
```

etc/docker/enterprise-gateway/Dockerfile
```diff
@@ -22,6 +22,10 @@ USER root
 
 RUN apt update && apt install -yq curl openjdk-8-jdk
 
+RUN mkdir -p /usr/lib/jvm/java
```
kevin-bates, line 25: Per https://github.com/jupyter-server/enterprise_gateway/pull/1194/files#r1020484617
```suggestion
```

etc/docker/demo-base/Dockerfile
```diff
@@ -49,11 +49,16 @@ RUN apt-get update && apt-get -yq dist-upgrade \
...
     libxrender1 \
-    openjdk-8-jdk \
+    software-properties-common \
     openssh-server \
     openssh-client \
+ && apt-add-repository 'deb http://security.debian.org/debian-security stretch/updates main' \
+ && apt-get update && apt-get install -yq --no-install-recommends openjdk-8-jdk \
  && rm -rf /var/lib/apt/lists/*
 
+RUN mkdir -p /usr/lib/java
```
kevin-bates, line 59: Per https://github.com/jupyter-server/enterprise_gateway/pull/1194/files#r1020484617 this can be removed.  It's actually a bug anyway (should be `/usr/lib/jvm/java`) - which is why this particular image is working...
```suggestion
```

leoyifeng:
thx @kevin-bates pr updated. but just trying to understand why remove `RUN mkdir -p /usr/lib/jvm/java` would fix the issue ?

kevin-bates:
> why remove RUN mkdir -p /usr/lib/jvm/java would fix the issue ?

Per my previous image captures, when `/usr/lib/jvm/java` exists it messes up the linking that happens in the next step, resulting the link created in `/usr/lib/jvm/java` rather than `java` being the link....
```
jovyan@c7778786e3ee:/usr/lib/jvm$ ls java
java-8-openjdk-amd64
jovyan@c7778786e3ee:/usr/lib/jvm$ ls java/java-8-openjdk-amd64
ASSEMBLY_EXCEPTION  bin  docs  include  jre  lib  man  src.zip  THIRD_PARTY_README
```

You can check this yourself by looking at the multiarch image:
```
$ docker run elyra/kernel-spark-py:dev ls -l /usr/lib/jvm/java
...
+ exec /usr/bin/tini -g -- ls -l /usr/lib/jvm/java
Non-spark-on-k8s command provided, proceeding in pass-through mode...
lrwxrwxrwx 1 root root 33 Nov 11 18:38 /usr/lib/jvm/java -> /usr/lib/jvm/java-8-openjdk-amd64
```
Then, looking _inside_ the directory (i.e., JAVA_HOME), you actually see the expected contents, not the other directory as before...
```
$ docker run elyra/kernel-spark-py:dev ls -l /usr/lib/jvm/java/
...
+ exec /usr/bin/tini -g -- ls -l /usr/lib/jvm/java/
total 20
lrwxrwxrwx 1 root root   22 Oct 25 17:26 ASSEMBLY_EXCEPTION -> jre/ASSEMBLY_EXCEPTION
drwxr-xr-x 2 root root 4096 Nov 11 18:38 bin
lrwxrwxrwx 1 root root   41 Oct 25 17:26 docs -> ../../../share/doc/openjdk-8-jre-headless
drwxr-xr-x 3 root root 4096 Nov 11 18:38 include
drwxr-xr-x 5 root root 4096 Nov 11 18:38 jre
drwxr-xr-x 3 root root 4096 Nov 11 18:38 lib
drwxr-xr-x 4 root root 4096 Nov 11 18:38 man
lrwxrwxrwx 1 root root   20 Oct 25 17:26 src.zip -> ../openjdk-8/src.zip
lrwxrwxrwx 1 root root   22 Oct 25 17:26 THIRD_PARTY_README -> jre/THIRD_PARTY_README
```

kevin-bates:
Thanks @leoyifeng.

We should talk about one final item I can see...

Since we build the images during `release.sh` with the `--dryRun` (which is fine IMO) should we update the [`release.sh` script](https://github.com/jupyter-server/enterprise_gateway/blob/746f8b340dc170b2ec2638b0d146dee8b39d1f1b/release.sh#L246-L250) to add the `MULTIARCH_BUILD=y` flag in both cases?  (I guess we really don't need the `push-images` tag since `--dryRun` will push by virtue of `MULTIARCH_BUILD`, but I'm okay leaving it for now.)

@lresende - what do you think?  I think this needs to happen either now or before we build 3.0.1 so might as well add it now.

kevin-bates:
I went ahead and pushed a commit to update the release script.

@lresende - could you please take a look?

lresende (approved):

lresende:
Just FYI: For building the multiarch builds, it seems that resources on your docker desktop have direct influence on your success and speed of the build. Regular resources like 2 or 3 cores with 4 GB ram make the build takes forever. Increasing that to 7 cores, 20 gb and 4 GB of swap made `demo + demo-base` and `eg + kernels` to be built in parallel and published much quicker...

BTW, I built + publish 3.1.0 in this experience

@kevin-bates @leoyifeng

PR #1199, opened 2022-11-17
Support Kubernetes scheduling constraints ( tolerations, nodeselector and affinity ) and add ingress class name

luong-komorebi:
Documentations:
1. https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/ for   `( tolerations, nodeselector and affinity )`
2. https://kubernetes.io/docs/concepts/services-networking/ingress/ for ingress class name `ingressClassName`

Real applications:
1. I want some machine level isolation between the main jupyter enterprise services with other spawned kernels. That cannot be ensured without some kubernetes specific constraints specification.  With this separation, I can provide dedicated resources for the gateway such that no other running kernels can compete with it. Plus, there might be other applications like running the gateway on ARM machine (to save money) and running the kernels on AMD x64 machines (for maximum compatability with libraries)....
2. For the ingress class name, it is the standard since K8s version 1.18. Without which we would have trouble using third party ingress providers / controllers like [Amazon ALB controllers](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.4/guide/ingress/ingress_class/#deprecated-kubernetesioingressclass-annotation), [nginx ingress controller ](https://kubernetes.github.io/ingress-nginx/user-guide/basic-usage/#:~:text=ingress%2Dmyservicea%0Aspec%3A-,ingressClassName,-%3A%20nginx%0A%20%20rules%3A%0A%20%20%2D%20host)
3. Minor update to how kernel whitelist is reference in docs

This is my first contribution, so let me know if I am not following the right convention.  I just loosely followed this guideline https://docs.jupyter.org/en/latest/contributing/content-contributor.html 
Signed-off-by: Luong Vo <vo.tran.thanh.luong@gmail.com>


etc/kubernetes/helm/enterprise-gateway/templates/ingress.yaml
```diff
@@ -16,6 +16,9 @@ metadata:
 {{ toYaml .Values.ingress.annotations | indent 4}}
 {{- end }}
 spec:
+{{ if .Values.ingress.ingressClassName }}
```
kevin-bates, line 19: Should we define an empty placeholder, similar to [`kernelspecs.image`](https://github.com/jupyter-server/enterprise_gateway/blob/2f9d635f9d9d85ae386b40f7b222b45ca580f978/etc/kubernetes/helm/enterprise-gateway/values.yaml#L96), so that folks have an idea that this is configurable?

kevin-bates (commented):
Hi @luong-komorebi - this PR is great.  So simple, yet extremely useful!  Also, thank you for the doc fix ups - they are much appreciated.

I just have the one comment about adding a placeholder for `ingressClassName`.

etc/kubernetes/helm/enterprise-gateway/templates/ingress.yaml
kevin-bates: > Would you regards a YAML schema important, so that each field and its types/ values can become self-documented and easily validated ?

That sounds great! I'd like us to leverage schemas/correctness as much as possible.

kevin-bates (approved):
LGTM - thank you @luong-komorebi

PR #1202, opened 2022-11-18
Handle the case when pod and/or pvc has already been created

luong-komorebi:
I am not entirely sure this is the best way to go, as well as I am aware we should keep the original script minimal to encourage downstream customizations. 

However, I find it confusing when out of the box, with the default kernel-pod.yaml.j2 (and kernel_volume enabled), we cannot have idempotency and jupyter EG will just return error because given the same kernel id, the same pod and pvc cannot be connected from the client side. 

This pull request shows my little adjustments to improve that. And while of course it can be applied to other resources, I would propose just keep it as minimal as possible and PVC as well as pod proves to be great targets (they can also be cleaned up by terminate_container_resources inside k8s.py process proxy)

### DIFF
--- a/etc/kernel-launchers/kubernetes/scripts/launch_kubernetes.py
+++ b/etc/kernel-launchers/kubernetes/scripts/launch_kubernetes.py
@@ -158,10 +158,15 @@ def launch_kubernetes_kernel(
                         )
                     except ApiException as exc:
                         if _parse_k8s_exception(exc) == K8S_ALREADY_EXIST_REASON:
-                            pod_created = client.CoreV1Api(client.ApiClient()).list_namespaced_pod(
-                                namespace=kernel_namespace,
-                                label_selector="kernel_id={}".format(kernel_id), watch=False
-                            ).items[0]
+                            pod_created = (
+                                client.CoreV1Api(client.ApiClient())
+                                .list_namespaced_pod(
+                                    namespace=kernel_namespace,
+                                    label_selector=f"kernel_id={kernel_id}",
+                                    watch=False,
+                                )
+                                .items[0]
+                            )
                         else:
                             raise exc
             elif k8s_obj["kind"] == "Secret":
@@ -184,8 +189,7 @@ def launch_kubernetes_kernel(
                             raise exc
             elif k8s_obj["kind"] == "PersistentVolume":
                 if pod_template_file is None:
-                    client.CoreV1Api(client.ApiClient()
-                                     ).create_persistent_volume(body=k8s_obj)
+                    client.CoreV1Api(client.ApiClient()).create_persistent_volume(body=k8s_obj)
             elif k8s_obj["kind"] == "Service":
                 if pod_template_file is None:
                     if pod_created is not None:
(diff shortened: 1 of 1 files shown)


etc/kernel-launchers/kubernetes/scripts/launch_kubernetes.py
```diff
@@ -131,23 +152,44 @@ def launch_kubernetes_kernel(
...
                 if pod_template_file is None:
-                    pod_created = client.CoreV1Api(client.ApiClient()).create_namespaced_pod(
-                        body=k8s_obj, namespace=kernel_namespace
-                    )
+                    try:
+                        pod_created = client.CoreV1Api(client.ApiClient()).create_namespaced_pod(
+                            body=k8s_obj, namespace=kernel_namespace
+                        )
+                    except ApiException as exc:
+                        if _parse_k8s_exception()(exc) == K8S_ALREADY_EXIST_REASON:
```
kevin-bates, line 160: I don't understand this syntax.  How are the _two_ sets of parenthesis handled?  I expected: `_parse_k8s_exception(exc)`.   

(I'm not up on all of the various python features.  When I tried searching for this syntax it appears to be used when embedded functions are involved, but that doesn't appear to be the case here so I'm curious.)

kevin-bates:
> I find it confusing when out of the box, with the default kernel-pod.yaml.j2 (and kernel_volume enabled), we cannot have idempotency and jupyter EG will just return error because given the same kernel id, the same pod and pvc cannot be connected from the client side.

I totally agree with this approach and am surprised this hasn't been brought up before. (I don't use volumes in my day-to-day usage, but probably should!). FWIW, we do trap `409` when attempting to create the kernel pod's namespace, but [tolerate its existence when the kernel is restarting](https://github.com/jupyter-server/enterprise_gateway/blob/15bcc596ecb630d2aacaf4bb492ec1428cc3cc06/enterprise_gateway/services/processproxies/k8s.py#L279).

Thank you for another great contribution!

etc/kernel-launchers/kubernetes/scripts/launch_kubernetes.py
kevin-bates: Cool - glad it wasn't some other aspect of Python that I need to become familiar with! :smile:

kevin-bates:
> Another point for future improvement if I have the time is to write some simple unit test for this script

Yes, that's another thing we're lacking.

Btw, I wanted to bring [this repo](https://github.com/gateway-experiments/remote_provisioners) to your attention.  [Kernel Provisioners](https://jupyter-client.readthedocs.io/en/latest/provisioning.html) essentially replace Process Proxies and are baked into the general Jupyter framework. They are nearly identical to process proxies (process proxies were essentially a proof of concept of kernel provisioners) as you can notice.  In addition, the launch scripts are identical (except any environment variables are prefixed differently).  EG 4.0 will have its process proxies and most all of its `etc` hierarchy removed and will, instead, leverage Remote Provisioners.

I bring this up because most items added to this repository, particularly anything in the `services/processproxies` or `etc` hierarchies also apply to https://github.com/gateway-experiments/remote_provisioners.  So, if you're interested in applying similar changes/contributions there (until EG 4.0 exists), it would be greatly appreciated.  If not, a maintainer will migrate applicable changes and add you (or whoever made the contribution) as a `co-author`.  FYI.

kevin-bates (approved):
LGTM - thank you again @luong-komorebi!

PR #1210, opened 2022-12-05
Bump spark version and use apache cdn to improve download time for Spark

luong-komorebi:
This PR proposes bumping Spark version to the latest 3.2.x release. The reason for the bump is to improve download time for Spark, which in return would reduce build time for docker. This will be helpful for someone who needs docker rebuild to work with other systems as fast as possible ( Kubernetes, DockerSwarm ). 

The latest version that the CDN supports is 3.2.3 and it is notable that for my location (Southeast Asia), using the current apache.org link makes it impossible to build the docker image under 4 hours. 

Plus, on the way to change the download link file, I did run the `itest-docker` test locally again. HAProxy is not running so I also fixed that. The tests will never succeed if the kernels images are not present so they are preloaded before the test run in `itest-docker-prep`

Running itest cases gives:
```bash
>       self.assertRegex(interrupted_result, "java.lang.InterruptedException")
E       AssertionError: Regex didn't match: 'java.lang.InterruptedException' not found in 'begin\nend\n'

enterprise_gateway/itests/test_scala_kernel.py:65: AssertionError
=============================================================== slowest 10 durations ================================================================
62.15s call     enterprise_gateway/itests/test_scala_kernel.py::TestScalaKernelLocal::test_interrupt
13.75s call     enterprise_gateway/itests/test_scala_kernel.py::TestScalaKernelLocal::test_restart
11.38s setup    enterprise_gateway/itests/test_scala_kernel.py::TestScalaKernelLocal::test_get_hostname
5.80s call     enterprise_gateway/itests/test_r_kernel.py::TestRKernelLocal::test_restart
5.59s call     enterprise_gateway/itests/test_python_kernel.py::TestPythonKernelLocal::test_restart
5.37s setup    enterprise_gateway/itests/test_r_kernel.py::TestRKernelLocal::test_get_hostname
4.00s call     enterprise_gateway/itests/test_python_kernel.py::TestPythonKernelLocal::test_interrupt
3.76s setup    enterprise_gateway/itests/test_python_kernel.py::TestPythonKernelLocal::test_get_hostname
3.67s call     enterprise_gateway/itests/test_r_kernel.py::TestRKernelLocal::test_interrupt
1.00s call     enterprise_gateway/itests/test_scala_kernel.py::TestScalaKernelLocal::test_get_hostname
============================================================== short test summary info ==============================================================
FAILED enterprise_gateway/itests/test_scala_kernel.py::TestScalaKernelLocal::test_interrupt - AssertionError: Regex didn't match: 'java.lang.InterruptedException' not found in 'begin\nend\n'
===================================================== 1 failed, 12 passed in 120.50s (0:02:00) ======================================================
``` 

I am not really familiar with scala kernel, and it seems to not be interrupted. I am stuck here currently. 
If this idea is worth it, maybe I will start over and try to debug the scala integration test. 

Reference:
1. Apache foundation is moving from mirrors to cdn: https://news.apache.org/foundation/entry/apache-software-foundation-moves-to
2. Linking directly to apache.org is discouraged: https://infra.apache.org/release-download-pages.html#:~:text=Do%20not%20link%20directly%20to%20dist.apache.org
3. Version3.2.1 is no longer found in the mirror if using cdn https://www.apache.org/dyn/closer.lua/spark/spark-3.2.1/spark-3.2.1-bin-hadoop2.7.tgz

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -241,6 +241,7 @@ endif
 
 PREP_TIMEOUT?=180
 itest-docker-prep:
+	@-echo elyra/kernel-py:$(TAG) elyra/kernel-r:$(TAG) elyra/kernel-scala:$(TAG) | xargs -P3 -n1 docker pull
 	@-docker service rm enterprise-gateway_enterprise-gateway enterprise-gateway_enterprise-gateway-proxy
 	@-docker swarm leave --force
 	# Check if swarm mode is active, if not attempt to create the swarm


kevin-bates (commented):
Hi @luong-komorebi - these changes look good.  Just the one question about the location of images since it would be good to be able to use local-only images for test purposes.

Regarding the interrupts on Scala, EG actually uses `SIGUSR2` which is "delivered" to the scala launcher for containerized, non-spark, instances via the [bootstrap-kernel.sh](https://github.com/jupyter-server/enterprise_gateway/blob/c6a2c766e16a34e9723011e50d982064119daa44/etc/kernel-launchers/bootstrap/bootstrap-kernel.sh#L65) script.  The Toree-launcher then sends `SIGUSR2` to toree - which then treats it as a `SIGINT`.  This is necessary because sending `SIGINT` across JVM instances is not allowed.

Since interrupts on scala-based kernels works on Kubernetes, I would expect this to work in docker, so there might be some disconnect happening somewhere.  I will try to look into this, but I don't feel this particular issue should hold up this PR.

etc/docker/demo-base/Dockerfile
```diff
@@ -90,7 +90,7 @@ RUN mkdir /home/$NB_USER/work && \
 
 # DOWNLOAD HADOOP AND SPARK
 RUN curl -sL https://archive.apache.org/dist/hadoop/common/hadoop-$HADOOP_VER/hadoop-$HADOOP_VER.tar.gz | tar -xz -C /usr/hdp/current
-RUN curl -sL https://archive.apache.org/dist/spark/spark-$SPARK_VER/spark-$SPARK_VER-bin-hadoop3.2.tgz | tar -xz -C /usr/hdp/current
+RUN curl -sL https://dlcdn.apache.org/spark/spark-$SPARK_VER/spark-$SPARK_VER-bin-hadoop3.2.tgz | tar -xz -C /usr/hdp/current
```
lresende, line 93: FYI -> we had this in the past, the issue is that we will then always have build breakages when the images move away...

lresende: Never mind, looks like apache itself has a new cdn...

Issue #1216, opened 2022-12-08
3.1.0 - Python and R Spark kernel images won't start


kevin-bates:
The `elyra/kernel-spark-py:3.1.0` and `elyra/kernel-spark-r:3.1.0` images fail to start due to the following:
```
$ docker run -it --rm elyra/kernel-spark-py:3.1.0 bash
++ id -u
+ myuid=1000
++ id -g
+ mygid=100
+ set +e
++ getent passwd 1000
+ uidentry=jovyan:x:1000:100::/home/jovyan:/bin/bash
+ set -e
+ '[' -z jovyan:x:1000:100::/home/jovyan:/bin/bash ']'
+ SPARK_CLASSPATH=':/opt/spark/jars/*'
+ env
+ sort -t_ -k4 -n
+ sed 's/[^=]*=\(.*\)/\1/g'
+ grep SPARK_JAVA_OPT_
+ readarray -t SPARK_EXECUTOR_JAVA_OPTS
+ '[' -n '' ']'
+ '[' -z ']'
+ '[' -z ']'
+ '[' -n '' ']'
+ '[' -z ']'
+ '[' -z ']'
+ '[' -z x ']'
+ SPARK_CLASSPATH='/opt/spark/conf::/opt/spark/jars/*'
+ case "$1" in
+ echo 'Non-spark-on-k8s command provided, proceeding in pass-through mode...'
Non-spark-on-k8s command provided, proceeding in pass-through mode...
+ CMD=("$@")
+ exec /usr/bin/tini -g -- bash
/opt/entrypoint.sh: line 105: /usr/bin/tini: No such file or directory
```

- Operating System and version:
- Browser and version:
- Jupyter Server version:

<details><summary>Troubleshoot Output</summary>
<pre>
Paste the output from running `jupyter troubleshoot` from the command line here.
You may want to sanitize the paths in the output.
</pre>
</details>

<details><summary>Command Line Output</summary>
<pre>
Paste the output from your command line running `jupyter lab` here, use `--debug` if possible.
</pre>
</details>

<details><summary>Browser Output</summary>

<pre>
Paste the output from your browser Javascript console here, if applicable.

</pre>
</details>

kevin-bates:
Here's output showing these images now start:
```
$  docker run -it --rm elyra/kernel-spark-py:dev whoami
++ id -u
+ myuid=1000
++ id -g
+ mygid=100
+ set +e
++ getent passwd 1000
[...]

PR #1217, opened 2022-12-09
Fix kernel-spark-py and kernel-spark-r images

kevin-bates:
Apparently, the "last minute" change made in #1207 for 3.1.0 came back to bite me!  Those base images do not include `/opt/conda/bin/tini` which we preferred over the default `/usr/bin/tini`.  As a result, the adjustments we make in our spark images that derive from the images built from those _docker-stacks_ images were DOA.  (What could possibly go wrong, right? :smile:)

This pull request removes the use of `/opt/conda/bin/tini` and merely performs the previous adjustment of `-s` to `-g` relative to `/usr/bin/tini`.

Fixes: #1216

### DIFF
--- a/etc/docker/kernel-spark-py/Dockerfile
+++ b/etc/docker/kernel-spark-py/Dockerfile
@@ -32,12 +32,10 @@ RUN curl -s https://archive.apache.org/dist/spark/spark-${SPARK_VER}/spark-${SPA
     ln -s ${SPARK_HOME}-${SPARK_VER}-bin-hadoop2.7 $SPARK_HOME
 
 # Download entrypoint.sh from matching tag
-# Use tini from Anaconda installation
 RUN cd /opt/ && \
     wget https://raw.githubusercontent.com/apache/spark/v${SPARK_VER}/resource-managers/kubernetes/docker/src/main/dockerfiles/spark/entrypoint.sh && \
     chmod a+x /opt/entrypoint.sh && \
-    sed -i 's/tini -s/tini -g/g' /opt/entrypoint.sh && \
-    ln -sfn /opt/conda/bin/tini /usr/bin/tini
+    sed -i 's/tini -s/tini -g/g' /opt/entrypoint.sh
 
 WORKDIR $SPARK_HOME/work-dir
 # Ensure that work-dir is writable by everyone
--- a/etc/docker/kernel-spark-r/Dockerfile
+++ b/etc/docker/kernel-spark-r/Dockerfile
@@ -27,12 +27,10 @@ RUN curl -s https://archive.apache.org/dist/spark/spark-${SPARK_VER}/spark-${SPA
     ln -s ${SPARK_HOME}-${SPARK_VER}-bin-hadoop2.7 $SPARK_HOME
 
 # Download entrypoint.sh from matching tag
-# Use tini from Anaconda installation
 RUN cd /opt/ && \
     wget https://raw.githubusercontent.com/apache/spark/v${SPARK_VER}/resource-managers/kubernetes/docker/src/main/dockerfiles/spark/entrypoint.sh && \
     chmod a+x /opt/entrypoint.sh && \
-    sed -i 's/tini -s/tini -g/g' /opt/entrypoint.sh && \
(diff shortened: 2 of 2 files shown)


kevin-bates:
Hi @luong-komorebi - thank you for taking a look at this and the great insight regarding `tini` and `HEALTHCHECK`!

I first noticed the latter when I implemented the [PR](https://github.com/jupyter/docker-stacks/pull/1825) to introduce a base image (`docker-stacks-foundation`) which splits the base notebook image into a further base image from which the notebook base image now derives.  Although I didn't update the kernel images in EG, this `docker-stacks-foundation` image serves as the base image for the [kernel-based images](https://github.com/gateway-experiments/remote_provisioners/blob/f927b6e897de90b2914aa347009cf224b2ea64e5/remote_provisioners/docker/kernel-image/Dockerfile#L5) that we either produce or document in [remote_provisioners](https://github.com/gateway-experiments/remote_provisioners).

Since our kernel images in EG derive from "server-based" images, you bring up a good point about disabling the `HEALTHCHECK` in them and I've gone ahead and created issue #1218.  I suspect the `HEALTHCHECK` is still valid for the EG image itself, but we should probably visit that topic when resolving that issue.

This topic also reminds me to introduce a new label named `remote-provisioners` which indicates that the EG issue or PR would also apply to [remote_provisioners](https://github.com/gateway-experiments/remote_provisioners) (and likely have an analogous `enterprise-gateway` label in remote_provisioners).

I think we can go ahead with this PR, assuming it checks out with others, but let's not cut the next (patch?) release until we've added `HEALTHCHECK NONE` to our kernel images.  Thank you for raising this.

kevin-bates:
Here's output showing these images now start after these changes:
```
$  docker run -it --rm elyra/kernel-spark-py:dev whoami
++ id -u
+ myuid=1000
++ id -g
+ mygid=100
+ set +e
++ getent passwd 1000
+ uidentry=jovyan:x:1000:100::/home/jovyan:/bin/bash
+ set -e
+ '[' -z jovyan:x:1000:100::/home/jovyan:/bin/bash ']'
+ SPARK_CLASSPATH=':/opt/spark/jars/*'
+ env
+ grep SPARK_JAVA_OPT_
+ sort -t_ -k4 -n
+ sed 's/[^=]*=\(.*\)/\1/g'
+ readarray -t SPARK_EXECUTOR_JAVA_OPTS
+ '[' -n '' ']'
+ '[' -z ']'
+ '[' -z ']'
+ '[' -n '' ']'
+ '[' -z ']'
+ '[' -z ']'
+ '[' -z x ']'
+ SPARK_CLASSPATH='/opt/spark/conf::/opt/spark/jars/*'
+ case "$1" in
+ echo 'Non-spark-on-k8s command provided, proceeding in pass-through mode...'
Non-spark-on-k8s command provided, proceeding in pass-through mode...
+ CMD=("$@")
+ exec /usr/bin/tini -g -- whoami
jovyan
```
```
$  docker run -it --rm elyra/kernel-spark-r:dev whoami
++ id -u
+ myuid=1000
++ id -g
+ mygid=100
+ set +e
++ getent passwd 1000
+ uidentry=jovyan:x:1000:100::/home/jovyan:/bin/bash
+ set -e
+ '[' -z jovyan:x:1000:100::/home/jovyan:/bin/bash ']'
+ SPARK_CLASSPATH=':/opt/spark/jars/*'
+ env
+ grep SPARK_JAVA_OPT_
+ sort -t_ -k4 -n
+ sed 's/[^=]*=\(.*\)/\1/g'
+ readarray -t SPARK_EXECUTOR_JAVA_OPTS
+ '[' -n '' ']'
+ '[' -z ']'
+ '[' -z ']'
+ '[' -n '' ']'
+ '[' -z ']'
+ '[' -z ']'
+ '[' -z x ']'
+ SPARK_CLASSPATH='/opt/spark/conf::/opt/spark/jars/*'
+ case "$1" in
+ echo 'Non-spark-on-k8s command provided, proceeding in pass-through mode...'
Non-spark-on-k8s command provided, proceeding in pass-through mode...
+ CMD=("$@")
+ exec /usr/bin/tini -g -- whoami
jovyan
```

lresende (approved):

PR #1223, opened 2022-12-15
Avoid Spark CRD rejection due to empty env var value

### LINKED ISSUE #1222: Kernel CRD Application rejected due to empty env vars value
## Description

When a declared env variable on the kernel CRD has no value, spark operator are rejecting applications:
```
HTTP response body: 
{
	"kind": "Status",
	"apiVersion": "v1",
	"metadata": {},
	"status": "Failure",
	"message": "SparkApplication.sparkoperator.k8s.io \"lresende-6975d2e5-bd58-4ac0-afdb-d1bbacbb7166\" is invalid: spec.driver.env.value: Invalid value: \"null\": spec.driver.env.value in body must be of type string: \"null\"",
	"reason": "Invalid",
	"details": {
		"name": "lresende-6975d2e5-bd58-4ac0-afdb-d1bbacbb7166",
		"group": "sparkoperator.k8s.io",
		"kind": "SparkApplication",
		"causes": [{
			"reason": "FieldValueInvalid",
			"message": "Invalid value: \"null\": spec.driver.env.value in body must be of type string: \"null\"",
			"field": "spec.driver.env.value"
		}]
	},
	"code": 422
}
```

lresende:
Fixes #1222 

When submitting the Kernel application CRD and processing the environment variables, if there is no value uses a String "None" to avoid Spark Operator rejecting the application with `spec.driver.env.value in body must be of type string`

### DIFF
--- a/etc/kernel-launchers/operators/scripts/launch_custom_resource.py
+++ b/etc/kernel-launchers/operators/scripts/launch_custom_resource.py
@@ -48,7 +48,12 @@ def launch_custom_resource_kernel(
 
     for name, value in os.environ.items():
         if name.startswith("KERNEL_"):
-            keywords[name.lower()] = yaml.safe_load(value)
+            if value:
+                keywords[name.lower()] = yaml.safe_load(value)
+            else:
+                # When there is no value, send a None string
+                # to avoid Spark Operator to reject the application submission
+                keywords[name.lower()] = yaml.safe_load("None")
 
     kernel_crd_template = keywords["kernel_crd_group"] + "-" + keywords["kernel_crd_version"]
     custom_resource_yaml = generate_kernel_custom_resource_yaml(kernel_crd_template, keywords)


etc/kernel-launchers/operators/scripts/launch_custom_resource.py
```diff
@@ -48,7 +48,12 @@ def launch_custom_resource_kernel(
 
     for name, value in os.environ.items():
         if name.startswith("KERNEL_"):
-            keywords[name.lower()] = yaml.safe_load(value)
+            if value:
+                keywords[name.lower()] = yaml.safe_load(value)
+            else:
```
kevin-bates, line 53: Would it be better to _skip_ the env altogether?  It might be good to issue a DEBUG log indicating that variable `foo` was skipped due to not having a value, but I think some applications might treat `"None"` as a value (and possibly invalid at that) so it would be better not to include at all.

I'm curious which `KERNEL_` envs don't have values.  That seems like a bug to me.

A bigger question I have is:  shouldn't we be pulling from `os.environ` [as we do in the kubernetes launcher](https://github.com/jupyter-server/enterprise_gateway/blob/7901422c1e99881f449d0cacba514af08e410c48/etc/kernel-launchers/kubernetes/scripts/launch_kubernetes.py#L45-L64)?  Otherwise, the `env:` stanza that resides in the corresponding kernel.json file will be ignored, as will any envs specified in the `inherited_envs` and `process_envs` lists.

lresende: So, the scenario is that a variable was defined to flow from hub, but there is no actual value in some cases. If we skip, which I try, the value will be none and Spark operator will reject the crd. I initially tried to add "INVALID" but later changed to "None" which is more like a visual representation of actual None value.

Yes, I believe the env from kernel.json is being ignored, but I would say that is a separate issue.

kevin-bates: Hmm.  By "skip" I mean do not add it to the pod's environment - literally, skip it - there should be no _key_, let alone an associated _value_.  Since the launcher only handles `KERNEL_`-prefixed envs, your response, assuming it was not included in the env, implies that there's a `KERNEL_` env that is _required_ by the Spark application - which I find hard to believe.  If you have always included the env `key` (with a placeholder value), could you please try not including the `key` at all?

Is this because the template includes that particular `KERNEL_` env?  I suspect this is the case.  If so, we should handle CRD launches like pods where we (IIRC) _build_ the env stanza at the time the pod is launched.

> Yes, I believe the env from kernel.json is being ignored, but I would say that is a separate issue.

I've created #1224.  May I assign this to you?

lresende: So, the CRD template `sparkoperator.k8s.io-v1beta2.yaml.j2` has an entry like:
```
driver:
...
env:
  - name: FOO
     value: {{ kernel_foo }}
```

In this case, the kernel_foo has no value...

if I skip processing it, the value for FOO will be None and CRD will be rejected

Now, if you are suggesting we inspect the generated custom_resource_yaml and/or custom_resource_object to remove it, I believe I could look into that, but would it make it hard to investigate issues where something is not working because a given env var didn't have a value and we removed it?

kevin-bates: The templates are examples.  If, in your example, the `env` entry `FOO` was placed there by the EG maintainers as part of our examples, and the EG maintainers know that CRDs require values for each env, then the EG maintainers must ensure that env `FOO` has a value so as to not trigger failures when starting the CRD.

If `FOO` was added, post-release, then whatever adds it must ensure any envs added have values.  So I guess I'm saying that the template should be updated or _that_ value entry altered to include logic to set a default value.  Since CRDs have this "envs must require a value behavior" then I would argue we not include any envs that don't have a value.  

This is what we do in the kubernetes pod launcher where the pod template's [env stanza is essentially empty](https://github.com/jupyter-server/enterprise_gateway/blob/7901422c1e99881f449d0cacba514af08e410c48/etc/kernel-launchers/kubernetes/scripts/kernel-pod.yaml.j2#L43-L46), although the commented out help text in the CRD template should have some statement like "any envs added here must have values".

I just don't think we should be adding arbitrary values to _any_ env that doesn't have a value and, because CRDs behave this way, believe it would be better to log the env and move on.  Since, in this case, the env is ALSO defined in the template itself, then I would argue that's the responsibility of whoever added that entry since that's something we probably shouldn't _remove_ in the launcher code.  If they added it, they need to ensure it has a value.

kevin-bates: Thinking more about this, it's still unclear to me where this _envs-require-a-value_ requirement is coming from.  Is it CRDs in general (which seems odd since general pods don't impose this requirement) or Spark itself?   I suspect the latter and that "`FOO`" is really a variable like `SPARK_HOME` or something like that - which would make sense and requires a value.

So, in that vein of thought, I think we should do exactly what is done in the kubernetes-launcher - which is nothing; let none/empty values flow. 

Since [`jinja2` has the ability to set default values](https://jinja.palletsprojects.com/en/3.1.x/templates/#jinja-filters.default), and since we (EG) are only responsible for the templates _we_ produce, I think it's incumbent on the template "customizer" to ensure envs that happen to be required by the target application have values and, in this case, the template should be fixed to something like:
```
driver:
...
env:
  - name: FOO
     value: {{ kernel_foo|default('NO_VALUE') }}
```
Or whatever the customizer deems to be the appropriate default value.

PR #1225, opened 2022-12-19
delete PATH env conflict with container

jeremythuon:
Hello 

I found problem with custom container.
Since pull https://github.com/jupyter-server/enterprise_gateway/pull/1164, Kubernetes container automatically inherits the variables from flow.
With this, custom containers have their PATH env overwrite.
To avoid this, you must remove the modification of this variable before the creation of the container.

I found where PATH is set : 
https://github.com/jupyter-server/enterprise_gateway/blob/7901422c1e99881f449d0cacba514af08e410c48/enterprise_gateway/services/kernels/handlers.py#L57-L63

I don't why it's needed, but it's more secure to delete it just for container

### DIFF
--- a/enterprise_gateway/services/kernels/handlers.py
+++ b/enterprise_gateway/services/kernels/handlers.py
@@ -54,13 +54,15 @@ class MainKernelHandler(
         if model is not None and "env" in model:
             if not isinstance(model["env"], dict):
                 raise tornado.web.HTTPError(400)
-            # Start with the PATH from the current env. Do not provide the entire environment
-            # which might contain server secrets that should not be passed to kernels.
-            env = {"PATH": os.getenv("PATH", "")}
+
+            # Delete by default PATH from client envs
+            # Keeping this variable may break the remote kernel with a different system than the client.
+            # But if user want to keep this var, it's can be set in inherited_envs
+            model['env'].pop("PATH")
+
             # Transfer inherited environment variables from current process
-            env.update(
-                {key: value for key, value in os.environ.items() if key in self.inherited_envs}
-            )
+            env = {key: value for key, value in os.environ.items() if key in self.inherited_envs}
+
             # Allow all KERNEL_* envs and those specified in client_envs and set from client.  If this EG
             # instance is configured to accept all envs in the payload (i.e., client_envs == '*'), go ahead
             # and add those keys to the "working" allowed_envs list, otherwise, just transfer the configured envs.
--- a/enterprise_gateway/services/kernels/handlers.py
+++ b/enterprise_gateway/services/kernels/handlers.py
@@ -55,11 +55,6 @@ class MainKernelHandler(
             if not isinstance(model["env"], dict):
                 raise tornado.web.HTTPError(400)
 
-            # Delete by default PATH from client envs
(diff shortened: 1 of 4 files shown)


kevin-bates:
Hi @jeremythuon - thanks for opening this PR.  I agree this is an issue, but I think we should use a different resolution.  The issue with this solution is that users can specify the env for their kernels via the [`kernel.json`'s `env:` stanza](https://github.com/jupyter-server/enterprise_gateway/blob/7901422c1e99881f449d0cacba514af08e410c48/etc/kernelspecs/spark_python_kubernetes/kernel.json#L14-L19).  As a result, we should not preclude `PATH` entries in that stanza.  

Instead, I think you make a very valid statement in your opening description: "I don't why it's needed"

The purpose of the `inherited_envs` configurable is to specify which envs of the EG process to transfer to the kernel.  I don't recall why we transferred `PATH` unconditionally, but, since practically all kernels are remote from the EG server, the EG process's `PATH` stands a good chance of not being correct relative to the target host - regardless of launch type.  So, I think we should remove[ this code from the handlers.py file](https://github.com/jupyter-server/enterprise_gateway/blob/7901422c1e99881f449d0cacba514af08e410c48/enterprise_gateway/services/kernels/handlers.py#L57-L59), initialize `env` to an empty dictionary, and let folks configure `inherited_envs` to include `PATH` if they need the EG's `PATH` inherited by the kernel process.

Just a note... It can be a little confusing but the references to `kwargs["env"]` in `launch_process` represent the accumulation of the env entries from the client, those inherited from the EG process, and those specified in the `kernel.json`, so had the `kernel.json` included a `PATH` entry, this PR would have removed _that_ (arguably correct) env from the kernel's process, which we don't want.

Could you please explore the other approach to tackling this from the handlers.py file?  I'd rather not introduce an env-based list of variables to deal with unless we find we have to.  Thank you.

enterprise_gateway/services/kernels/handlers.py
```diff
@@ -54,13 +54,15 @@ async def post(self):
...
             if not isinstance(model["env"], dict):
                 raise tornado.web.HTTPError(400)
-            # Start with the PATH from the current env. Do not provide the entire environment
-            # which might contain server secrets that should not be passed to kernels.
-            env = {"PATH": os.getenv("PATH", "")}
+
+            # Delete by default PATH from client envs
+            # Keeping this variable may break the remote kernel with a different system than the client.
+            # But if user want to keep this var, it's can be set in inherited_envs
+            model['env'].pop("PATH")
```
kevin-bates, line 61: I think this needs to be removed because the user may have specified `"PATH"` in the set of `client_envs`.  In such cases, that configuration needs to be honored.  If that happens to cause issues, that's the user's issue to resolve.

jeremythuon: Yes you right,
I delete this too quickly 
Thanks

kevin-bates (changes requested):
The `Dockerfile` changes are reverting the work done in #1217. They need to be removed from this PR. The rest looks good.

kevin-bates (approved):
LGTM - Thank you @jeremythuon!

PR #1227, opened 2022-12-30
Support building image with dedicated platform by specifying compatible docker flags

luong-komorebi:
This pull is pretty simple, it adds a TARGET_ARCH variable to makefile and enable users to do something like:

```bash
make enterprise-gateway TARGET_ARCH='linux/amd64'
```

With the growing number of client machines like macbook m1, this can help users who do not want to build for multiple architecture but still can specify their desired architecture to work perfectly without having to change the source code. I have run into situation where docker build on macbook would just build an ARM version so the cloud linux amd 64 machine cannot run the executable file. 

---

### DIFF
--- a/etc/Makefile
+++ b/etc/Makefile
@@ -2,12 +2,14 @@
 # Distributed under the terms of the Modified BSD License.
 
 .PHONY: help clean clean-images clean-enterprise-gateway clean-enterprise-gateway-demo clean-demo-base \
-    clean-kernel-images clean-py clean-tf-py clean-tf-gpu-py clean-r clean-spark-r clean-scala toree-launcher \
-    kernelspecs_all kernelspecs_yarn kernelspecs_conductor kernelspecs_kubernetes kernelspecs_docker clean-kernel-image-puller
+		clean-kernel-images clean-py clean-tf-py clean-tf-gpu-py clean-r clean-spark-r clean-scala toree-launcher \
+		kernelspecs_all kernelspecs_yarn kernelspecs_conductor kernelspecs_kubernetes kernelspecs_docker clean-kernel-image-puller
 
 SA?=source activate
 ENV:=enterprise-gateway-dev
 SHELL:=/bin/bash
+SUPPORTED_ARCHS=linux/arm64 linux/amd64
+PLATFORM_ARCHS=`echo ${SUPPORTED_ARCHS} | sed "s/ /,/g"`
 
 # Docker attributes - hub organization and tag.  Modify accordingly
 HUB_ORG:=elyra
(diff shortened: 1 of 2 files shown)


etc/Makefile
```diff
@@ -156,6 +156,9 @@ ifdef MULTIARCH_BUILD
 	(cd ../build/docker/$1; docker buildx build ${NO_CACHE} --platform linux/arm64,linux/amd64 --build-arg HUB_ORG=${HUB_ORG} --build-arg TAG=${TAG} --build-arg SPARK_VERSION=${SPARK_VERSION} -t $(HUB_ORG)/$1:$(TAG) . --push)
 	@echo "remove builder instance $1"
 	-(docker buildx rm $1)
+else ifeq ($(TARGET_ARCH), $(filter $(TARGET_ARCH), linux/arm64 linux/amd64))
```
kevin-bates, line 159: Hmm.  I'm seeing an empty `--platform` entry when no target arch is specified.  (I believe CI is behaving the same as well.):
```bash
$ make kernel-py
make WHEEL_FILE=dist/jupyter_enterprise_gateway-3.2.0.dev0-py3-none-any.whl VERSION=3.2.0.dev0 NO_CACHE= TAG=dev SPARK_VERSION=3.2.1 MULTIARCH_BUILD= TARGET_ARCH= -C etc kernel-py
Untagged: elyra/kernel-py:dev
rm -f ../dist/jupyter_enterprise_gateway_kernel_image_files-3.2.0.dev0.tar.gz
a .
a ./bootstrap-kernel.sh
a ./kernel-launchers
a ./kernel-launchers/python
a ./kernel-launchers/R
a ./kernel-launchers/scala
a ./kernel-launchers/scala/lib
a ./kernel-launchers/scala/lib/toree-launcher_2.12-3.2.0.SNAPSHOT.jar
a ./kernel-launchers/scala/lib/toree-assembly-0.5.0-incubating.jar
a ./kernel-launchers/R/scripts
a ./kernel-launchers/R/scripts/server_listener.py
a ./kernel-launchers/R/scripts/launch_IRkernel.R
a ./kernel-launchers/python/scripts
a ./kernel-launchers/python/scripts/launch_ipykernel.py
(cd ../build/docker/kernel-py; docker build  --platform  --build-arg HUB_ORG=elyra --build-arg TAG=dev --build-arg SPARK_VERSION=3.2.1 -t elyra/kernel-py:dev .)
"docker build" requires exactly 1 argument.
See 'docker build --help'.

Usage:  docker build [OPTIONS] PATH | URL | -

Build an image from a Dockerfile
make[1]: *** [../.image-kernel-py] Error 1
gmake: *** [Makefile:160: kernel-py] Error 2
```

kevin-bates (commented):
Hi @luong-komorebi - great idea!  I'm seeing an invalid `--platform` option occurring however.

Also, since the set of supported architectures are now duplicated (one of which is comma-separated), I'm wonder if we could introduce a couple more constants to `etc/Makefile` - something like...

```
SUPPORTED_ARCHS=linux/arm64 linux/amd64
PLATFORM_ARCHS=`echo ${SUPPORTED_ARCHS} | sed "s/ /,/g"`
```
Ultimately, we could expose `SUPPORTED_ARCHS` to `Makefile` and flow its value, but that could be deferred as well.

kevin-bates (approved):
@luong-komorebi - thank you for another great contribution!

(Yeah, the build issue appears related to the linting step and `ruff` isn't very informative about which files are at fault.  I'm unable to reproduce the issue using the latest `ruff` which implies a transitive dependency.  I'll look into upon my return next week - unless someone wants to take a closer look.  We may want to enable `-v` or `--show-source` options, or something along those lines.)

PR #1230, opened 2023-01-03
Start EG in demo-image to inherit PYSPARK_PYTHON env

kevin-bates:
I suspect the [change to not unconditionally inherit the value of env `PATH`](https://github.com/jupyter-server/enterprise_gateway/pull/1225) has side-affected the CI test `test_r_kernel.py::TestRKernelClient` because it can no longer find the `python` command (based on troubleshooting the issue).

Rather than flow `PATH` via the `inherited_envs` configurable (as suggested [here](https://github.com/jupyter-server/enterprise_gateway/pull/1225#issuecomment-1357919983)), this change updates the start script of the `enterprise-gateway-demo` image to add the value of env `PYSPARK_PYTHON` via `inherited_envs` since that's the minimal value necessary to get the `TestRKernelClient` test passing.  It also serves as an example of how to propagate a value from the EG process to kernels.


kevin-bates:
Thank you for your review @luong-komorebi - much appreciated!

PR #1231, opened 2023-01-04
Support 'cull connected kernels' as a config in helm and create a way to dynamically pass other environments variables to helm deployment

luong-komorebi:
To minimise the work for helm users in catching up with all the environment variables that EG might create and use, I create a `extraEnv` field, which will help the users create envs and values, even before maintainers port some of them as a config in helm

Speaking of configs, one major thing that enterprise gateway demo and start script is using that is yet to be configurable via helm is EG_CULL_CONNECTED. This PR also exposes that as a config, since we already have `EG_CULL_IDLE_TIMEOUT` in helm (although to be fair, one could use the extraEnv to achieve the same goal)


etc/kubernetes/helm/enterprise-gateway/values.yaml
```diff
@@ -62,6 +62,11 @@ deployment:
   tolerations: []
   affinity: {}
   nodeSelector: {}
+  extraEnv: {}
+    # EG_DEFAULT_KERNEL_MANAGER_CLASS: enterprise_gateway.services.kernels.remotemanager.YourOwnKernelManager
```
kevin-bates, line 66: I think we should reference an _existing_ env if we're going to use an `EG_`-prefixed env as an example.  Also, custom kernel manager class substitutions probably wouldn't come from the EG package itself, so this example is a little misleading (IMHO).

Could we use something like `EG_MAX_KERNELS_PER_USER` with a numeric value of, say, `10` instead?

kevin-bates (commented):
@luong-komorebi - this is great!  I just had a minor comment regarding the example used in the comment.  Thank you!

kevin-bates:
@luong-komorebi - would you mind merging with `main` so we can ensure the tests are clean?  I think the failures are due to not having the recently merged change in #1230.

etc/kubernetes/helm/enterprise-gateway/values.yaml
```diff
@@ -63,10 +63,13 @@ deployment:
   affinity: {}
   nodeSelector: {}
   extraEnv: {}
-    # EG_DEFAULT_KERNEL_MANAGER_CLASS: enterprise_gateway.services.kernels.remotemanager.YourOwnKernelManager
-    # SOME_ENV_VAR_WITH_LONG_VALUE: |-
-    #   Split it like this
-    #   For better readability
+    # EG_LIST_KERNELS: False
+    # EG_ALLOWED_KERNELS: >
```
kevin-bates, line 67: I thought this was an excellent example, then recalled that this will conflict with `kernel.allowedKernels` below.

Hmm, I also see `kernel.cullConnected` is defined as well - so we seem to have a number of helm variables that (ultimately) manifest themselves as envs in [`deployment.yaml`](https://github.com/jupyter-server/enterprise_gateway/blob/eb88e1b44a68e870c025c9a806d061db3394fa72/etc/kubernetes/helm/enterprise-gateway/templates/deployment.yaml#L49).

Note: There is a bug, however, in that there's no mapping in `deployment.yaml` from `.Values.kernel.cullConnected` to `EG_CULL_CONNECTED`.   Could you please include a fix for that in this PR?

How about using descriptive names like in your first commits, something like `SOME_ENV_VAR_WITH_SIMPLE_VALUE` and `SOME_ENV_VAR_WITH_LONG_VALUE`?

I really appreciate that you included both kinds of examples btw - thank you!

etc/kubernetes/helm/enterprise-gateway/values.yaml
kevin-bates: Excellent, thank you.

etc/kubernetes/helm/enterprise-gateway/values.yaml
kevin-bates: >> There is a bug, however, in that there's no mapping in deployment.yaml from .Values.kernel.cullConnected to EG_CULL_CONNECTED.

This was an oversight on my part actually.  I forgot that you had connected the dots already.

> In my opinion, it would be best to let the users manipulate all the EG related environment variables by themselves via a meta way like extraEnv and focus on exposing the list of possible environment variables 

I agree with this and wonder if we should refactor the helm scripts a bit to address this in the 4.0 timeframe.  I suppose this could be done prior to a major release if we found that necessary.

kevin-bates (approved):
The last set of changes look great - thank you!

PR #1234, opened 2023-01-12
Fix shutdown of CRD process proxy (and subclasses)

kevin-bates:
@lresende alerted me to an issue when restarting kernels managed by the `SparkOperatorProcessProxy`.  Looking into the issue, there were _several_ issues encountered that were preventing the graceful termination and, thus, successful restarts.

While troubleshooting the issue and given that regular `KubernetesProcessProxy` objects behaved properly, it was found that the override methods in both `CustomResourceProcessProxy` and `SparkOperatorProcessProxy` were essentially unnecessary with the exception of the API used to delete the CRD.

This PR refactors some of the `KubernetesProcessProxy` class hierarchy methods surrounding container status and termination.

While looking into this, it was also discovered that the EG's PATH was not available to the launch logic that takes place within the environment of the to-be-launched kernel.  This was another side-effect of #1164 that was similarly addressed in #1230.  In this case, we leverage the new "extra envs" addition recently added to the helm files via #1231 to inherit EG's `PATH`.

In addition, I noticed we weren't setting labels on the spark executor objects as we do in our other spark kernels, so I updated the spark operator launch script to set the same labels so administrators can include those objects in any "application analysis" that may be performed.


etc/kernel-launchers/operators/scripts/sparkoperator.k8s.io-v1beta2.yaml.j2
```diff
@@ -29,6 +29,10 @@ spec:
     coreLimit: 1000m
     memory: 1g
   executor:
+    labels:
+      kernel_id: "{{ kernel_id }}"
+      app: enterprise-gateway
+      component: worker
```
lresende, line 35: worker -> executor ?

lresende (approved):

lresende:
I have been testing this and all looks good to me. Thank you @kevin-bates for all the help here.

Issue #1235, opened 2023-01-13
Allow enterprise gateway to launch k8s kernels on remote clusters


Shrinjay:
### Problem
- Currently enterprise gateway launches kernels on the cluster where it is currently running. 
- This poses limitations in cases where we want the kernel to have access to resources on a remote cluster without running the enterprise gateway on that cluster specifically. 
- While the k8s client supports connecting to and launching/managing resources on a remote cluster, this feature just isn't implemented in the current client. 

### Proposed Solution
- The k8s client provides us a [simple example](https://github.com/kubernetes-client/python/blob/master/examples/remote_cluster.py) of configuring for remote cluster access
- To enable this in our client, we want to be able to pass in parameters for remote kernel access to a client and have it reuse that configuration in every instance. 
Therefore the code changes are as follows
- Wrap the k8s client in a python class that configures using the appropriate environment variables for remote or local cluster access
- Export this client atomically so only one instance is used in the codebase
- Replace current usages of the k8s client in the k8s proxy, custom resource proxy and launchers with our wrapped client

kevin-bates:
This sounds really interesting.  So a given EG server would still target exactly one Kubernetes cluster in which the kernels are launched - these changes simply enable the ability for _that_ cluster to be remote.  Is that correct?

You'll also need to take into account the possibility of the user having configured the `EG_SHARED_NAMESPACE` functionality.  In that case, the remote cluster may not have the `enterprise-gateway` namespace and I would argue that we should probably raise an error if remote-k8s-cluster AND eg-shared-namespace are both enabled.  The BYO namespace (using `KERNEL_NAMESPACE`) should still work, although we should probably ensure that `KERNEL_NAMESPACE` does not reflect the EG namespace in which EG resides.

At any rate, those are implementation details that we can work out.  We look forward to your pull request.

(Just a note that changes within the Process Proxies functionality, will be dropped in our EG 4.0 release in favor of using the kernel provisioner framework that is now part of the general Jupyter stack, so it would be great if you can participate in the [Gateway Provisioners](https://github.com/gateway-experiments/gateway_provisioners) project.  I'm (frantically) trying to get things in place for a release over there as we speak.  If you're unable to make applicable changes, no worries, we'll port them over at some point closer to making the switch.  I just wanted to make you aware of that future transition.)

Shrinjay:
[...]

PR #1244, opened 2023-01-27
Multi-Cluster Environments

Shrinjay:
# Problem Statement

This pull requests addresses issue #1235 and enables multi-cluster operations for Jupyter Enterprise Gateway. To reiterate the problem, currently enterprise gateway launches kernels on the cluster where it is currently running. This poses limitations in cases where we want the kernel to have access to resources on a remote cluster without running the enterprise gateway on that cluster specifically. Such cases often occur in the interest of security and isolation of internal services from production services. While the k8s client supports connecting to and launching/managing resources on a remote cluster, this feature just isn't implemented in the current client.

# Feature Description 

The changes in this PR implement the ability for users to provide a kubeconfig file for Jupyter Enterprise Gateway to use to launch kernels on a remote cluster that the kubeconfig points to. Specifically, Jupyter Enterprise Gateway will:

- Create a service account (if enabled)
- Create a namespaced role for the kernel pod
- Create a namespaced role binding for the kernel pod
- Create the kubernetes resource required for the kernel

## Operator Instructions:

1. Ensure your two clusters have interconnceted networks. Pods in the two clusters must be able to communicate with each other over pod IP alone.
2. Provide a kubeconfig file for use in the `config/` subdirectory of `etc/kubernetes/helm/enterprise-gateway` chart. 
3. Set `multicluster.enabled` to `true`. 

# Implementation Details

When the operator enables multicluster operation and installs/upgrades the helm chart, the following steps occur:
 - A ConfigMap is created for the kubeconfig and mounted into the enterprise gateway pod
 - A ConfigMap is created for the kernel pod role and mounted into the enterprise gateway pod
 - The following env vars are set for multi-cluster operation: `EG_USE_REMOTE_CLUSTER`, `EG_REMOTE_CLUSTER_KUBECONFIG_PATH`, `EG_DEFAULT_KERNEL_SERVICE_ACCOUNT_NAME`

When the enterprise gateway pod starts:
- An atomic kubernetes client is produced from the Kubernetes Client Factory pointing to the kubeconfig the operator passed in. All further operations use this client.

When a kernel is launched:
- If configured, a service account is created in the remote cluster (if one does not exist) for the kernel pod
- A namespaced role is created in the remote cluster for the kernel pod
- A namespaced role binding between the role and service account is created for the kernel pod
- The env variables above are passed to the kernel-launcher 
- The kernel-launcher uses the kubernetes client to launch the kernel in the remote cluster

Assuming network interconnection is setup correctly, the kernel pod should now launch and be able to communicate. 

Some notes on the implementation
- We've done our best to maintain as much of the smooth experience enterprise gateway currently has. However, automating network interconnection between two clusters is an infrastructure problem that's far too broad to be included as an automation here. Therefore, network interconnection is left as the responsibility of the operator. 
- We switch from using a statically defined client to an atomic one produced by a factory for best practice. It's much more clear where the configuration for a client comes from in this approach.
- We also strived to make as much of this configurable via Helm as possible. Therefore, the kernel role for the kernel pod is defined in helm, mounted as a configmap and then read by the process proxy, rather than defined in the process proxy itself like the role bindings.


Shrinjay:
Seems it was mentioned in another PR that the python interrupt test failures are a red herring, is that correct?

kevin-bates:
Hi @Shrinjay - thank you for providing this pull request.   I hope to be able to start reviewing this sometime this afternoon or tomorrow (PST).

> Seems it was mentioned in another PR that the python interrupt test failures are a red herring, is that correct?

That is correct.

lresende:
Are these merely to access resources on a different cluster in general? Or do we have the intention to enable mapping userA to clusterA and userB to clusterB? And how is that done?

kevin-bates:
Good question @lresende.  Reading the (excellent!) description, I believe this is more of a one-time thing -EG is either managing kernel pods within the cluster in which it resides, or EG is managing kernel pods within an _external_ cluster in which it has access but does not reside.  If that statement is correct, it might be better to update the title (and feature name) to something like _External Cluster Support_, or similar since the current title can imply multiple, and simultaneous, cluster support.

@Shrinjay - thoughts?

Shrinjay:
@kevin-bates I absolutely agree, I didn't think about it like that but calling it External Cluster and switching the helm charts to use the prefix `externalCluster` makes sense.

Shrinjay:
Updated!

enterprise_gateway/services/utils/envutils.py
```diff
@@ -0,0 +1,7 @@
+"""""Utilities to make checking environment variables easier"""
+import os
+
+
+def is_env_true(env_variable_name: str) -> bool:
+    """If environment variable is set and value is "TRUE" or "true", then return true. Else return false"""
```
kevin-bates, line 6:
```suggestion
    """If environment variable is set and value is case-insensitively "true", then return true. Else return false"""
```

etc/kubernetes/helm/enterprise-gateway/templates/deployment.yaml
```diff
@@ -103,27 +144,87 @@ spec:
         volumeMounts:
         - name: nfs-kernelspecs
           mountPath: "/usr/local/share/jupyter/kernels"
+        {{- if .Values.externalCluster.enabled }}
+        - name: kubeconfig
+          readOnly: true
+          mountPath: {{ .Values.externalCluster.configPath }}
+        {{- end}}
+        {{- if .Values.externalCluster.autoconfigureRemote }}
```
kevin-bates, line 152:
```suggestion
        {{- if .Values.externalCluster.autoConfigureRemote }}
```

etc/kubernetes/helm/enterprise-gateway/templates/deployment.yaml
```diff
@@ -103,27 +144,87 @@ spec:
...
       - name: nfs-kernelspecs
         nfs:
           server: {{ .Values.nfs.internalServerIPAddress }}
           path: "/usr/local/share/jupyter/kernels"
+      {{- if .Values.externalCluster.enabled }}
+      - name: kubeconfig
+        configMap:
+          name: {{ .Release.Name }}-configmap
+      {{- end}}
+      {{- if .Values.externalCluster.autoconfigureRemote }}
```
kevin-bates, line 167:
```suggestion
      {{- if .Values.externalCluster.autoConfigureRemote }}
```

etc/kubernetes/helm/enterprise-gateway/templates/deployment.yaml
```diff
@@ -103,27 +144,87 @@ spec:
...
   {{- else if .Values.kernelspecsPvc.enabled }}
         volumeMounts:
         - name: pvc-kernelspecs
           mountPath: "/usr/local/share/jupyter/kernels"
+        {{- if .Values.externalCluster.enabled }}
+        - name: kubeconfig
+          readOnly: true
+          mountPath: {{ .Values.externalCluster.configPath }}
+        {{- end}}
+        {{- if .Values.externalCluster.autoconfigureRemote }}
```
kevin-bates, line 181:
```suggestion
        {{- if .Values.externalCluster.autoConfigureRemote }}
```

etc/kubernetes/helm/enterprise-gateway/templates/deployment.yaml
```diff
@@ -103,27 +144,87 @@ spec:
...
       volumes:
       - name: pvc-kernelspecs
         persistentVolumeClaim:
           claimName: {{ .Values.kernelspecsPvc.name }}
+      {{- if .Values.externalCluster.enabled }}
+      - name: kubeconfig
+        configMap:
+          name: {{ .Release.Name }}-configmap
+      {{- end}}
+      {{- if .Values.externalCluster.autoconfigureRemote }}
```
kevin-bates, line 195:
```suggestion
      {{- if .Values.externalCluster.autoConfigureRemote }}
```

etc/kubernetes/helm/enterprise-gateway/templates/deployment.yaml
```diff
@@ -103,27 +144,87 @@ spec:
...
   {{- else if .Values.kernelspecs.image }}
         volumeMounts:
         - name: image-kernelspecs
           mountPath: "/usr/local/share/jupyter/kernels"
+        {{- if .Values.externalCluster.enabled }}
+        - name: kubeconfig
+          readOnly: true
+          mountPath: {{ .Values.externalCluster.configPath }}
+        {{- end}}
+        {{- if .Values.externalCluster.autoconfigureRemote }}
```
kevin-bates, line 209:
```suggestion
        {{- if .Values.externalCluster.autoConfigureRemote }}
```

etc/kubernetes/helm/enterprise-gateway/templates/deployment.yaml
```diff
@@ -103,27 +144,87 @@ spec:
...
       volumes:
       - name: image-kernelspecs
         emptyDir:
           medium: Memory
+      {{- if .Values.externalCluster.enabled }}
+      - name: kubeconfig
+        configMap:
+          name: {{ .Release.Name }}-configmap
+      {{- end}}
+      {{- if .Values.externalCluster.autoconfigureRemote }}
```
kevin-bates, line 223:
```suggestion
      {{- if .Values.externalCluster.autoConfigureRemote }}
```

etc/kubernetes/helm/enterprise-gateway/templates/kernel-role.yaml
```diff
@@ -0,0 +1,25 @@
+{{- define "kernel-role" }}
+{{- if .Values.externalCluster.autoconfigureRemote }}
```
kevin-bates, line 2:
```suggestion
{{- if .Values.externalCluster.autoConfigureRemote }}
```

etc/kubernetes/helm/enterprise-gateway/templates/role-configmap.yaml
```diff
@@ -0,0 +1,11 @@
+{{- if .Values.externalCluster.autoconfigureRemote }}
```
kevin-bates, line 1:
```suggestion
{{- if .Values.externalCluster.autoConfigureRemote }}
```

etc/kubernetes/helm/enterprise-gateway/values.yaml
```diff
@@ -179,3 +179,13 @@ kip:
...
+
+externalCluster:
+  # Set to enable to point k8s client to remote cluster
+  enabled: false
+  # Mount config at this path inside pod
+  configPath: /etc/kube/config/
+  # Filename of kubeconfig
+  configFilename: kubeconfig
+  # Automatically create service account and role in remote cluster
+  autoconfigureRemote: false
```
kevin-bates, line 191:
```suggestion
  autoConfigureRemote: false
```

enterprise_gateway/services/processproxies/k8s.py
```diff
@@ -227,7 +229,6 @@ def _determine_kernel_pod_name(self, **kwargs: dict[str, Any] | None) -> str:
         return pod_name
 
     def _determine_kernel_namespace(self, **kwargs: dict[str, Any] | None) -> str:
```
kevin-bates, line 231: (The following applies to lines L240-L254 below...)

1. Today, if the user is bringing their own namespace via `KERNEL_NAMESPACE` it is their responsibility to ensure proper operation _within_ the namespace.  With this PR, and if `EG_USE_REMOTE_CLUSTER` is True, I'm assuming the same requirement holds.  Are there any pieces of new functionality that would preclude users from bringing their own namespace (where that namespace resides in the remote cluster)?

2. Clearly, both `EG_SHARED_NAMESPACE` = True and `EG_USE_REMOTE_CLUSTER` = True conflict.  Perhaps we should check for `EG_SHARED_NAMESPACE` = True when obtaining the k8s client and either log and return the _local_ client or raise an exception?  

((Just to note, in `gateway_provisioners` the _default_ value for `EG_SHARED_NAMESPACE` is True to allow for easier deployments outside of something like since they are available to _any_ `jupyter_client` application.))

kevin-bates (commented):
Hi @Shrinjay.  First I want to thank for the pull request and the _most excellent_ write up you provided to describe this feature.  That is much appreciated!

I've provided a few comments (most of which are a nit about camelCasing :camel: :smile:) but there are some questions regarding edge cases related to `EG_SHARED_NAMESPACE` (in which the kernels are meant to run in the same namespace as EG) and BYO namespaces (in which `KERNEL_NAMESPACE` is provided in the request from the client) among others - mostly minor.

Also, I don't have a way to test this. Have you ensured that spark-based kernels are properly behaved (primarily wrt creation of the executor pods, etc.)?

Thank you!

kevin-bates:
Hi @Shrinjay - gentle ping - could you please address the review comments?

Shrinjay:
@kevin-bates My deepest apologies, I was working on this as part of an initiative at work, but I got pulled off to work on something else and this completely slipped my mind. I'm back on this now, and my contract is coming to an end, so I'll be addressing these comments and looking to get this PR done within the upcoming weeks.

enterprise_gateway/services/external/k8s_client.py
```diff
@@ -0,0 +1,5 @@
+"""Instantiates a static global factory and a single atomic client"""
```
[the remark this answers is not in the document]
Shrinjay: Nope, I think that's fine, I can move them into processproxies and go from there

enterprise_gateway/services/processproxies/k8s.py
Shrinjay: 1. Something I realize should be documented better is that remote cluster operation _requires_ the namespace to be already created before we launch a kernel there. This was mostly a security concern, as long as operations are isolated within a namespace in a remote cluster it's fine, but creating a namespace is a cluster level operation which can be scary for operators to enable. 

2. Agreed on this, we should log a warning then ignore it if `EG_USE_REMOTE_CLUSTER` is enabled

Shrinjay:
@kevin-bates Made some updates based off your suggestions, however your comment on testing got me thinking, and I can't figure out a great way to test this in a repeatable way. Currently, we just test this manually, the only success criteria is that a kernel is able to launch as everything after that is the same as if it were running on a local cluster. I can't really tell if the processproxies and such are tested in itests as well. If they are, then we could replicate the processproxies test harness for this. Any guidance?

kevin-bates:
Hi @Shrinjay - thanks for the update.

Regarding testing, we have no tests for Kubernetes process proxies.  The integration test we have uses the `enterprise-gateway-demo` image to test process proxies `YarnClusterProcessProxy` and `DistributedProcessProxy` (in loopback mode).  I'm building a mock layer in Gateway Provisioners to try to improve this, but, for now, we can keep things as is and rely on manual testing.

My biggest areas of concern are the behaviors around _BYO Namespace_ (where the client pre-creates the namespace and references that name in `KERNEL_NAMESPACE` in the start kernel request) and _Shared Namespace_ (where the kernels reside in the EG namespace).  The latter doesn't really apply since namespaces can't span clusters, but it would be good to understand the behavior - especially since this is the default in Gateway Provisioners.  However, if the other cluster happens to also defined the same-named namespace, what happens?  I think we should probably raise an exception in this case.

I will try to spend some time reviewing this PR in the next few days.

enterprise_gateway/services/processproxies/k8s.py
```diff
@@ -10,10 +10,15 @@
...
-from kubernetes import client, config
+import yaml
+from kubernetes import client
+from kubernetes.utils.create_from_yaml import create_from_yaml_single_item
+
+from enterprise_gateway.services.processproxies.k8s_client import kubernetes_client
 
 from ..kernels.remotemanager import RemoteKernelManager
 from ..sessions.kernelsessionmanager import KernelSessionManager
+from ..utils.envutils import is_env_true
```
kevin-bates, line 21: Could we refactor this filename to include an underscore for word separation?
```suggestion
from ..utils.env_utils import is_env_true
```

enterprise_gateway/services/processproxies/k8s.py
```diff
@@ -239,7 +248,7 @@ def _determine_kernel_namespace(self, **kwargs: dict[str, Any] | None) -> str:
 
         # If KERNEL_NAMESPACE was provided, then we assume it already exists.  If not provided, then we'll
         # create the namespace and record that we'll want to delete it as well.
-        namespace = kwargs["env"].get("KERNEL_NAMESPACE")
+        namespace = os.environ.get("KERNEL_NAMESPACE")
```
kevin-bates, line 251: Prior to launch, `KERNEL_NAMESPACE` must be pulled from the start request arguments (`kwargs["env"]`) rather than the EG process env, so this change needs to be reverted.
```suggestion
        namespace = kwargs["env"].get("KERNEL_NAMESPACE")
```

enterprise_gateway/services/processproxies/k8s.py
```diff
@@ -310,14 +326,66 @@ def _create_kernel_namespace(self, service_account_name: str) -> str:
...
+                        name=namespace, body=body
+                    )
                     self.log.warning(f"Deleted kernel namespace: {namespace}")
                 else:
                     reason = f"Error occurred creating namespace '{namespace}': {err}"
                 self.log_and_raise(http_status_code=500, reason=reason)
 
         return namespace
 
+    def _create_service_account_if_not_exists(
```
kevin-bates, line 339: If this is strictly for external clusters, could we rename this to something like: `_create_external_service_account_if_not_exists` or `_create_remote_service_account_if_not_exists`.  Since the other configurable options refer to "external", I guess I'd prefer the former.

enterprise_gateway/services/processproxies/k8s.py
```diff
@@ -310,14 +326,66 @@ def _create_kernel_namespace(self, service_account_name: str) -> str:
...
+            service_account_to_create: client.V1ServiceAccount = client.V1ServiceAccount(
+                kind="ServiceAccount", metadata=service_account_metadata
+            )
+
+            client.CoreV1Api(api_client=kubernetes_client).create_namespaced_service_account(
+                namespace=namespace, body=service_account_to_create
+            )
+
+            self.log.info(
+                f"Created service account {service_account_name} in namespace {namespace}"
```
kevin-bates, line 365: Is there any way to access the "name" of the external cluster?  Seems really helpful to include that here.
```suggestion
                f"Created service account {service_account_name} in namespace {namespace} of external cluster {external_cluster}"
```

enterprise_gateway/services/processproxies/k8s.py
```diff
@@ -310,14 +326,66 @@ def _create_kernel_namespace(self, service_account_name: str) -> str:
...
+            )
+
+    def _create_role_if_not_exists(self, namespace: str) -> None:
```
kevin-bates, line 368: A similar name change to that suggested previously would be nice here.

enterprise_gateway/services/processproxies/k8s.py
```diff
@@ -310,14 +326,66 @@ def _create_kernel_namespace(self, service_account_name: str) -> str:
...
+        """If role doesn't exist in target cluster, create one. Occurs if a remote cluster is being used"""
+        role_yaml_path = os.getenv('EG_REMOTE_CLUSTER_ROLE_PATH')
```
kevin-bates, line 370: Should `role_yaml_path` be validated for `None` and the file's existence?

This line (and any validation) should be moved within the `if kernel_cluster_role not in remote_cluster_role_names:` block at L379 so that it's only performed if needed.

enterprise_gateway/services/processproxies/k8s_client_factory.py
```diff
@@ -0,0 +1,42 @@
...
+                context = os.getenv('EG_REMOTE_CLUSTER_CONTEXT', None)
+                config.load_kube_config(
+                    client_configuration=kubernetes_config,
+                    config_file=kubeconfig_path,
+                    context=context,
+                )
+            else:
+                if is_env_true('EG_USE_REMOTE_CLUSTER'):
+                    self.log.warning(
+                        "Cannot use EG_USE_REMOTE_CLUSTER and EG_SHARED_NAMESPACE at the same time. Using local cluster...."
```
kevin-bates, line 34: :+1:

kevin-bates (commented):
Hi @Shrinjay - I didn't have a chance to verify _existing_ behavior, but plan to do so before its merge.  I had some relative minor comments.  Thanks again for contributing this!

Qloop:
We have recently encountered a similar requirement and are considering trying this PR. Do you have any other suggestions?

PR #1249, opened 2023-02-04
Dump Kubernetes YAML/CRD on kernel startup fails

lresende:
A submitted kernel can fail due to the malformation of 
YAML/CRD contents and in this case, dumping the 
files in the log can help with problem determination.

### DIFF
--- a/etc/kernel-launchers/kubernetes/scripts/launch_kubernetes.py
+++ b/etc/kernel-launchers/kubernetes/scripts/launch_kubernetes.py
@@ -225,7 +225,7 @@ def launch_kubernetes_kernel(  # noqa
                     f"kernel launch terminating!"
                 )
         else:
-            print(f"ERROR processing Kubernetes yaml file - kernel launch terminating!")
+            print("ERROR processing Kubernetes yaml file - kernel launch terminating!")
             print(k8s_yaml)
             sys.exit(
                 f"ERROR - Unknown Kubernetes object '{k8s_obj}' found in yaml file - kernel launch terminating!"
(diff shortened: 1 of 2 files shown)


etc/kernel-launchers/kubernetes/scripts/launch_kubernetes.py
```diff
@@ -225,6 +225,8 @@ def launch_kubernetes_kernel(  # noqa
                     f"kernel launch terminating!"
                 )
         else:
+            print("ERROR processing Kubernetes yaml file - kernel launch terminating!")
+            print(k8s_yaml)
```
kevin-bates, line 229: Hi @lresende.  Do you have an example of what the print output looks like?  I suspect this will output the entire yaml on a single line and I wonder, since this would be used primarily for troubleshooting, if using pretty print `pprint` would be more helpful?

lresende: It does in a pretty way, at least on the spark operator:
```
apiVersion: "sparkoperator.k8s.io/v1beta2"
kind: SparkApplication
metadata:
  name: xxxx-dafe8b37-9579-4cda-a8ab-d9f33a6dadba
  labels:
    proxyUser: "xxxxx"
spec:
  serviceAccountName: xxx
  mode: cluster
  type: Python
```

kevin-bates: Ok. Cool.

kevin-bates (approved):
Thanks @lresende, this will be helpful.

kevin-bates (approved):
Thanks @lresende, this will be useful.

PR #1257, opened 2023-02-23
A few small changes to gateway client

pq43:
I made a few small changes to the gateway client locally that I wanted to contribute back. One big issue I dealt with was with my connection to the websocket and http api endpoints failing, so I think it would be really helpful if in the code there was somehow an indication to use secure endpoints. Additionally, we found that we needed to pass a lot of environment variables, so an extra_env dictionary is helpful if you're calling the gateway client and need to pass, for example, queue-specific environment variables. Additionally, in kernel.execute, we should somehow let whatever's calling that function know whether an error was returned or not. As it stands now, kernel.execute will take an error and convert it to text - but what if for example, we're testing some notebook code - to do that, you would need to know if the kernel got an error when executing a cell or not.

I'm open to any edits/suggestions, these are just changes we had to make locally and think could be helpful for other people testing :)

### DIFF
--- a/enterprise_gateway/client/gateway_client.py
+++ b/enterprise_gateway/client/gateway_client.py
@@ -119,9 +119,13 @@ class KernelClient:
         self.log = logger
         self.log.debug(f"Initializing kernel client ({kernel_id}) to {self.kernel_ws_api_endpoint}")
 
-        self.kernel_socket = websocket.create_connection(
-            self.kernel_ws_api_endpoint, timeout=timeout, enable_multithread=True
-        )
+        try:
+            self.kernel_socket = websocket.create_connection(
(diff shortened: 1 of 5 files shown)


pq43:
sorry, will fix the remaining tests tmrw

pq43:
i have no clue why the link check is failing lol, seemed to be working fine yesterday

kevin-bates (approved):
Hi @pq43 - thank you for the pull request!  Although GatewayClient is more for testing and demonstration, these changes definitely make its use more practical and robust - thank you!

The latest test failures are related to the Gitter badge not being found due to their migration to matrix/element.  I'll fix that via another PR.

PR #1267, opened 2023-03-03
Replace "is defined" in Jinja templates if conditions

pq43:
Removed "is defined" for Jinja templates if conditions. I believe `if variable` is preferred to `if variable is defined` because `if variable` will return false if the variable is defined but empty, e.g. like an empty string or array. It's also cleaner unless the intended behavior is actually to return true even with an empty string/array passed as the env variable.

For example, using `if variable`:
image
as compared to `if variable is defined`...:
image

### DIFF
--- a/docs/source/operators/deploy-kubernetes.md
+++ b/docs/source/operators/deploy-kubernetes.md
@@ -722,12 +722,12 @@ metadata:
 spec:
   restartPolicy: Never
   serviceAccountName: "{{ kernel_service_account_name }}"
-  {% if kernel_uid is defined or kernel_gid is defined %}
+  {% if kernel_uid or kernel_gid %}
   securityContext:
-    {% if kernel_uid is defined %}
+    {% if kernel_uid %}
     runAsUser: {{ kernel_uid | int }}
     {% endif %}
-    {% if kernel_gid is defined %}
+    {% if kernel_gid %}
     runAsGroup: {{ kernel_gid | int }}
     {% endif %}
     fsGroup: 100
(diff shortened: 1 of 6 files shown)


kevin-bates:
Hi @pq43 - was it your intention to close this PR?

pq43:
Woops, my bad, reopened! It is failing some tests though, which I'm kinda confused about - but I will take a look and see if I can fix them :)

pq43:
```Step 4/24 : FROM $BASE_CONTAINER
manifest for elyra/demo-base:3.3.0.dev0 not found: manifest unknown: manifest unknown
```
This is that it looks like the error is: [729](https://github.com/jupyter-server/enterprise_gateway/actions/runs/4327504739/jobs/7556212005#step:7:730)... any idea what's going on?

kevin-bates:
Hi @pq43 - don't worry about the build failure - I see what is happening and it's unrelated to your PR.  I'll submit a fix for it soon.

I have a question regarding your changes and the "falsey" behavior.  

So, is that possible if a user wanted to set the CPUs to 0 (or any numeric value to 0)?  It seems like with this change, the updates to the numeric-based entries could never be set to 0 resulting in their "default values" being used (since they won't appear to be defined to 0).

pq43:
Oh yeah, that's definitely true - I actually had that same concern as well. I think removing `is defined` only makes sense for non-numeric values, thank you for spotting that.

I just made the change now to add the `is defined` behavior back for numeric values, and only removed the `is defined` for non numeric values.

docs/source/operators/deploy-kubernetes.md
```diff
@@ -722,12 +722,12 @@ metadata:
 spec:
   restartPolicy: Never
   serviceAccountName: "{{ kernel_service_account_name }}"
-  {% if kernel_uid is defined or kernel_gid is defined %}
+  {% if kernel_uid or kernel_gid %}
   securityContext:
-    {% if kernel_uid is defined %}
+    {% if kernel_uid %}
```
kevin-bates, line 727: I think `kernel_uid` and `kernel_gid` should be settable to 0 as well -although it's never recommended to run as `root`, we shouldn't preclude it.

pq43: Oh interesting, didn't know that, fixed

pq43:
Should be good now, sorry about all that

kevin-bates:
Hi @pq43 - PR #1268 fixes the build issue.  Would you mind updating to `main` so we can ensure the builds are green?  Thank you.

kevin-bates (approved):
LGTM - thanks @pq43.

PR #1269, opened 2023-03-05
Set restart strategy to never for spark operator

lresende:
While reviewing some of the spark operator artifacts 
I noticed it was not setting the Kubernetes 
restart policy to never as the other kernelspecs


kevin-bates (commented):
Good catch on the restart policy. 

Does the disable evictions apply to non crd objects as well?  If so, it seems like it should be set in the kernel pod yaml as well.

lresende:
@kevin-bates Kubernetes Python kernel updated

kevin-bates (approved):
These changes look good, thank you @lresende!

Issue #1258, opened 2023-02-23
Enterprise gateway Spark Operator (Python) kernel fails to systematic restart


vivekbhide:
Details - When using the Spark Operator (Python)  kernel, the kernel restart fails on selecting any of the 'Restart kernel ...' option from notebook file menu

Enterprise gateway version tested - 3.2.0

Steps to reproduce -

1. Create a spark notebook with enterprise gateway managing the kernels to run with
2. Run the notebook with kernel 'Spark Operator (Python)'.. Basic operation of creating spark session (hence creating spark driver and executor pods behind the scene) with some spark read files or spark sql are expected
3. Go to Kernel menu on jupyter lab. Choose any 'Restart Kernel...' option

Expected Result -

1. Kernel restarts with
   a. Any executor pod getting terminated
   b. Any driver prod restarting and going to 'Running' phase

Actual Results -

1. Driver and executor pods terminates but driver pod doesn't restart. 
2. Enterprise gateway goes into infinite loop of checking the driver status, followed by erroring with 'Kernel launch timeout' exception and continuing status on the driver pod
3. Jupyter lab UI, stuck with busy kernel status

Observation -

1. There are couple of issues reported with spark operator regarding issues with delete and/or restart of spark application. It might be contributing to this behavior https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/issues/1566 and https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/issues/1558
2. Issue is not always reproducible but the probability of it occurring increases when you not only start a kernel and try to restart but also run some spark operations and then try to restart

Logs -
Enterprise gateway logs -

```
[D 2023-02-21 00:33:23.314 EnterpriseGatewayApp] container jovyan-72252bdb-da19-42b0-8aed-61fd424b513a-driver status in get_container_status -> 'Running'
[D 2023-02-21 00:33:23.813 EnterpriseGatewayApp] Signal (2) sent via gateway communication port.
[D 2023-02-21 00:33:23.814 EnterpriseGatewayApp] Shutdown request sent to listener via gateway communication port.
[...]

PR #1271, opened 2023-03-05
Properly propagate errors on Spark Operator CRD failures

### LINKED ISSUE #1266: Failed to start kernels (Spark Operator) keeps trying to connect indefinitely
## Description

When a Spark Operator/CRD kernel fails to start on the spark operator side, the gateway keeps pulling for connection indefinitely.

## Expected behavior

I believe there are two issues here:
- When a submitted CRD fails, gateway should recognize it as failed
- In the case this is caused any other way, after some sort of timeout/number of retries, it should stop, mark kernel as dead, and stop the loop

## Logs

```
[D 2023-03-03 16:26:26.175 EnterpriseGatewayApp] 193: Waiting to connect to k8s sparkapplication in namespace 'spark-applications'. Name: 'some-user-68415e17-3ced-497f-86de-9c59285e2fec-driver', Status: 'None', Pod IP: 'None', KernelID: '68415e17-3ced-497f-86de-9c59285e2fec'
[D 2023-03-03 16:26:26.190 EnterpriseGatewayApp] Nudge: attempt 77 on kernel 68415e17-3ced-497f-86de-9c59285e2fec
[D 2023-03-03 16:26:26.690 EnterpriseGatewayApp] 194: Waiting to connect to k8s sparkapplication in namespace 'spark-applications'. Name: 'some-user-68415e17-3ced-497f-86de-9c59285e2fec-driver', Status: 'None', Pod IP: 'None', KernelID: '68415e17-3ced-497f-86de-9c59285e2fec'
[D 2023-03-03 16:26:26.695 EnterpriseGatewayApp] Nudge: attempt 78 on kernel 68415e17-3ced-497f-86de-9c59285e2fec
[D 2023-03-03 16:26:27.208 EnterpriseGatewayApp] 195: Waiting to connect to k8s sparkapplication in namespace 'spark-applications'. Name: 'some-user-68415e17-3ced-497f-86de-9c59285e2fec-driver', Status: 'None', Pod IP: 'None', KernelID: '68415e17-3ced-497f-86de-9c59285e2fec'
[D 2023-03-03 16:26:27.213 EnterpriseGatewayApp] Nudge: attempt 79 on kernel 68415e17-3ced-497f-86de-9c59285e2fec
[D 2023-03-03 16:26:27.729 EnterpriseGatewayApp] 196: Waiting to connect to k8s sparkapplication in namespace 'spark-applications'. Name: 'some-user-68415e17-3ced-497f-86de-9c59285e2fec-driver', Status: 'None', Pod IP: 'None', KernelID: '68415e17-3ced-497f-86de-9c59285e2fec'
[W 2023-03-03 16:26:27.734 EnterpriseGatewayApp] Nudge: attempt 80 on kernel 68415e17-3ced-497f-86de-9c59285e2fec
[D 2023-03-03 16:26:28.247 EnterpriseGatewayApp] 197: Waiting to connect to k8s sparkapplication in namespace 'spark-applications'. Name: 'some-user-68415e17-3ced-497f-86de-9c59285e2fec-driver', Status: 'None', Pod IP: 'None', KernelID: '68415e17-3ced-497f-86de-9c59285e2fec'
[D 2023-03-03 16:26:28.252 EnterpriseGatewayApp] Nudge: attempt 81 on kernel 68415e17-3ced-497f-86de-9c59285e2fec
[D 2023-03-03 16:26:28.765 EnterpriseGatewayApp] 198: Waiting to connect to k8s sparkapplication in namespace 'spark-applications'. Name: 'some-user-68415e17-3ced-497f-86de-9c59285e2fec-driver', Status: 'None', Pod IP: 'None', KernelID: '68415e17-3ced-497f-86de-9c59285e2fec'
[D 2023-03-03 16:26:28.770 EnterpriseGatewayApp] Nudge: attempt 82 on kernel 68415e17-3ced-497f-86de-9c59285e2fec
[D 2023-03-03 16:26:29.286 EnterpriseGatewayApp] 199: Waiting to connect to k8s sparkapplication in namespace 'spark-applications'. Name: 'some-user-68415e17-3ced-497f-86de-9c59285e2fec-driver', Status: 'None', Pod IP: 'None', KernelID: '68415e17-3ced-497f-86de-9c59285e2fec'
[D 2023-03-03 16:26:29.287 EnterpriseGatewayApp] Nudge: attempt 83 on kernel 68415e17-3ced-497f-86de-9c59285e2fec
[D 2023-03-03 16:26:29.789 EnterpriseGatewayApp] Nudge: attempt 84 on kernel 68415e17-3ced-497f-86de-9c59285e2fec
[D 2023-03-03 16:26:29.807 EnterpriseGatewayApp] 200: Waiting to connect to k8s sparkapplication in namespace 'spark-applications'. Name: 'some-user-68415e17-3ced-497f-86de-9c59285e2fec-driver', Status: 'None', Pod IP: 'None', KernelID: '68415e17-3ced-497f-86de-9c59285e2fec'
[D 2023-03-03 16:26:30.295 EnterpriseGatewayApp] Nudge: attempt 85 on kernel 68415e17-3ced-497f-86de-9c59285e2fec
[D 2023-03-03 16:26:30.327 EnterpriseGatewayApp] 201: Waiting to connect to k8s sparkapplication in namespace 'spark-applications'. Name: 'some-user-68415e17-3ced-497f-86de-9c59285e2fec-driver', Status: 'None', Pod IP: 'None', KernelID: '68415e17-3ced-497f-86de-9c59285e2fec'
[D 2023-03-03 16:26:30.798 EnterpriseGatewayApp] Nudge: attempt 86 on kernel 68415e17-3ced-497f-86de-9c59285e2fec
[D 2023-03-03 16:26:30.846 EnterpriseGatewayApp] 202: Waiting to connect to k8s sparkapplication in namespace 'spark-applications'. Name: 'some-user-68415e17-3ced-497f-86de-9c59285e2fec-driver', Status: 'None', Pod IP: 'None', KernelID: '68415e17-3ced-497f-86de-9c59285e2fec'
```

lresende:
Properly propagate errors on Spark Operator CRD failures
- Properly report failures on Spark Operator kernels
- Simplify get_container_status signature to always return a string and in lowercase

Fixes #1266 
Should also address #1258

### DIFF
--- a/enterprise_gateway/services/processproxies/container.py
+++ b/enterprise_gateway/services/processproxies/container.py
@@ -148,7 +148,7 @@ class ContainerProcessProxy(RemoteProcessProxy):
         # Do not check whether container_status is None
         # EG couldn't restart kernels although connections exists.
         # See https://github.com/jupyter-server/enterprise_gateway/issues/827
-        if container_status and container_status.lower in self.get_initial_states():
+        if container_status and container_status.lower() in self.get_initial_states():
             result = None
         return result
 
--- a/enterprise_gateway/services/processproxies/container.py
+++ b/enterprise_gateway/services/processproxies/container.py
@@ -244,7 +244,7 @@ class ContainerProcessProxy(RemoteProcessProxy):
 
     @abc.abstractmethod
     def get_container_status(self, iteration: int | None) -> str:
-        """Return current container state or empty string."""
+        """Returns the current container state (in lowercase) or the empty string if not available."""
         raise NotImplementedError
 
     @abc.abstractmethod
--- a/enterprise_gateway/services/processproxies/container.py
+++ b/enterprise_gateway/services/processproxies/container.py
@@ -239,7 +239,7 @@ class ContainerProcessProxy(RemoteProcessProxy):
 
     @abc.abstractmethod
     def get_error_states(self):
-        """Return list of states indicating container is starting (includes running)."""
+        """Returns the list of error states (in lowercase)."""
         raise NotImplementedError
 
     @abc.abstractmethod
--- a/enterprise_gateway/services/processproxies/container.py
+++ b/enterprise_gateway/services/processproxies/container.py
@@ -191,6 +191,7 @@ class ContainerProcessProxy(RemoteProcessProxy):
 
     async def confirm_remote_startup(self) -> None:
         """Confirms the container has started and returned necessary connection information."""
+        self.log.debug("Trying to confirm kernel container startup status")
         self.start_time = RemoteProcessProxy.get_current_time()
         i = 0
         ready_to_connect = False  # we're ready to connect when we have a connection file to use
@@ -200,12 +201,18 @@ class ContainerProcessProxy(RemoteProcessProxy):
 
             container_status = self.get_container_status(str(i))
             if container_status:
-                if self.assigned_host != "":
-                    ready_to_connect = await self.receive_connection_info()
-                    self.pid = (
-                        0  # We won't send process signals for kubernetes lifecycle management
+                self.log.debug(f"Received startup status: {container_status}")
+                if container_status.lower() == "failed":
+                    self.log_and_raise(
+                        http_status_code=500, reason="Error starting kernel container."
                     )
-                    self.pgid = 0
+                else:
+                    if self.assigned_host != "":
+                        ready_to_connect = await self.receive_connection_info()
+                        self.pid = (
+                            0  # We won't send process signals for kubernetes lifecycle management
+                        )
+                        self.pgid = 0
             else:
                 self.detect_launch_failure()
 
@@ -226,6 +233,11 @@ class ContainerProcessProxy(RemoteProcessProxy):
 
     @abc.abstractmethod
     def get_initial_states(self):
+        """Return list of states indicating container is starting (includes running)."""
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def get_error_states(self):
         """Return list of states indicating container is starting (includes running)."""
         raise NotImplementedError
 
--- a/enterprise_gateway/services/processproxies/container.py
+++ b/enterprise_gateway/services/processproxies/container.py
@@ -234,7 +234,7 @@ class ContainerProcessProxy(RemoteProcessProxy):
 
     @abc.abstractmethod
     def get_initial_states(self):
-        """Return list of states indicating container is starting (includes running)."""
+        """Return list of states in lowercase indicating container is starting (includes running)."""
         raise NotImplementedError
 
     @abc.abstractmethod
--- a/enterprise_gateway/services/processproxies/container.py
+++ b/enterprise_gateway/services/processproxies/container.py
@@ -204,7 +204,7 @@ class ContainerProcessProxy(RemoteProcessProxy):
                 self.log.debug(f"Received startup status: {container_status}")
                 if container_status in self.get_error_states():
                     self.log_and_raise(
-                        http_status_code=500, reason="Error starting kernel container."
+                        http_status_code=500, reason=f"Error starting kernel container; status: '{container_status}'."
                     )
                 else:
                     if self.assigned_host != "":
--- a/enterprise_gateway/services/processproxies/container.py
+++ b/enterprise_gateway/services/processproxies/container.py
@@ -199,9 +199,8 @@ class ContainerProcessProxy(RemoteProcessProxy):
             i += 1
             await self.handle_timeout()
 
-            container_status = self.get_container_status(str(i))
+            container_status = self.get_container_status(i)
             if container_status:
-                self.log.debug(f"Received startup status: {container_status}")
                 if container_status in self.get_error_states():
                     self.log_and_raise(
                         http_status_code=500,
--- a/enterprise_gateway/services/processproxies/container.py
+++ b/enterprise_gateway/services/processproxies/container.py
@@ -148,7 +148,7 @@ class ContainerProcessProxy(RemoteProcessProxy):
         # Do not check whether container_status is None
         # EG couldn't restart kernels although connections exists.
         # See https://github.com/jupyter-server/enterprise_gateway/issues/827
-        if container_status.lower() in self.get_initial_states():
+        if container_status and container_status.lower in self.get_initial_states():
             result = None
         return result
 
@@ -202,7 +202,7 @@ class ContainerProcessProxy(RemoteProcessProxy):
             container_status = self.get_container_status(str(i))
             if container_status:
                 self.log.debug(f"Received startup status: {container_status}")
-                if container_status.lower() == "failed":
+                if container_status.lower() in self.get_error_states():
                     self.log_and_raise(
                         http_status_code=500, reason="Error starting kernel container."
                     )
--- a/enterprise_gateway/services/processproxies/container.py
+++ b/enterprise_gateway/services/processproxies/container.py
@@ -148,7 +148,7 @@ class ContainerProcessProxy(RemoteProcessProxy):
         # Do not check whether container_status is None
         # EG couldn't restart kernels although connections exists.
         # See https://github.com/jupyter-server/enterprise_gateway/issues/827
-        if container_status and container_status.lower() in self.get_initial_states():
+        if container_status in self.get_initial_states():
             result = None
         return result
 
@@ -202,7 +202,7 @@ class ContainerProcessProxy(RemoteProcessProxy):
             container_status = self.get_container_status(str(i))
             if container_status:
                 self.log.debug(f"Received startup status: {container_status}")
-                if container_status.lower() in self.get_error_states():
+                if container_status in self.get_error_states():
                     self.log_and_raise(
                         http_status_code=500, reason="Error starting kernel container."
                     )
--- a/enterprise_gateway/services/processproxies/container.py
+++ b/enterprise_gateway/services/processproxies/container.py
@@ -148,7 +148,7 @@ class ContainerProcessProxy(RemoteProcessProxy):
         # Do not check whether container_status is None
         # EG couldn't restart kernels although connections exists.
         # See https://github.com/jupyter-server/enterprise_gateway/issues/827
-        if container_status in self.get_initial_states():
+        if container_status.lower() in self.get_initial_states():
             result = None
         return result
 
--- a/enterprise_gateway/services/processproxies/container.py
+++ b/enterprise_gateway/services/processproxies/container.py
@@ -204,7 +204,8 @@ class ContainerProcessProxy(RemoteProcessProxy):
                 self.log.debug(f"Received startup status: {container_status}")
                 if container_status in self.get_error_states():
                     self.log_and_raise(
-                        http_status_code=500, reason=f"Error starting kernel container; status: '{container_status}'."
+                        http_status_code=500,
+                        reason=f"Error starting kernel container; status: '{container_status}'.",
                     )
                 else:
                     if self.assigned_host != "":
--- a/enterprise_gateway/services/processproxies/container.py
+++ b/enterprise_gateway/services/processproxies/container.py
@@ -242,8 +242,8 @@ class ContainerProcessProxy(RemoteProcessProxy):
         raise NotImplementedError
 
     @abc.abstractmethod
-    def get_container_status(self, iteration_string):
-        """Return current container state."""
+    def get_container_status(self, iteration: int | None) -> str:
+        """Return current container state or empty string."""
         raise NotImplementedError
 
     @abc.abstractmethod
--- a/enterprise_gateway/services/processproxies/crd.py
+++ b/enterprise_gateway/services/processproxies/crd.py
@@ -75,7 +75,7 @@ class CustomResourceProcessProxy(KubernetesProcessProxy):
             application_state = custom_resource['status']['applicationState']['state'].lower()
 
             if iteration:
-                self.log.debug(f"Checking CRD status: {application_state}")
+                self.log.debug(f"CRD status is: {application_state}")
 
             if application_state in self.get_error_states():
                 exception_text = self._get_exception_text(
--- a/enterprise_gateway/services/processproxies/crd.py
+++ b/enterprise_gateway/services/processproxies/crd.py
@@ -61,3 +61,7 @@ class CustomResourceProcessProxy(KubernetesProcessProxy):
         result = delete_status and delete_status.get("status", None) in termination_stati
 
         return result
+
+    def get_initial_states(self) -> set:
+        """Return list of states in lowercase indicating container is starting (includes running)."""
+        return ["submitted", "pending", "running"]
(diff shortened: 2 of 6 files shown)


lresende:
@kevin-bates couple of questions here:
- What might be the better strategy here? Mostly if failures should be detected as quickly as in `get_container_status` or use `detect_launch_failure`. My understanding is that, even if  `detect_launch_failure` notify there is a failure, it keeps retrying until timeout which I believe it should not in this case.

lresende:
It looks like it's mostly working now, it can properly identify and fail quickly when a CRD fail, but it seems that there is a bug somewhere that, in case of success, is causing the kernel to restart.

kevin-bates:
> But what are your thoughts on what might be causing the restarts?

No idea. Where is the information for being able to analyze that? (A dozen lines of nudge logging isn't sufficient.)

If the restarts are automatic, that implies that the process proxy is unable to determine the pod is alive and is likely due to k8s resources, etc (since this isn't generally reproducible).

If we're purely going by the information in #1266, I can't tell that this is even a restart - it could be the initial startup.  Also, since there is no Pod IP or Status, I think that implies k8s has yet to schedule the pod.  I would try to issue some `describe` commands against the pod - assuming you see it running.  There's just not enough information here for me to answer your question.

enterprise_gateway/services/processproxies/container.py
```diff
@@ -200,12 +201,18 @@ async def confirm_remote_startup(self) -> None:
...
             container_status = self.get_container_status(str(i))
             if container_status:
-                if self.assigned_host != "":
-                    ready_to_connect = await self.receive_connection_info()
-                    self.pid = (
-                        0  # We won't send process signals for kubernetes lifecycle management
+                self.log.debug(f"Received startup status: {container_status}")
+                if container_status in self.get_error_states():
+                    self.log_and_raise(
+                        http_status_code=500, reason="Error starting kernel container."
```
kevin-bates, line 207: Could we add the status to this message?
```suggestion
                        http_status_code=500, reason=f"Error starting kernel container; status: '{container_status}'."
```

enterprise_gateway/services/processproxies/container.py
```diff
@@ -230,8 +237,13 @@ def get_initial_states(self):
         raise NotImplementedError
 
     @abc.abstractmethod
-    def get_container_status(self, iteration_string):
-        """Return current container state."""
+    def get_error_states(self):
+        """Return list of states indicating container is starting (includes running)."""
```
kevin-bates, line 241: Wrong docstring.  Let's convey its lowercase as well.
```suggestion
        """Returns the list of error states (in lowercase)."""
```

enterprise_gateway/services/processproxies/container.py
```diff
@@ -230,8 +237,13 @@ def get_initial_states(self):
...
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def get_container_status(self, iteration: int | None) -> str:
+        """Return current container state or empty string."""
```
kevin-bates, line 246: Let's emphasize the contract...
```suggestion
        """Returns the current container state (in lowercase) or the empty string if not available."""
```

enterprise_gateway/services/processproxies/docker_swarm.py
```diff
@@ -89,14 +89,14 @@ def get_container_status(self, iteration: int | None) -> str:
         """Return current container state."""
         # Locates the kernel container using the kernel_id filter.  If the status indicates an initial state we
         # should be able to get at the NetworksAttachments and determine the associated container's IP address.
-        task_state = None
+        task_state = ""
```
kevin-bates, line 92: This file is missing an implementation for 
```python
def get_error_states(self):
```

enterprise_gateway/services/processproxies/container.py
```diff
@@ -230,8 +237,13 @@ def get_initial_states(self):
         raise NotImplementedError
```
kevin-bates, line 237: Emphasize lowercase...
```python
"""Returns the list of states (in lowercase) indicating the container is starting (includes running)."""
```

enterprise_gateway/services/processproxies/spark_operator.py
```diff
@@ -22,3 +28,61 @@ def __init__(self, kernel_manager: RemoteKernelManager, proxy_config: dict):
         self.group = "sparkoperator.k8s.io"
         self.version = "v1beta2"
         self.plural = "sparkapplications"
+
+    def get_initial_states(self) -> set:
```
kevin-bates, line 32: These methods should reside on the `CustomResourceProcessProxy` since they provide nothing specific to the Spark Operator. (The help string in `get_container_status` should replace "spark operator" with "custom resource" or "custom resource definition".)

kevin-bates (commented):
I haven't taken this for a spin yet.  The general logic looks good.

enterprise_gateway/services/processproxies/spark_operator.py
```diff
@@ -22,3 +28,61 @@ def __init__(self, kernel_manager: RemoteKernelManager, proxy_config: dict):
...
+        delegate to parent to check if the application pod is running.
+
+        Returns
+        -------
+        Empty string if the container cannot be found otherwise.
+        The pod application status in case of success on Spark Operator side
+        Or the retrieved spark operator submission status in other cases (e.g. Failed)
+        """
+        with suppress(Exception):
+            custom_resource = client.CustomObjectsApi().get_namespaced_custom_object(
```
kevin-bates, line 52: How is this guaranteeing that the CRD relative to _this_ kernel is being returned?  Shouldn't this be using a [label selector like the superclass](https://github.com/jupyter-server/enterprise_gateway/blob/185f73bb5701cb0433790cd857a3a56d5da053ec/enterprise_gateway/services/processproxies/k8s.py#L83-L86)?

In fact, what is the purpose of this method?  We should get the status once.  Is the pod status from k8s_process_proxy not sufficient?

enterprise_gateway/services/processproxies/spark_operator.py
```diff
@@ -22,3 +28,61 @@ def __init__(self, kernel_manager: RemoteKernelManager, proxy_config: dict):
...
+                self.plural,
+                self.kernel_resource_name,
+            )
+
+            if not custom_resource:
+                return ""
+
+            application_state = custom_resource['status']['applicationState']['state'].lower()
+
+            self.log.debug(f"Checking CRD status: {application_state}")
```
kevin-bates, line 65: This message is being produced every 3 seconds - please make this conditional on `iteration is not None` (if we're keeping this method).

enterprise_gateway/services/processproxies/spark_operator.py
kevin-bates: If I comment out `get_container_status()` I still get properly started operators in the same amount of time with more consistent log statements.

lresende:
```        Submitting a new kernel application to the spark operator
        will take a while to be Running and can also fail before it
        can actually start any pods. So we will validate that the
        application succeeds on the Spark Operator side and then
        delegate to the parent to check if the application pod is running.
```

So, basically, this handles when the crd submission fails, and allow us to fail fast instead of waiting for the timeout period

kevin-bates: This method isn't failing fast.  Its caller, in the `ContainerProcessProxy` is what is deciding to fail fast.  This is ONLY logging a debug message that isn't providing any value (and skipping the display of useful troubleshooting information that could be provided prior to the CRD reaching the "running" state).

I really don't see what value this is providing, but, if it must be retained, this should be moved to the `CustomResourceProcessProxy` since there's literally nothing specific to Spark Application CRDs.

enterprise_gateway/services/processproxies/docker_swarm.py
lresende: they are not in use here, do you still want me to add them here?

enterprise_gateway/services/processproxies/spark_operator.py
lresende: When using CRD, the following happens -> Client --> Submit CRD -> Spark Operator -> Kubernetes -> POD.

The existing code in main today does not identify when a CRD fails (e.g. is rejected for malformation) and never creates a pod. In this case, the code in `ContainerProcessProxy` keeps waiting for the pod to be in a running state for a long period. With these changes, when we identify that the CRD submission failed, we pass to ContainerProcessProxy the failed status and it then fails right away, and when the CRD is processed, we then continue to wait for the pod to be in running state.

As for the location of the code, now I understand you suggestion, I am moving that to CRD.

enterprise_gateway/services/processproxies/container.py
```diff
@@ -191,6 +191,7 @@ def shutdown_listener(self):
 
     async def confirm_remote_startup(self) -> None:
         """Confirms the container has started and returned necessary connection information."""
+        self.log.debug("Trying to confirm kernel container startup status")
```
kevin-bates, line 194: Could we remove this please?

enterprise_gateway/services/processproxies/container.py
```diff
@@ -200,12 +201,19 @@ async def confirm_remote_startup(self) -> None:
 
             container_status = self.get_container_status(str(i))
             if container_status:
-                if self.assigned_host != "":
-                    ready_to_connect = await self.receive_connection_info()
-                    self.pid = (
-                        0  # We won't send process signals for kubernetes lifecycle management
+                self.log.debug(f"Received startup status: {container_status}")
```
kevin-bates, line 204: I don't think this is providing much either.

enterprise_gateway/services/processproxies/crd.py
```diff
@@ -40,6 +42,53 @@ async def launch_process(
...
+        will take a while to be Running and can also fail before it
+        can actually start any pods. So we will validate that the
+        application succeeds on the Spark Operator side and then
+        delegate to parent to check if the application pod is running.
+
+        Returns
+        -------
+        Empty string if the container cannot be found otherwise.
+        The pod application status in case of success on Spark Operator side
+        Or the retrieved spark operator submission status in other cases (e.g. Failed)
```
kevin-bates, line 58: please remove/replace references to spark operator.

enterprise_gateway/services/processproxies/crd.py
```diff
@@ -40,6 +42,53 @@ async def launch_process(
...
+            custom_resource = client.CustomObjectsApi().get_namespaced_custom_object(
+                self.group,
+                self.version,
+                self.kernel_namespace,
+                self.plural,
+                self.kernel_resource_name,
+            )
+
+            if not custom_resource:
+                return ""
```
kevin-bates, line 70: I really don't like how the logs will "go silent" for about 10 seconds, where you would normally see status messages.  Would it be helpful to produce a message in the same format as what the superclass `get_container_status()` produces:
```python
        if iteration:  # only log if iteration is not None (otherwise poll() is too noisy)
            self.log.debug(
                f"{iteration}: Waiting to connect to k8s {self.object_kind.lower()} in "
                f"namespace '{self.kernel_namespace}'. Name: '{self.container_name}', "
                f"Status: '{pod_status}', Pod IP: '{self.assigned_ip}', KernelID: '{self.kernel_id}'"
            )
```
If you did nothing but address the variable names, it would look similar to this (which was produced by commenting out this method):
```
[D 2023-03-07 19:35:30.743 EnterpriseGatewayApp] 1: Waiting to connect to k8s sparkapplication in namespace 'jovyan-bb097194-63ed-4586-bdc1-0411279c48ff'. Name: '', Status: '', Pod IP: 'None', KernelID: 'bb097194-63ed-4586-bdc1-0411279c48ff'
```
We could probably drop "Pod IP" (and Status).

enterprise_gateway/services/processproxies/crd.py
```diff
@@ -40,6 +42,53 @@ async def launch_process(
...
+
+            if application_state in self.get_error_states():
+                exception_text = self._get_exception_text(
+                    custom_resource['status']['applicationState']['errorMessage']
+                )
+                error_message = (
+                    f"CRD submission for kernel {self.kernel_id} failed: {exception_text}"
+                )
+                self.log.debug(error_message)
+            elif application_state == "running" and not self.assigned_host:
```
kevin-bates, line 85: Similar to the previous comment, could we change this to `else:` and call the superclass unconditionally at this point.  It already contains logic to scrutinize the `assigned_host` and `"running"` status.

enterprise_gateway/services/processproxies/crd.py
```diff
@@ -40,6 +42,53 @@ async def launch_process(
...
+                self.kernel_resource_name,
+            )
+
+            if not custom_resource:
+                return ""
+
+            application_state = custom_resource['status']['applicationState']['state'].lower()
+
+            if iteration:
+                self.log.debug(f"Checking CRD status: {application_state}")
```
kevin-bates, line 75:
```suggestion
                self.log.debug(f"CRD status is: {application_state}")
```

enterprise_gateway/services/processproxies/spark_operator.py
```diff
@@ -3,9 +3,13 @@
 # Distributed under the terms of the Modified BSD License.
 from __future__ import annotations
 
+import os
+
 from ..kernels.remotemanager import RemoteKernelManager
 from .crd import CustomResourceProcessProxy
 
+enterprise_gateway_namespace = os.environ.get('EG_NAMESPACE', 'default')
+
```
kevin-bates, line 12: These additions can be dropped.

kevin-bates (commented):
Just some further cleanup of logging stuff.

enterprise_gateway/services/processproxies/spark_operator.py
kevin-bates: Thanks, this is helpful. I keep forgetting that CRDs are intermediate objects.

enterprise_gateway/services/processproxies/crd.py
```diff
@@ -91,7 +91,8 @@ def get_container_status(self, iteration: int | None) -> str:
             if iteration:  # only log if iteration is not None (otherwise poll() is too noisy)
                 self.log.debug(
                     f"{iteration}: Waiting from CRD status from resource manager {self.object_kind.lower()} in "
-                    f"namespace '{self.kernel_namespace}'."
+                    f"namespace '{self.kernel_namespace}' Name: '{self.container_name}', "
+                    f"Status: '{application_state}', KernelID: '{self.kernel_id}'."
```
kevin-bates, line 95: Sorry, but this will do double logging with that from the superclass.  My suggestion was to add this only when not custom_resource status is available.

enterprise_gateway/services/processproxies/container.py
lresende: it's debug log, and helps to try to understand the area that things are happening on issues that are hard to find

enterprise_gateway/services/processproxies/container.py
lresende: it's debug log, and helps to try to understand the area that things are happening on issues that are hard to find

lresende:
All is good after the latest refactoring for the logs. Thank you @kevin-bates

kevin-bates (approved):
LGTM

PR #1277, opened 2023-03-13
Fix docker swarm not being to start due to missing inherited function on the latest version

luong-komorebi:
Since this [commit](https://github.com/jupyter-server/enterprise_gateway/commit/56dffccb7740dd0eeea115cfe7f4233599b6bc1e), starting kernel-py containers via dockerswarm proxy will eventually fail because when get_error_states is reached, NotImplementedError is raise, thus causing a disruption in the scheduling logic

This fix urges to improve that, by specifying error states, taken from https://docs.docker.com/engine/swarm/how-swarm-mode-works/swarm-task-states/

### DIFF
--- a/enterprise_gateway/services/processproxies/docker_swarm.py
+++ b/enterprise_gateway/services/processproxies/docker_swarm.py
@@ -50,6 +50,10 @@ class DockerSwarmProcessProxy(ContainerProcessProxy):
         """Return list of states in lowercase indicating container is starting (includes running)."""
         return {"preparing", "starting", "running"}
 
+    def get_error_states(self) -> set:
+        """Returns the list of error states indicating container is shutting down or receiving error."""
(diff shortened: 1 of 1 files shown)


kevin-bates:
Shoot.  I had [mentioned this](https://github.com/jupyter-server/enterprise_gateway/pull/1271#discussion_r1128184179) in my review, but the conversation was marked as "resolved" when it wasn't.  This is probably why it's best that the author of the comment determines its resolution.

I think the build issue is unrelated.  Looks like the apache repositories site is not available.  @lresende, are you able to determine what might be wrong?

kevin-bates (approved):
LGTM - thank you and sorry for the inconvenience.

lresende (approved):

lresende:
The build is ok now, merging.

PR #1279, opened 2023-03-14
Filter sensitive values from being logged in ProcessProxy

pq43:
Allows users to set kwarg `sensitive_env` for ProcessProxy, an array of phrases indicating secrets that users may not want logged.

### DIFF
--- a/enterprise_gateway/services/processproxies/processproxy.py
+++ b/enterprise_gateway/services/processproxies/processproxy.py
@@ -521,14 +521,24 @@ class BaseProcessProxyABC(metaclass=abc.ABCMeta):
         # Filter sensitive values from being logged
         sensitive_env = kwargs.get("sensitive_env")
         if sensitive_env is None:
-            sensitive_env = ["pwd", "password", "secret", "token", "key", "code", "auth", "state", "xsrf"]
+            sensitive_env = [
+                "pwd",
+                "password",
+                "secret",
+                "token",
+                "key",
+                "code",
+                "auth",
+                "state",
+                "xsrf",
+            ]
 
         env = kwargs.get("env").copy()
         for key in env.keys():
             if any(phrase in key.lower() for phrase in sensitive_env):
                 env.pop(key, None)
 
-        self.log.debug("BaseProcessProxy.launch_process() env: {}".format(env))
+        self.log.debug(f"BaseProcessProxy.launch_process() env: {env}")
 
     def launch_kernel(
         self, cmd: list[str], **kwargs: dict[str, Any] | None
--- a/enterprise_gateway/services/processproxies/processproxy.py
+++ b/enterprise_gateway/services/processproxies/processproxy.py
@@ -57,7 +57,7 @@ logging.getLogger("paramiko").setLevel(os.getenv("EG_SSH_LOG_LEVEL", logging.WAR
 env_pop_list = ["EG_REMOTE_PWD", "LS_COLORS"]
 
 # Comma separated list of env variables that shouldn't be logged
-sensitive_env_keys = os.getenv("EG_SENSITIVE_ENV_KEYS", "").split(",")
+sensitive_env_keys = os.getenv("EG_SENSITIVE_ENV_KEYS", "").lower().split(",")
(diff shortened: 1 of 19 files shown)


enterprise_gateway/services/processproxies/processproxy.py
```diff
@@ -518,7 +518,27 @@ async def launch_process(self, kernel_cmd: str, **kwargs: dict[str, Any] | None)
 
         self._enforce_authorization(**kwargs)
 
-        self.log.debug("BaseProcessProxy.launch_process() env: {}".format(kwargs.get("env")))
+        # Filter sensitive values from being logged
+        sensitive_env = kwargs.get("sensitive_env")
```
kevin-bates, line 522: Rather than make this something users can only configure programmatically, it seems like we should expose this as a configurable list of values: `SensitiveEnvKeys` - which can be seeded from a comma-separated list of values in env `EG_SENSITIVE_ENV_KEYS`.  We should default this value to `None` or the empty string (`""`).

enterprise_gateway/services/processproxies/processproxy.py
```diff
@@ -518,7 +518,27 @@ async def launch_process(self, kernel_cmd: str, **kwargs: dict[str, Any] | None)
...
+                "code",
+                "auth",
+                "state",
+                "xsrf",
+            ]
+
+        env_copy = kwargs.get("env").copy()
+        for key in list(env_copy):
+            if any(phrase in key.lower() for phrase in sensitive_env):
+                env_copy[key] = "[secret]"
```
kevin-bates, line 539: Similarly, we should expose a configurable string named `RedactionMask` (env: `EG_REDACTION_MASK`) which _could_ default to `"[secret]"`, although I think using `********` is probably a common default value for a redaction mask.

kevin-bates (commented):
Hi @pq43 - thank you for the pull request.  This seems like a decent idea although I think it needs to be configurable by an operator

enterprise_gateway/services/processproxies/processproxy.py
pq43: Done!

enterprise_gateway/services/processproxies/processproxy.py
kevin-bates: Yes.  Since we're offering this redaction capability (which I don't necessarily agree with in the first place) I think it would be good to offer a configurable mask value.  Organizations roll up logs and may have a common redaction mask they use to identify redacted areas.

If we want to change the default mask to `"[secret]"`, I'm okay with that, but it should be configurable.

pq43:
Oh okay, sounds good! To do what you said via traitlets, should I define EG_SENSITIVE_ENV_KEYS in mixins.py or is there a better place?

enterprise_gateway/services/processproxies/processproxy.py
```diff
@@ -56,6 +56,10 @@
 # Pop certain env variables that don't need to be logged, e.g. remote_pwd
 env_pop_list = ["EG_REMOTE_PWD", "LS_COLORS"]
 
+# Comma separated list of env variables that shouldn't be logged
+sensitive_env_keys = os.getenv("EG_SENSITIVE_ENV_KEYS", "").split(",")
```
kevin-bates, line 60: We should probably ensure these are lowercased prior to comparison.
```suggestion
sensitive_env_keys = os.getenv("EG_SENSITIVE_ENV_KEYS", "").lower().split(",")
```

kevin-bates:
> To do what you said via traitlets, should I define EG_SENSITIVE_ENV_KEYS in mixins.py or is there a better place?

Yes, but I'm thinking it might be best to just keep these as envs for now.  These will need to reside in [Gateway Provisioners](https://github.com/jupyter-server/gateway_provisioners) when EG moves to those and we could decide to use traits at that time, but introducing traits now will only create migration headaches.

enterprise_gateway/services/processproxies/processproxy.py
pq43: Done!

kevin-bates (approved):
Looks good @pq43 - thank you.

lresende (approved):

PR #1281, opened 2023-03-14
Move shutdown kernel code to be within KernelClient

pq43:
The heart of the code that shuts down the kernel (the API request) should be within `KernelClient` itself. This way, when the kernel can shut itself done without the help of `GatewayClient`. This is important in handling the websocket exception, as when we call `kernel.shutdown()`, we actually want to run the API request to shut down the kernel as well. 

Additionally, we should still raise the exception on `KernelClient.__init__()` if the websocket connection fails, as that way `GatewayClient.start_kernel` will error out if the connection to the `KernelClient` isn't working.

### DIFF
--- a/enterprise_gateway/client/gateway_client.py
+++ b/enterprise_gateway/client/gateway_client.py
@@ -88,7 +88,6 @@ class GatewayClient:
 
         kernel.shutdown()
 
-
 class KernelClient:
     """A kernel client class."""
 
--- a/enterprise_gateway/client/gateway_client.py
+++ b/enterprise_gateway/client/gateway_client.py
@@ -109,38 +109,37 @@ class KernelClient:
         self.kernel_id = kernel_id
         self.log = logger
         self.log.debug(f"Initializing kernel client ({kernel_id}) to {self.kernel_ws_api_endpoint}")
+        self.response_queues = {}
+        # startup reader thread
+
+        self.response_reader = Thread(target=self._read_responses)
+        self.response_reader.start()
+        self.interrupt_thread = None
+        self.kernel_socket = None
 
         try:
             self.kernel_socket = websocket.create_connection(
                 f"{ws_api_endpoint}/{kernel_id}/channels", timeout=timeout, enable_multithread=True
             )
         except Exception as e:
-            self.kernel_socket = None
             self.log.error(e)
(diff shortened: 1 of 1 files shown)


enterprise_gateway/client/gateway_client.py
```diff
@@ -139,20 +132,29 @@ def shutdown(self):
         # Terminate thread, close socket and clear queues.
         self.shutting_down = True
 
-        if self.kernel_socket:
+        if hasattr(self, 'kernel_socket') and self.kernel_socket:
```
kevin-bates, line 135: Are these attributes checks necessary?  How can `KernelClient.shutdown()` be called if there isn't a `KernelClient` instance - which is the only way these attributes would not be present?

kevin-bates (commented):
This makes good sense - thanks for the PR.

Just had the one comment about the attribute existence checks.

enterprise_gateway/client/gateway_client.py
pq43: I believe they are necessary, since when I tested exceptions on the `websocket.create_connection()` in `KernelClient.__init__()`, this line would crash as `kernel_socket` was not defined. The same occurred for `response_queues` and `response_reader`.

kevin-bates:
Hmm - I guess Python still creates an instance and `__init__()` isn't like a constructor in Java.

Could we please move the initializing of these 3 attributes to PRIOR to the `create_connection` call?

pq43:
Done

kevin-bates (approved):
Thanks @pq43.   I'm not sure if the thread start should be deferred until after the websocket connection has been established.  Let's keep our eyes out for issues there - although, again, this would only be a concern when the websocket connection fails.  The thread should terminate when the (parent) process terminates, so we're probably ok.

pq43:
Agreed, I was unsure about that too, I'll definitely submit another PR if I notice the thread being an issue

lresende (approved):

kevin-bates:
Why were the attribute checks restored?  Please see https://github.com/pq43/enterprise_gateway/pull/1 and consider merging it as part of this PR.

kevin-bates (approved):
Thanks @pq43!

Issue #1294, opened 2023-05-04
Conda -> Venv for local development


bloomsa:
### Problem

I'm wondering if there's any appetite for migrating from conda to venv for local development environment setup. I recently began the process in order to try tackling #1089 and I thought it might make more sense to use the [Python endorsed](https://packaging.python.org/en/latest/guides/installing-using-pip-and-virtual-environments/#creating-a-virtual-environment) virtual environment tool as long as we didn't need any specific conda features. It would reduce the barrier to entry for new contributions since more people might be familiar with setting up their virtual environments using that. 
### Proposed Solution

* Rewrite requirements.yml -> requirements.txt
  * this could instead be left off in favor of `pip install .` which seems to be what j[upyter_server](https://github.com/jupyter-server/jupyter_server/blob/main/CONTRIBUTING.rst#installing-the-jupyter-server) and [ipykernel](https://github.com/ipython/ipykernel/blob/main/CONTRIBUTING.md) suggest
* update the Makefile `env` target to use venv rather than conda

### Additional context

<!-- Add any other context or screenshots about the feature request here. You can also include links to examples of other programs that have something similar to your request. For example:

* Another project [...] solved this by [...]

kevin-bates:
Hi @bloomsa.  I certainly don't have any affinity for a conda environment to do development in.  That said, I don't think we should preclude devs from using either - which you not proposing either.  IMO, the dev should be responsible for ensuring the env in which they are operating is correct, and that is why we removed the forced use of (conda) envs in our `make` targets.  Yes, we left the ability to _create_, _activate_, and _destroy_ (conda) envs, but no other targets perform those actions.

We do, however, recommend the use of conda within our docker images, primarily because a conda installation typically includes a data science stack, so it eases the burden.  That said, most of our images derive from the jupyter docker stacks images, which already provide most of that heavy lifting.

I'm not sure a `requirements.txt` file is necessary since `pip install -e .` should obtain requirements from `pyproject.toml`.  Is that a correct statement?

So, in the end, I don't really see an issue with changing the recommended dev env type from conda to venv.  I will probably stick with mamba as my env "vendor".

I'm curious if @lresende or others have any opinions about this.

bloomsa:
[...]

PR #1297, opened 2023-05-06
Switch from conda to venv for development env

bloomsa:
addresses #1294 

Upon doing this, the "happy path" is simpler but overall it feels a little more frail / platform dependent. I'm curious to hear thoughts on that tradeoff or any suggestions people have seen from other repos

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -82,3 +82,6 @@ _site/
 # jetbrains ide stuff
 *.iml
 .idea/
+
+# virtual environment
+.venv/
--- a/Makefile
+++ b/Makefile
@@ -8,8 +8,7 @@
     push-kernel-images push-enterprise-gateway push-kernel-py push-kernel-spark-py push-kernel-r push-kernel-spark-r \
     push-kernel-scala push-kernel-tf-py push-kernel-tf-gpu-py push-kernel-image-puller publish helm-chart
 
-SA?=source activate
-ENV:=enterprise-gateway-dev
+ENV:=.venv
 SHELL:=/bin/bash
 MULTIARCH_BUILD?=
 TARGET_ARCH?=undefined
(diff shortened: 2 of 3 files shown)


bloomsa:
marked as draft until I get the docs updated

lresende:
Could you elaborate on the why?

bloomsa:
@lresende beyond what is in issue [#1294](https://github.com/jupyter-server/enterprise_gateway/issues/1294)?

lresende:
is there a way to make it a choice? instead of one or another?

kevin-bates:
Or remove this topic altogether.  To my earlier point, the environment is up to the dev, so should we even bother with providing these trivial helper methods at all?  We can _document suggestions_, but do we need the make target helpers?

bloomsa:
> We can document suggestions, but do we need the make target helpers?

I like this idea. I can update the docs to include suggestions for both venv and conda, but pull out the make targets altogether. On one hand, it feels kinda wrong to pull out something that already works, but my thinking behind it is that having the targets there makes it seem like conda is the _only_ way to build the local dev environment when that isn't really the case.

Are you both okay with that solution @lresende @kevin-bates

kevin-bates:
Hi @bloomsa - I apologize for the delay.  I'm fine with documenting suggestions.  I just don't really see the utility of having helper targets at this point.

@lresende - do you have a strong objection?

lresende:
@kevin-bates  Well, it looks like we have been there:
https://github.com/jupyter-server/enterprise_gateway/pull/1079

And than we came back:
https://github.com/jupyter-server/enterprise_gateway/pull/1108

Do you remember why we had to come back?

kevin-bates:
It appears we have come full circle. :smile:  @bloomsa contributed https://github.com/jupyter-server/enterprise_gateway/pull/1107 to introduce only the helper targets, but then we hit a bit of a snafu due to other changes that #1108 fixed.

Given this, I think we can proceed with removing the helper targets and _documenting_ suggestions for setting up `conda`/`mamba` and `venv` environments - as discussed above.

@lresende, @bloomsa - does this sound like a plan?

PR #1316, opened 2023-08-04
fix: passing eg env variables to kernel_pod

Chiragasourabh:
This fix passes all the environment variables of the host container to ProcessProxy while giving more preference to user overridden variables .


kevin-bates:
Hi @Chiragasourabh - please see this comment: https://github.com/jupyter-server/enterprise_gateway/discussions/1315#discussioncomment-6641642.  We really don't want to unconditionally transfer EG process envs to each kernel process.

PR #1343, opened 2023-10-19
Making bootstrap script respect current working dir

allstrive:
# Problem
The current usage of KERNEL_WORKING_DIR and EG_MIRROR_WORKING_DIRS in EG  works fine when EG is in charge of launching kernel PODs. 
Some of the usecases, uses sparkoperator to launch kernel POD. While the environment variable KERNEL_WORKING_DIR is still passed on , there's no effect of this environment var on actual execution environment.

# Solution
The bootstrap script change will ensure that KERNEL_WORKING_DIR is respected and that kernel is launched on right location.

### DIFF
--- a/etc/kernel-launchers/bootstrap/bootstrap-kernel.sh
+++ b/etc/kernel-launchers/bootstrap/bootstrap-kernel.sh
@@ -21,12 +21,20 @@ launch_python_kernel() {
   else
     kernel_class_option="--kernel-class-name ${KERNEL_CLASS_NAME}"
   fi
-
+  if [ ! -z "${KERNEL_WORKING_DIR}" ]
+  then
+    CURR_DIR=$(pwd)
+    cd "${KERNEL_WORKING_DIR}"
+  fi
 	set -x
 	python ${KERNEL_LAUNCHERS_DIR}/python/scripts/launch_ipykernel.py --kernel-id ${KERNEL_ID} \
 	      --port-range ${PORT_RANGE} --response-address ${RESPONSE_ADDRESS} --public-key ${PUBLIC_KEY} \
 	      --spark-context-initialization-mode ${KERNEL_SPARK_CONTEXT_INIT_MODE} ${kernel_class_option}
 	{ set +x; } 2>/dev/null
+  if [ ! -z "${CURR_DIR}" ]
+  then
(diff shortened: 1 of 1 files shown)


kevin-bates:
Hi @allstrive - thank for opening this pull request.  It's really too bad, but it doesn't appear operators have the notion of a [`workingDir`](https://github.com/jupyter-server/enterprise_gateway/blob/46a214c342a75cd599b175a0ee2ffa646b8de48b/etc/kernel-launchers/kubernetes/scripts/kernel-pod.yaml.j2#L77) setting.  (Have you confirmed that?)

Since these changes are in a shell script, I would like to see the same change applied to all kernel types.  Here are some suggestions.
1. Apply the change to all kernel types.
2. Only change the directory to `KERNEL_WORKING_DIR` if the directory exists (otherwise echo a statement that it doesn't exist and operations will remain in `cur_dir`).
3. Please use a lowercase variable name for `CURR_DIR` as we try to reserve all uppercase for env vars.
4. Since the kernel's lifecycle should be that of the container, I'm not sure its that critical to restore the directory anyway - so I view the whole `cur_dir` stuff as optional.

Also, it would be fantastic if you could contribute an identical change to the Gateway Provisioners repo in the [analogous file](https://github.com/jupyter-server/gateway_provisioners/blob/main/gateway_provisioners/kernel-launchers/bootstrap/bootstrap-kernel.sh) - since this repo will eventually replace the process-proxies (and their respective kernels) used in EG.

Thank you for your help and contribution!

allstrive:
Thank you @kevin-bates , for your suggestions.

> Apply the change to all kernel types.
Yes, that's what I intend to do before I put back the PR for review again

> Only change the directory to KERNEL_WORKING_DIR if the directory exists (otherwise echo a statement that it doesn't exist and operations will remain in cur_dir).
Great  point, will add that check.

> Please use a lowercase variable name for CURR_DIR as we try to reserve all uppercase for env vars.
Will follow that.

> Since the kernel's lifecycle should be that of the container, I'm not sure its that critical to restore the directory anyway - so I view the whole cur_dir stuff as optional.
I kept it as a good practice for any shell script, my realization is most of the components in Jupyter's ecosystem past EG are used in different ways across various usage. This will minimize the side effects of my change.

PR #1350, opened 2023-11-17
docs: reformat list

IITII:
The old list was very weird, sometimes it used "_" and sometimes it used "*", which confused me. So I created this PR

image


kevin-bates (approved):
Thanks for the updates and checking out EG!

PR #1353, opened 2023-12-13
Expose kernel_info_timeout configuration

lresende:


lresende:
@kevin-bates could you take a look and see if I forget any places related to exposing this new property

docs/source/operators/config-add-env.md
```diff
@@ -32,6 +32,9 @@ Besides those environment variables associated with configurable options, the fo
     startup attempt will take place.  If a second timeout occurs, Enterprise
     Gateway will report a failure to the client.
 
+  EG_KERNEL_LAUNCH_TIMEOUT=60
```
kevin-bates, line 35: Should this be `EG_KERNEL_INFO_TIMEOUT`?

docs/source/operators/config-add-env.md
```diff
@@ -32,6 +32,9 @@ Besides those environment variables associated with configurable options, the fo
...
+    The time (in seconds) Enterprise Gateway will wait before giving up on a kernel
```
kevin-bates, line 36: Can we be more explicit here?  What does "giving up on a kernel" mean?

docs/source/users/kernel-envs.md
```diff
@@ -55,6 +55,9 @@ There are several supported `KERNEL_` variables that the Enterprise Gateway serv
     expected to exceed that of the EG_KERNEL_LAUNCH_TIMEOUT set when Enterprise
     Gateway starts.
 
+  KERNEL_INFO_TIMEOUT=<from user> or EG_KERNEL_INFO_TIMEOUT=60
+    The time (in seconds) Enterprise Gateway will wait before giving up on a kernel
```
kevin-bates, line 59: Perhaps use the same text as the `EG_` variable (once clarified by previous comment), but also please add _how_ this is different (e.g., this only applies to the single kernel launch because its conveyed from the client or something).

enterprise_gateway/services/kernels/remotemanager.py
```diff
@@ -513,6 +515,10 @@ def _capture_user_overrides(self, **kwargs: dict[str, Any] | None) -> None:
         self.kernel_launch_timeout = float(
             env.get("KERNEL_LAUNCH_TIMEOUT", default_kernel_launch_timeout)
         )
+        # if KERNEL_INFO_TIMEOUT is passed in the payload, override it.
+        self.kernel_launch_timeout = float(
```
kevin-bates, line 519:
```suggestion
        self.kernel_info_timeout = float(
```

kevin-bates (commented):
Hi @lresende.  Aside from the comments posted here, I have some other comments/questions:
1. Could you please update the PR's description with _what_ this is, and _why_ it is necessary?  I'm assuming its something about the time to _receive_ the `kernel_info_reply` and is why its default spans that of `KERNEL_LAUNCH_TIMEOUT`.
2. What enforces the info timeout?  In process proxy we enforce the launch timeout if we have not received the connection info after that period of time.  I don't see similar for kernel-info.  

Since the kernel-info response is really handled via the websocket negotiation logic - which EG doesn't provide code for, it seems like this really needs to be part in JupyterServer rather than EG since there's no _place_ for the enforcement.  Hmm, and checking JupyterServer, I see `kernel_info_timeout` is [defined there already](https://github.com/jupyter-server/jupyter_server/blob/b0bbedf960ab3f600cd2bdf7e2e32c5c0400fff8/jupyter_server/services/kernels/kernelmanager.py#L132) so I guess I'm confused.  

Why does EG need to be concerned about this at all?  Seems like this is fully configurable (and usable) from the client entirely.

lresende:
> 2\. What enforces the info timeout?  In process proxy we enforce the launch timeout if we have not received the connection info after that period of time.  I don't see similar for kernel-info.

By exposing the property here and passing it to kernel manager, won't it be managed/enforced on the parent class? or are you saying I need to add additional support for enforcing on our side? this seemed to have worked on tests...

```
--RemoteMappingKernelManager.kernel_info_timeout=${EG_KERNEL_INFO_TIMEOUT}
```

lresende:
> 1. Could you please update the PR's description with _what_ this is, and _why_ it is necessary?  I'm assuming its something about the time to _receive_ the `kernel_info_reply` and is why its default spans that of `KERNEL_LAUNCH_TIMEOUT`.

Yes

kevin-bates:
>By exposing the property here and passing it to kernel manager, won't it be managed/enforced on the parent class? or are you saying I need to add additional support for enforcing on our side? this seemed to have worked on tests...
>
>--RemoteMappingKernelManager.kernel_info_timeout=${EG_KERNEL_INFO_TIMEOUT}

Ah, sorry, you're right, EG does _inherit_ most all of the plumbing for starting the kernel (not close enough to this anymore).  So you shouldn't need any of the changes in the `RemoteKernelManager` since `kernel_info_timeout` pertains only to the `MappingKernelManager` (and its subclasses).  As a result, from an EG config perspective, I think all that you need is `EG_KERNEL_INFO_TIMEOUT`.  Calling the command line with `--RemoteMappingKernelManager.kernel_info_timeout=${EG_KERNEL_INFO_TIMEOUT}` _should_ just work (since RMKM is subclass or `MappingKernelManager`), no need for any of the changes in `remotemanager.py` or `processproxy.py` or need to document `KERNEL_INFO_TIMEOUT`.  You do need the shell and yaml changes, provided you want your value set from the env or command line.

((Of course, I probably something wrong here, but if my memory serves me correctly (fat chance) I believe this can be trimmed down to essentially documenting the EG variable and CLI.))

lresende:
@kevin-bates sorry it took a little while to get back to this, but I believe I did the cleanups necessary, let me know what else needs to be done here...

docs/source/users/kernel-envs.md
```diff
@@ -55,6 +55,10 @@ There are several supported `KERNEL_` variables that the Enterprise Gateway serv
     expected to exceed that of the EG_KERNEL_LAUNCH_TIMEOUT set when Enterprise
     Gateway starts.
 
+  KERNEL_INFO_TIMEOUT=<from user> or EG_KERNEL_INFO_TIMEOUT=60
```
kevin-bates, line 58: Please remove mention of `KERNEL_INFO_TIMEOUT` since it can't be used for this

kevin-bates (commented):
Just one doc cleanup then we're good.

lresende:
> Just one doc cleanup then we're good.

@kevin-bates done, also rebased so the doc builds should be passing now

kevin-bates (approved):
Thank you!

PR #1358, opened 2024-01-06
Replace conda with mamba on EG docker image

lresende:


kevin-bates (approved):
Sounds good @lresende, thanks!

PR #1361, opened 2024-01-24
fix to use secret authentication information not containerd setting file hard coding

mmmommm:


mmmommm:
I'm sorry, I meant to create a patch for my own use, but I accidentally created a pull request instead.

Issue #1363, opened 2024-01-30
upgrade the os version used by the kernel-image-puller from Debian 11 to Debian 12


mmmommm:
relate: https://github.com/jupyter-server/enterprise_gateway/issues/759

### Problem
I am trying to pull from a private registry using the `CRICTL_AUTH` environment variable. However, this feature is only available in crictl (cri-tools) versions v1.24.0 and above.
Currently, I am using v1.21.0 which is related to the OS version.

[cri-tools v1.21.0](https://github.com/kubernetes-sigs/cri-tools/blob/80bcedb1222865e848a4232a4bcf1a388b328ba9/cmd/crictl/image.go#L56-L60)
[cri-tools v1.24.0](https://github.com/kubernetes-sigs/cri-tools/blob/f9710f792992ae7264d83cc894621174454d719c/cmd/crictl/image.go#L57-L62)

### Proposed Solution
I propose two potential solutions:
- Upgrade to Debian 12. This OS version uses cri-tools version v1.26.0.
- Independently upgrade cri-tools to v1.24.0 or higher.

If there are no major issues, it would be preferable to upgrade the OS version.

### Additional context
I tried the method of mounting k8s secrets as suggested in [this issue](https://github.com/jupyter-server/enterprise_gateway/issues/759), but it did not work well with crictl.

I have applied a patch locally and it works fine. However, maintaining the patch is quite challenging, so I would like to incorporate this change.

https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/Debian_11/
https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/Debian_12/

mmmommm:
This issue is resolved by the following pr.
https://github.com/jupyter-server/enterprise_gateway/pull/1364

PR #1364, opened 2024-01-30
update crictl version from v1.21.0 to v1.24.0

mmmommm:
relate: https://github.com/jupyter-server/enterprise_gateway/issues/1363

This is **Independently upgrade cri-tools to v1.24.0 or higher** solution.

> I am trying to pull from a private registry using the CRICTL_AUTH environment variable. However, this feature is only available in crictl (cri-tools) versions v1.24.0 and above.
Currently, I am using v1.21.0 which is related to the OS version.

### DIFF
--- a/etc/docker/kernel-image-puller/Dockerfile
+++ b/etc/docker/kernel-image-puller/Dockerfile
@@ -14,6 +14,9 @@ ARG OS=Debian_11
 RUN echo "deb https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/${OS}/ /"|tee /etc/apt/sources.list.d/devel:kubic:libcontainers:stable.list
 RUN curl -L https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/$OS/Release.key | apt-key add -
 RUN apt-get update && apt-get install cri-tools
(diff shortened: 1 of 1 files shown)


kevin-bates:
Hi @mmmommm - thank you for the issue and subsequent pull request - I apologize for the lack of response.

Since we're talking only about the KIP image and assuming it would make on-going maintenance easier over installing an explicit version of crictl, I'd be in favor of moving to Debian 12 instead (as I believe you prefer).  Would you mind pivoting to that approach?

mmmommm:
@kevin-bates 

Thank you for your response.

>I'd be in favor of moving to Debian 12 instead

Of course, I agree with pivoting this pr to move to Debian 12 👍

mmmommm:
operation confirmation

```
$ cd ./etc/docker/kernel-image-puller
$ docker build  -t kernel-image-puller .
[+] Building 54.8s (25/25) FINISHED 
####################### Responses for command execution are being truncated.

$ docker run -d --name kip kernel-image-puller&& docker exec kip crictl -v
0d426a1de30e7cf6814f5266e23a88efc7132b7d5732cb0894c9592019be738a
crictl version v1.26.0
```

kevin-bates (approved):
LGTM - thank you @mmmommm!

mmmommm:
@kevin-bates 
I have a question, when do you think we will be able to use the fixed image?

PR #1370, opened 2024-02-29
Better handling of an empty container status in `confirm_remote_startup`

OrenZ1:
Modified `confirm_remote_startup` method to make handling of empty container status better, so that if a kernel pod dies while still in startup, the startup process will fail and the user won't have to wait for timeout.


OrenZ1:
Hey! Is there any update on this PR?

Issue #1357, opened 2023-12-27
Add the option to terminate pending kubernetes kernels if they have events preventing them from starting


OrenZ1:
### Additional context

<!-- Add any other context or screenshots about the feature request here. You can also include links to examples of other programs that have something similar to your request. For example:

* Another project [...] solved this by [...]

OrenZ1:
If I can get an update about this request, that would be great.
I will be happy to contribute and add this option, so if you can state the relevant files, I can try to implement this and contribute :)

kevin-bates:
Hi @OrenZ1 - I apologize for the delay.  Unfortunately, I'm unable to spend much time on EG (and Jupyter in general) these days. 

I think this would be a great addition. Ideally, if we can determine that a Pending state is going to remain pending until the prescribed (and long) timeout, it would better to abort.  The location where we can detect this _during the startup sequence_ is in the `KubernetesProcessProxy` and the status loop where we could add more intelligence would be [here](https://github.com/jupyter-server/enterprise_gateway/blob/6334b01bdf3df326e6cec1b13e0a8ee253d63b66/enterprise_gateway/services/processproxies/k8s.py#L82).

I hope you find that helpful but imagine you've probably poked around a bit already so let me know if this isn't what you were looking for. 

Thank you for your interest and helping out!

lresende:
There are multiple ways you can go about this:
- Configure kernel image pullers to avoid delays in downloading images and reduce the startup timeouts
- Configure culling kernels to avoid kernels wasting resources
- If this is related to spark? Enable dynamic allocation to help reduce idle usage of resources

Also, having what @kevin-bates proposes above would not only help your use case but also fix a file-handlers leak that I have seen in the past.

OrenZ1:
Hi! Sorry for the delay but I managed to make a PR for the first thing we've discussed here!
For now the PR is for when the kernel pod dies while still in startup, the EG will throw a matching exception to the user, to prevent the need to wait for timeout.

I am still trying to think of a way to handle kernels which are stuck on Pending state. Hope to make a different PR for that too soon :)
#1370

OrenZ1:
Just created a new PR, which enables the option to configure different timeouts for different events which occur during startup, including a "0 seconds" timeout -which means the startup will terminate immediately after such event occurs.

PR #1383, opened 2024-06-02
Feature/configure kernel launch terminate on events

OrenZ1:
This pull request adds the option to configure kernel container events which should cause a termination of the kernel launch process. 
The user can now configure "unresolvable" events, to prevent a long wait for the timeout when such events occur.
It also allows the users to set a different timeout for some events, in example when a user knows an event may resolve after a small wait, but if it didn't, they want to terminate the kernel launch process.

### DIFF
--- a/enterprise_gateway/services/processproxies/container.py
+++ b/enterprise_gateway/services/processproxies/container.py
@@ -51,6 +51,12 @@ class ContainerProcessProxy(RemoteProcessProxy):
         self._initialize_kernel_launch_terminate_on_events()
 
     def _initialize_kernel_launch_terminate_on_events(self):
+        """
+        Parse the `kernel_launch_terminate_on_events` configuration, for easier access during startup.
+        [{"type": "Warning", "reason": "FailedMount", "timeout_in_seconds": 0},
+        {"type": Warning", "reason": "Unschedulable", "timeout_in_seconds": 30}] ->
+        {"Warning": {"FailedMount": 0, "Unschedulable": 30}}
+        """
         self.kernel_launch_terminate_on_events = defaultdict(dict)
         for configuration in self.kernel_manager.parent.kernel_launch_terminate_on_events:
             self.kernel_launch_terminate_on_events[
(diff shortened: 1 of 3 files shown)


OrenZ1:
Discussed in #1357

enterprise_gateway/services/processproxies/container.py
```diff
@@ -212,6 +231,33 @@ async def confirm_remote_startup(self) -> None:
...
             else:
                 self.detect_launch_failure()
+        self.kernel_events_to_occurrence_time = {}
+
+    def _handle_pending_kernel(self):
+        self.log.debug("Sampling kernel container events")
+        kernel_pod_events = self.get_container_events()
+        for event in kernel_pod_events:
+            if (
+                event.type in self.kernel_launch_terminate_on_events
```
kevin-bates, line 241: Since the event type and reason need to be matches, is there a doc reference we could add that enumerates the various values that could be expected?  That said, I suppose folks will need to be reactive (vs. proactive) in this sense and will have the event information when troubleshooting, but it still might be helpful to have a link or two, or add the enumeration directly.

OrenZ1: Added a docstring to `handle_pending_kernel` method in `container.py`, as well as a more detailed configuration help. I found it hard to find a detailed list of Kubernetes events by type and reason, so I added an enumeration which may differ in different platforms (or even versions of K8S itself). I agree that folks will need to be reactive, but I did also add a link to a guide I found helpful with finding the correct event type and reason.

OrenZ1:
Hi! Any update on this?

kevin-bates (commented):
I think this option should have a presence in the [Operator's Guide](https://jupyter-enterprise-gateway.readthedocs.io/en/latest/operators/index.html) somewhere but can't really find a good location.  The only way folks will know about this right now is looking at the code.  Thinking we should at least add it into [Command-line Options](https://jupyter-enterprise-gateway.readthedocs.io/en/latest/operators/config-cli.html#command-line-options) in the `RemoteMappingKernelManager` section.

OrenZ1:
Added presence in the `Command-Line Options` in the `RemoteMappingKernelManager` section.

kevin-bates (approved):
LGTM - thank you @OrenZ1.

Issue #1379, opened 2024-04-02
JEG 3.2.3 on K3s v1.28.5 unable to start kernels. 3.2.2 pyspark worked.


paf91:
## Description
Whenever I start kernel based on kubernetes I'm getting this error on version 3.2.3:

Error Starting Kernel
HTTP 500: Internal Server Error (Error from Gateway: [Error occurred creating role binding for namespace 'guest-fab3e59b-edbb-4e1d-912e-087b1798425b': module 'kubernetes.client' has no attribute 'V1Subject'] Error occurred creating role binding for namespace 'guest-fab3e59b-edbb-4e1d-912e-087b1798425b': module 'kubernetes.client' has no attribute 'V1Subject'. Ensure gateway url is valid and the Gateway instance is running.)

## Reproduce

values.yaml:

```yaml
service:
  type: "LoadBalancer"
  # Master public IP on which to expose EG.
  k8sMasterPublicIP: '<redacted, private ip like 10.x.x.x>'
  ports:
    - name: "http"
      port: 8888
      targetPort: 8888
    - name: "http-response"
      port: 8877
      targetPort: 8877
ingress:
  enabled: false
kernel:
  shareGatewayNamespace: false
  allowedKernels:
    - r_kubernetes
    - python_kubernetes
    - python_tf_kubernetes
    - python_tf_gpu_kubernetes
    - scala_kubernetes
    - spark_r_kubernetes
    - spark_python_kubernetes
    - spark_scala_kubernetes
    - spark_python_operator
    - python3
  defaultKernelName: python_kubernetes
kip:
  enabled: true
  serviceAccountName: 'kernel-image-puller-sa'
  criSocket: /run/containerd/containerd.sock
```

`helm  upgrade --install  enterprise-gateway   https://github.com/jupyter-server/enterprise_gateway/releases/download/v3.2.3/jupyter_enterprise_gateway_helm-3.2.3.tar.gz    --namespace enterprise-gateway    -f ~/jupyter/gateway/values-balancer.yaml`

kubectl get pods -n enterprise-gateway:
![image](https://github.com/jupyter-server/enterprise_gateway/assets/12195041/2e430911-d37a-4d3f-b5f3-da4f73c5ae5d)

curl -X POST -i 'http://<redacted_private_ip>:8888/api/kernels' --data '{ "name": "spark_python_kubernetes", "env": { "KERNEL_USERNAME": "jovyan" }}'

[...]

PR #1384, opened 2024-06-03
change V1Subject to RbacV1Subject in k8s.py

merqri:
This change fix - Error: module 'kubernetes.client' has no attribute 'V1Subject'

This error occurs when an attempt is made to create a role binding.
The full error looks as follows:
```
[E 2024-06-03 08:58:36.252 EnterpriseGatewayApp] Error occurred creating role binding for namespace 'jovyan-153ca7c1-c68f-440d-b678-f8926b48afbc': module 'kubernetes.client' has no attribute 'V1Subject'
```

### DIFF
--- a/enterprise_gateway/services/processproxies/k8s.py
+++ b/enterprise_gateway/services/processproxies/k8s.py
@@ -349,7 +349,7 @@ class KubernetesProcessProxy(ContainerProcessProxy):
         binding_role_ref = client.V1RoleRef(
             api_group="", kind="ClusterRole", name=kernel_cluster_role
         )
-        binding_subjects = client.V1Subject(
+        binding_subjects = client.RbacV1Subject(
             api_group="", kind="ServiceAccount", name=service_account_name, namespace=namespace
         )
 
--- a/enterprise_gateway/services/processproxies/k8s.py
+++ b/enterprise_gateway/services/processproxies/k8s.py
@@ -10,6 +10,7 @@ import os
 import re
 from typing import Any
 
+import kubernetes
 import urllib3
 from jinja2 import BaseLoader, Environment
 from kubernetes import client, config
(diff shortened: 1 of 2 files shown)


kevin-bates:
Sorry, just caught up on #1379.

Since this appears to be due to version changes, do we need to concern ourselves with breaking older installations?

@lresende - looks like you had some thoughts on this (capping client?).  Does that prevent use on newer versions of k8s like 1.30 (or even 1.29)?

lresende:
Capping the client to 1.28 would make it work for current users, but would make it fail for users that have upgraded to >= 1.29 then... Another approach would be to try to check in the code for the current version and use one or the other.

enterprise_gateway/services/processproxies/k8s.py
Vivaciouslady, line 1: Delete it from a hacker

huynhsontung:
> Does that prevent use on newer versions of k8s like 1.30 (or even 1.29)?

@lresende @kevin-bates I am running Kubernetes 1.31 with EG 3.2.2 since 3.2.3 has this issue, and everything works fine as far as I can see. That said, I think this PR is backward-compatible as looking at the [commit that changes V1Subject to RbacV1Subject](https://github.com/kubernetes-client/python/commit/ab097519b892fee271fa6366d7e2893bee36dfc3#diff-6956e8fe55d1526e2e671f4558de6d2a8d55291059eabc1bcd4c4f36bc0e730f), only the model name is updated. I don't think this is a breaking change in the API, as I couldn't find any mentions of it in the v29 changelog.

lresende:
@merqri how about we do something like:
```
+def get_subject_class():
+    """
+    Returns the appropriate Subject class based on the kubernetes client version.
+
+    In kubernetes-client, V1Subject was renamed to RbacV1Subject.
+    This function returns the appropriate class based on the installed version.
+    """
+    # Check if V1Subject exists in the client
+    if hasattr(client, 'V1Subject'):
+        logging.debug("Using client.V1Subject for Kubernetes client version: %s", kubernetes.__version__)
+        return client.V1Subject
+    # Fall back to RbacV1Subject for older versions
+    logging.debug("Using client.RbacV1Subject for Kubernetes client version: %s", kubernetes.__version__)
+    return client.RbacV1Subject
+
```
and then something like:
```
-        binding_subjects = client.RbacV1Subject(
+        # Use the appropriate Subject class based on kubernetes client version
+        SubjectClass = get_subject_class()
+        binding_subjects = SubjectClass(
             api_group="", kind="ServiceAccount", name=service_account_name, namespace=namespace
         )
```

if you are ok I can push an update to your pr.

lresende (approved):

PR #1408, opened 2025-06-23
Enable specifying helm chart deployment namespace

### LINKED ISSUE #1407: Allow specifying namespace when deploy using helm chart
When using JEG as sub-chart, the current helm chart implementation forces the use of the release namespace and does not allow for deploying JEG on it's own namespace

fresende:
Fixes #1407


lresende:
@fresende could you please rebase the pr

PR #1409, opened 2025-06-23
Remove Python 3.8 to fix build issues

lresende:


lresende:
Thank you @kevin-bates , I believe there are two main issues here:
- zmq build related issues
- default python 3.12 from jupyterlab maintainer tools not allowing to downgrade python

Will find some time and continue on this until we get a green build

PR #1412, opened 2025-07-25
Fix KERNEL_POD_NAME substitution to avoid SSTI

lresende:


lresende:
Thank you @Zsailer , good catch. I have updated it:
  1. Updated the regex pattern in enterprise_gateway/services/processproxies/k8s.py:226 to prevent variables starting with underscore:
    - Changed from: r'\{\{\s*([a-zA-Z_][a-zA-Z0-9_]*)\s*\}\}'
    - To: r'\{\{\s*([a-zA-Z][a-zA-Z0-9_]*)\s*\}\}'
  2. Enhanced the test case in enterprise_gateway/tests/test_process_proxy.py to verify that underscore-prefixed variables (like __class__, __dict__, __globals__) are properly blocked and treated as invalid template syntax.

Zsailer (approved):

PR #1428, opened 2025-11-04
[WIP] Interactive Ray support

### LINKED ISSUE #939: Add support for Ray on Kubernetes
## Description
It would be great to be able to support the same behavior from `Spark on Kubernetes` but using Ray where a cluster+workers would be created for running a kernel's workload.

lresende:
Add support for launching interactive Jupyter kernels on Ray clusters managed by the Kubernetes Ray Operator. It adds 
a new RayOperatorProcessProxy to manage kernel lifecycle through Ray Custom Resource Definitions (CRDs). This process proxy intelligently monitors both the RayCluster CRD state (managed by ray.io/v1alpha1 API) and the underlying Kubernetes pod status, implementing a two-phase readiness check that first verifies the Ray cluster reaches "ready" state before confirming the kernel head pod is running. The implementation includes sophisticated container status detection that returns unified pod states compatible with   Enterprise Gateway's existing lifecycle management while properly handling Ray-specific states like cluster initialization, head pod scheduling, and worker node scaling. 

The commit provides a complete end-to-end solution, including a custom Jinja2 template (ray.io-v1alpha1.yaml.j2) that generates RayCluster CR manifests with optimized configurations for kernel workloads with configurable idle timeout, and resource limits. A new Docker image (kernel-ray-py) packages the Ray-enabled Python kernel with all necessary dependencies, while the ray_python_operator kernel specification integrates seamlessly with Enterprise Gateway's
  existing kernel management infrastructure. The changes include updates to Helm charts for proper RBAC configuration (adding Ray CRD permissions to the cluster role), and Makefile enhancements for building Ray kernel images. The implementation maintains backward compatibility with existing kernel types while enabling data scientists to leverage Ray's distributed computing capabilities directly from interactive Jupyter notebooks with full kernel lifecycle management, automatic cleanup, and proper resource isolation.

Fixes #939


enterprise_gateway/services/processproxies/container.py
```diff
@@ -197,21 +200,34 @@ async def confirm_remote_startup(self) -> None:
...
                     )
                 else:
+                    self.log.debug(
+                        f">>> container.confirm_remote_startup(): is hosted assigned => {self.assigned_host}"
+                    )
+                    self.log.debug(">>> should call receive_connection_info()")
                     if self.assigned_host:
                         ready_to_connect = await self.receive_connection_info()
+                        self.log.debug(
+                            f">>> container.confirm_remote_startup(): ready to connect => {ready_to_connect}"
```
kevin-bates: I'm cool with leaving these debug statements in - perhaps we can remove some of them later. However, can we avoid having two statements display the same message?  This statement is the same as that on L229, so it would be good differentiate these.

enterprise_gateway/services/processproxies/ray_operator.py
```diff
@@ -0,0 +1,211 @@
...
+        self.kernel_resource_name = self._determine_kernel_pod_name(**kwargs)
+        kwargs["env"]["KERNEL_RESOURCE_NAME"] = self.kernel_resource_name
+        kwargs["env"]["KERNEL_CRD_GROUP"] = self.group
+        kwargs["env"]["KERNEL_CRD_VERSION"] = self.version
+        kwargs["env"]["KERNEL_CRD_PLURAL"] = self.plural
+
+        await super().launch_process(kernel_cmd, **kwargs)
+        return self
+
+    def get_container_status(self, iteration: int | None) -> str:
```
kevin-bates: The code here looks very generic and feels like it belongs on the superclass.  In speaking with you offline - it sounds like the RayOperator CRD expects different status values for running and ready, etc.  Perhaps we can introduce override methods to account for CRD's with different status requirements in their lifecycle.

We would need to be aware of the Spark operator, but, actually, the default implementations on the base class would match what that operator expects anyway - so perhaps its not too risky.

etc/docker/demo-base/Dockerfile
```diff
@@ -27,8 +27,8 @@ ENV SHELL=/bin/bash \
 ENV HOME=/home/$NB_USER \
     PATH=$JAVA_HOME/bin:$ANACONDA_HOME/bin:$HADOOP_HOME/bin:$SPARK_HOME/bin:$PATH
 
-ENV SPARK_VER $SPARK_VERSION
-ENV HADOOP_VER 3.3.1
+ENV SPARK_VER=$SPARK_VERSION
```
kevin-bates: Thanks for updating this (and elsewhere)!

etc/kubernetes/helm/enterprise-gateway/templates/deployment.yaml
```diff
@@ -73,6 +73,8 @@ spec:
           value: !!str {{ .Values.kernel.launchTimeout }}
         - name: EG_KERNEL_INFO_TIMEOUT
           value: !!str {{ .Values.kernel.infoTimeout }}
+        - name: EG_REQUEST_TIMEOUT
```
kevin-bates: :+1:

kevin-bates (commented):
Good stuff @lresende - thank you!  This will be useful

Just a couple nits about logging.  I'm cool with keeping the debugs in place since this area is a common problem when bootstrapping new process proxies.

The other comment is considering perhaps moving some of the CRD-based code to the superclass as we better understand the patterns of CRDs.

PR #1445, opened 2026-07-10
Clarify client vs. server authentication in user docs

### LINKED ISSUE #1444: How to change default "guest" username and password for Enterprise Gateway?
Hi everyone,

I'm trying to change the default username and password for Jupyter Enterprise Gateway. I noticed that the documentation mentions `guest` and `guest-password` as the default credentials.

I might have missed it in the documentation, but I couldn't find a clear way to change these defaults on the server side.

Is it possible to configure a custom username and password on the Enterprise Gateway server, or is there a specific configuration file / environment variable to override these defaults?

Thank you!

lresende:
Users frequently assume Enterprise Gateway ships with a default "guest" / "guest-password" account and look for a server-side way to change it. In reality, http_user/http_pwd are the *client's* HTTP Basic Auth credentials, both default to None, and a stock Gateway server does not validate them at all.

Reduce this confusion in the Users Guide:

- Replace the literal guest/guest-password example values with <username>/<password> placeholders so they no longer read as defaults.
- Add an "Authenticating to Enterprise Gateway" section explaining that the Gateway performs no user authentication by default, and describing the three request-auth mechanisms (Basic Auth, bearer/auth_token gated by EG_AUTH_TOKEN, and KERNEL_USERNAME identity), with a Mermaid diagram showing where each option applies.
- Split the connect example into an unauthenticated form and an explicit HTTP Basic authentication form to show credentials are optional.
- Clarify in the Users Guide index and client installation pages that JupyterLab and Jupyter Notebook both run on top of Jupyter Server.

Fixes #1444

### DIFF
--- a/docs/source/users/connecting-to-eg.md
+++ b/docs/source/users/connecting-to-eg.md
@@ -1,4 +1,4 @@
-# Connecting the server to Enterprise Gateway
+# Connecting JupyterLab or Jupyter Notebook to Enterprise Gateway
 
 To leverage the benefits of Enterprise Gateway, it's helpful to redirect a Jupyter server's kernel management to the Gateway server. This allows better separation of the user's notebooks from the managed computer cluster (Kubernetes, Hadoop YARN, Docker Swarm, etc.) on which Enterprise Gateway resides. A Jupyter server can be configured to relay kernel requests to an Enterprise Gateway server in several ways.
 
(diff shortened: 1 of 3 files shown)


Zsailer (approved):

Zsailer:
LGTM! Thanks @lresende!

